PR #212, opened 2026-06-15
feat: reconstruction wrapping gaps — outerShell, mesh quality flag, WireCurve (closes #211) → v1.5.2

gsdali:
## Summary

Closes the confirmed wrapping gaps from the mesh→CAD reconstruction coverage audit (#211). All grep-verified as genuinely missing before this PR.

### Implemented

- **`Shape.outerShell -> Shape?`** — `BRepClass3d::OuterShell`. The outer body shell of a solid, distinguishing it from internal void shells (e.g. decompose a part into outer-body + cavities). `nil` for non-solids; also accepts a compound wrapping a single solid.
- **`MeshParameters.allowQualityDecrease`** — `IMeshTools_Parameters::AllowQualityDecrease` (default `false`), the one mesh knob that was missing. Lets a re-mesh at a different deflection actually replace an existing triangulation instead of OCCT silently keeping it — needed for a deviation re-measure at a finer/coarser resolution.
- **`WireCurve`** — `BRepAdaptor_CompCurve` as a standalone Swift type: a multi-edge wire as one **arc-length** curve. `length`, `point(atAbscissa:)` / `tangent(atAbscissa:)` (walk across edge boundaries seamlessly), plus native `parameterRange` / `point(atParameter:)` / `tangent(atParameter:)`. Replaces ad-hoc per-edge sampling when placing loft sections along a measured wire.

### Verified, no change needed

- **`Shape.minDistance(to:) -> Double?` already exists** (the issue's minor convenience) — and returns `Double?`, which is more correct than the proposed `Double`.
- **Scattered point-cloud `GeomAPI_PointsToBSplineSurface` is not an OCCT capability** — every constructor takes a grid (`Array2<gp_Pnt>` / `Array2<double>`); there is no unstructured-cloud overload. A real cloud fit means resampling to a grid (already wrapped: `Surface.fromPointGrid`) or `GeomPlate` / `BRepOffsetAPI_MakeFilling` (already wrapped). Documented so it isn't re-investigated.

## Tests

10 tests, all passing (run `--no-parallel`):
- **Curve** (`Issue211WireCurve`): length across edges (L-wire = 20); `point(atAbscissa:)` walks the corner (5→(5,0,0), 10→corner, 15→(10,5,0)); tangent flips (1,0,0)→(0,1,0); even sampling count.
- **Topology** (`Issue211OuterShell`): hollow 20-cube (8-cube cavity) → `shellCount ≥ 2` and `outerShell` spans the full 20-cube; plain box → outer shell present; a face → `nil`.
- **Mesh** (`Issue211MeshParam`): default `false`; meshes with the flag; a coarser re-mesh replaces a finer one.

Additive, source-only (no xcframework change). CHANGELOG → **v1.5.2**. Closes #211.

### DIFF
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -2739,6 +2739,18 @@ void OCCTCompCurveParamRange(OCCTCompCurveRef ref, double* first, double* last);
 bool OCCTCompCurvePointAtParam(OCCTCompCurveRef ref, double u, double* x, double* y, double* z);
 bool OCCTCompCurveTangentAtParam(OCCTCompCurveRef ref, double u, double* x, double* y, double* z);  // unit tangent (D1)
 bool OCCTCompCurveParamAtAbscissa(OCCTCompCurveRef ref, double s, double* outParam); // param at arc length s from start
+int32_t OCCTCompCurveSampleUniform(OCCTCompCurveRef ref, int32_t count, double* outXYZ); // count equally-spaced (arc length) points; outXYZ holds count*3
+
+// EdgeCurve adaptor: a single edge as an arc-length curve (BRepAdaptor_Curve). #211/#212
+typedef struct OCCTEdgeCurve* OCCTEdgeCurveRef;
+OCCTEdgeCurveRef OCCTEdgeCurveCreate(OCCTEdgeRef edge);
+void OCCTEdgeCurveRelease(OCCTEdgeCurveRef ref);
+double OCCTEdgeCurveLength(OCCTEdgeCurveRef ref);
+void OCCTEdgeCurveParamRange(OCCTEdgeCurveRef ref, double* first, double* last);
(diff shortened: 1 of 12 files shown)


gsdali:
**Expanded scope** (follow-on commit a9415b9): added the nearest-neighbour ops to the three above, per review —

- **`WireCurve.points(count:)` / `points(spacing:)`** — even arc-length sampling (`GCPnts_UniformAbscissa`), the common "sample evenly along a measured wire" case in one pass.
- **`EdgeCurve`** (`BRepAdaptor_Curve`) — single-edge sibling of `WireCurve`; adds the arc-length side that `Edge`'s native param-based `point/tangent` lacked.
- **`Shape.innerShells`** — the void/cavity shells (all except `outerShell`); completes outer-body + cavities decomposition.

Shared bridge helper `sampleAdaptorUniform(Adaptor3d_Curve&)` drives both adaptors. Now **13 tests** across Curve/Topology/Mesh. Still additive/source-only; CHANGELOG v1.5.2 updated.

PR #265, opened 2026-06-25
fix(#263): carry OCCT kernel patch — ShapeFix_Face non-face context replacement

gsdali:
Draft — carries the upstream OCCT kernel fix for #263. **Takes effect only on the next xcframework rebuild**; the in-wrapper guard shipped in v1.8.3 (#264) protects consumers until then.

## Root cause (upstream OCCT — Open-Cascade-SAS/OCCT#1322)

`ShapeFix_Face::Perform` (ShapeFix_Face.cxx:382-383) does:

```cpp
TopoDS_Shape S = myFace;
if (!Context().IsNull()) S = Context()->Apply(myFace);
TopoDS_Shape emptyCopied = S.EmptyCopied();
TopoDS_Face  tmpFace     = TopoDS::Face(emptyCopied);   // <-- unchecked cast
```

When an earlier fix sharing the same `ShapeBuild_ReShape` context has replaced the face with a **compound** (a self-intersecting face split into several faces), `Apply()` returns a non-face. `TopoDS::Face` over a compound `TShape` builds an invalid face handle → heap corruption → uncatchable OS signal downstream (`FixOrientation` → `BRep_Tool::Curve` → `BRep_TEdge::EmptyCopy`, SIGSEGV/SIGBUS at varying addresses — the exact #263 backtrace).

Pinned by linking a debug-built `ShapeFix_Face.cxx` over the prebuilt lib: at `-O0` the cast throws `Standard_TypeMismatch` at line 383 with `S.type=COMPOUND`, `myFace.type=FACE` (on the 6th of 6 prism faces; the first 5 fix cleanly).

## Fix

Guard the cast — if `S` is not a `TopAbs_FACE`, record it as the result and return (the replacement is already in the context; nothing to fix here). One carried patch: `Scripts/patches/0001-ShapeFix_Face-guard-non-face-context-replacement-263.patch`.

## Validation (single-TU override-link at -O2, no full rebuild)

- Bowtie reproducer: crashes 3/3 on stock OCCT 8.0.0p1 → **survives 5/5** patched.
- All four `OCCTReconstruct` `crash-repro` fixtures (b1214/b693/b135/b106): **survive**.
- `ShapeFix_Shape` output on valid box/sphere/cylinder is **byte-identical** to stock (the guard only triggers for a non-face replacement, which never happens for a well-formed face).

## Next steps

- Rebuild `OCCT.xcframework` with the patch (`Scripts/build-occt.sh`), attach the `.zip` asset, bump `Package.swift` URL+checksum — then the in-wrapper guard becomes belt-and-suspenders.
- Upstream: offer the same fix to Open-Cascade-SAS/OCCT#1322.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,8 +35,8 @@ let occtTarget: Target = useLocalBinary
     )
     : .binaryTarget(
         name: "OCCT",
-        url: "https://github.com/SecondMouseAU/OCCTSwift/releases/download/v1.7.1/OCCT.xcframework.zip",
-        checksum: "588aea7eb588063b906878fb66f4a2b91de6b8034be95dcb5212470deff8bccf"
+        url: "https://github.com/SecondMouseAU/OCCTSwift/releases/download/v1.8.4/OCCT.xcframework.zip",
+        checksum: "adee66e64ba9c613d66a77f246efd3cde667df9e46eae9817826f9b8ef19affb"
     )
 
 let package = Package(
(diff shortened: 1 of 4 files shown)


gsdali:
Upstream fix submitted: Open-Cascade-SAS/OCCT#1323 (same top-of-Perform guard + a GTest regression case). When that merges and ships in an OCCT release, this carried patch can be dropped.

Issue #377, opened 2026-07-26
Segmented duplication audit: programme record, deferred passes 2a to 5d


gsdali:
Ecosystem policy: [Search before building](https://github.com/SecondMouseAU/ecosystem/blob/main/okf/policies/search-before-building.md).

## What this issue is now

The programme-level record for the segmented duplication audit. **It no longer tracks the work that
is in the v2.0.0 release.** That moved to #669 when the release scope was set:

| Moved to #669 | |
|---|---|
| #380 | Pass 1a, geometry primitives |
| #381 | Pass 1b, C++ bridge header |
| #393, #394, #395 | The three file breakdowns |

Those five are in the release. Everything still listed below is **deferred until after v2.0.0**.
They remain real work and they stay on the board; they are simply not gating the major.

Read [docs/v2.0.0-plan.md](https://github.com/SecondMouseAU/OCCTSwift/blob/refactor/381-pass1b/docs/v2.0.0-plan.md)
for why the line was drawn there.

## Branch policy (corrected 2026-08-11)

Every sub-issue below used to declare a single shared `Branch: refactor/377-segmented-audit`. That
branch is dead: measured against `main`, it is frozen at 2026-07-29 (its tip is literally the old
Pass 1a integration branch, PRs #453–#483), 38 commits behind `main` (missing Pass 1b, OCCT 8.0.1,
the Shape.swift/Document.swift/bridge-header breakouts, and all of v2.0.0), and its own "ahead"
commits are stale duplicates of work that already landed on `main` through `refactor/381-pass1b`
under different SHAs. It was abandoned the day it was cut and nothing since Pass 1a ever used it.

**Corrected policy: one fresh branch per pass, cut from `main`'s tip when that pass actually starts**,
not one long-lived shared branch. Naming follows #381's own convention: `refactor/<issue>-pass<name>`
(e.g. `refactor/382-pass2a`). No release-bundling pressure applies here the way it did for 1b (no
version bump riding along), so there's no need for a shared integration branch holding several
passes' work at once — each pass's PRs can target `main` directly once its branch is cut.

Every sub-issue's `Branch:` line has been corrected to match. `refactor/377-segmented-audit` itself
is left in place for now (nothing currently points at it) rather than deleted as part of this note.

## Why the audit was segmented in the first place

`Sources/OCCTSwift` is roughly 56,000 lines across 71 files, about 2.6x
[...]

PR #423, opened 2026-07-27
fix(#399): unify the conic preconditions shared by Curve3D's two factory families

gsdali:
## What & why

`Curve3D` builds four conic types through two parallel factory families with identical
signatures: a direct one (`circle`/`ellipse`/`parabola`/`hyperbola`, which constructs `Geom_*`
objects from a hand-built `gp_Ax2`) and a `*FromCenterNormal` one routed through OCCT's
`gce_Make*` algorithms. They are geometrically identical — `gce_MakeCirc(P, N, R)` builds the
very same `gp_Ax2(P, N)` frame — but only the direct family ever got explicit dimension guards.

The `gce_Make*` algorithms reject only *strictly negative* dimensions, so all four pairs
disagreed at the zero boundary. `Curve3D.circle(radius: 0)` returns `nil`;
`Curve3D.circleFromCenterNormal(radius: 0)` returned a live, degenerate zero-radius curve. Same
for `ellipseFromCenterNormal(minorRadius: 0)`, `hyperbolaFromCenterNormal` with either radius 0,
and `parabolaFromCenterNormal(focal: 0)`.

Both families now route through one set of preconditions in `OCCTBridge_Curve3D.mm`
(`occtValidCircleRadius`, `occtValidEllipseRadii`, `occtValidHyperbolaRadii`,
`occtValidParabolaFocal`), so the contract is stated once and cannot drift apart again.

Refs #399. Part of the #380 / #377 duplication audit — targets `refactor/377-segmented-audit`,
not `main`.

## Checklist

      verification)

New suite `Curve3DConicFactoryParityTests` (`Tests/OCCTCurveTests`) cross-checks each pair at the
boundary in both directions: all four pairs reject the same degenerate inputs, and for valid
inputs both families are sampled along the curve and asserted to agree to 1e-9. It also closes
the direct family's own gaps the issue flagged — there was no zero-rejection test for
`hyperbola` or `parabola` at all.

## Notes for the reviewer

**The auditor's original claim needed correcting, and the issue body already records this**: the
finding said the two families disagreed on inverted ellipse radii (`majorRadius < minorRadius`).
They do not — `gce_MakeElips` rejects that itself with `gce_InvertRadius`. The real, verified
divergence is the zero boundary, and it affected all four pairs rather than just the ellipse.

**Verification that the bug was live**, not just theoretically reachable: reverting only
`OCCTBridge_Curve3D.mm` and re-running the new suite fails with 7 issues — circle 1, ellipse 2,
hyperbola 3, parabola 1. That distribution is itself the confirmation of the correction above:
the inverted-radii and negative-radii cases in the same test tables pass *before* the fix, so
only the zero cases were ever divergent.

**Why guards rather than delegation.** Having the gce family call the direct bridge functions
would also give one code path, but it would make the documented `gce_Make*` provenance false
(`docs/API_REFERENCE.md` maps these to the gce factories) and would silently change which OCCT
algorithm runs. Sharing the precondition helpers fixes the drift where it actually lived, keeps
both OCCT paths intact, and also fixes any direct C consumer of the bridge functions.

**Unified on the stricter contract** (reject zero) rather than the looser one: a zero-radius
circle or zero-focal parabola is a degenerate curve with no useful downstream behaviour, and the
direct family's `nil` is the already-documented and already-tested contract.

Full `swift test` (4438 tests) clean. One unrelated intermittent failure appeared in one of three
runs — `"Shape.loadRobust interrupts repair, not just the transfer (#300)"`, the CPU-timing-
sensitive cancellation test whose own source comment documents it as flaky under parallel
execution — and passed in the other two.

No `CHANGELOG.md` / version entry here deliberately: the audit branch merges to `main` as one
one. The release entry lands with the branch merge.

### DIFF
--- a/Sources/OCCTBridge/src/OCCTBridge_Curve3D.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Curve3D.mm
@@ -251,6 +251,28 @@ void OCCTCurve3DD2(OCCTCurve3DRef c, double u,
 
 // Primitive Curves
 
+// Conic dimension preconditions, shared by the two factory families that build the
+// same four curve types: OCCTCurve3DCreate{Circle,Ellipse,Parabola,Hyperbola} (direct
+// Geom_* construction) and OCCTGceMake{CircFromCenterNormal,Elips,Hypr,Parab} (gce_Make*
+// construction). The gce_Make* algorithms only reject strictly-negative dimensions, so
+// before #399 they silently produced degenerate zero-radius/zero-focal curves where the
+// direct family returned null for the identical input. One definition, both families.
+static inline bool occtValidCircleRadius(double radius) {
+    return radius > 0;
+}
+
+static inline bool occtValidEllipseRadii(double majorR, double minorR) {
+    return majorR > 0 && minorR > 0 && minorR <= majorR;
+}
+
+static inline bool occtValidHyperbolaRadii(double majorR, double minorR) {
+    return majorR > 0 && minorR > 0;
+}
+
+static inline bool occtValidParabolaFocal(double focal) {
+    return focal > 0;
+}
+
 OCCTCurve3DRef OCCTCurve3DCreateLine(double px, double py, double pz,
                                       double dx, double dy, double dz) {
     try {
@@ -281,7 +303,7 @@ OCCTCurve3DRef OCCTCurve3DCreateCircle(double cx, double cy, double cz,
                                         double nx, double ny, double nz,
                                         double radius) {
     try {
-        if (radius <= 0) return nullptr;
+        if (!occtValidCircleRadius(radius)) return nullptr;
         gp_Pnt center(cx, cy, cz);
         gp_Dir normal(nx, ny, nz);
         gp_Ax2 axis(center, normal);
@@ -324,7 +346,7 @@ OCCTCurve3DRef OCCTCurve3DCreateEllipse(double cx, double cy, double cz,
                                          double nx, double ny, double nz,
                                          double majorR, double minorR) {
     try {
-        if (majorR <= 0 || minorR <= 0 || minorR > majorR) return nullptr;
+        if (!occtValidEllipseRadii(majorR, minorR)) return nullptr;
         gp_Ax2 axis(gp_Pnt(cx, cy, cz), gp_Dir(nx, ny, nz));
         Handle(Geom_Ellipse) ellipse = new Geom_Ellipse(axis, majorR, minorR);
         return new OCCTCurve3D(ellipse);
@@ -337,7 +359,7 @@ OCCTCurve3DRef OCCTCurve3DCreateParabola(double cx, double cy, double cz,
                                           double nx, double ny, double nz,
                                           double focal) {
     try {
-        if (focal <= 0) return nullptr;
+        if (!occtValidParabolaFocal(focal)) return nullptr;
         gp_Ax2 axis(gp_Pnt(cx, cy, cz), gp_Dir(nx, ny, nz));
         Handle(Geom_Parabola) parabola = new Geom_Parabola(axis, focal);
         return new OCCTCurve3D(parabola);
@@ -350,7 +372,7 @@ OCCTCurve3DRef OCCTCurve3DCreateHyperbola(double cx, double cy, double cz,
                                            double nx, double ny, double nz,
                                            double majorR, double minorR) {
     try {
-        if (majorR <= 0 || minorR <= 0) return nullptr;
+        if (!occtValidHyperbolaRadii(majorR, minorR)) return nullptr;
         gp_Ax2 axis(gp_Pnt(cx, cy, cz), gp_Dir(nx, ny, nz));
         Handle(Geom_Hyperbola) hyp = new Geom_Hyperbola(axis, majorR, minorR);
         return new OCCTCurve3D(hyp);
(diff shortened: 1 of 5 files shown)


gsdali (commented):
## Review

Verified the claims against the OCCT sources and the surrounding bridge code. The diagnosis and the fix are sound, and the PR body's correction of the auditor's original claim is right. This should merge; the notes below are one real (and positive) behaviour change that is undocumented, and one scope gap.

### What I verified independently

- `gce_Make*` really does reject only strictly-negative dimensions: `gce_MakeCirc.cxx:186` (`Radius < 0.`), `gce_MakeHypr.cxx:64` (`MajorRadius < 0.0 || MinorRadius < 0.0`), `gce_MakeParab.cxx:26` (`Focal < 0.0`), `gce_MakeElips.cxx:29-35` (`InvertRadius` when major < minor, `NegativeRadius` when minor < 0). So the fix targets the right boundary, and the correction in the PR body (inverted ellipse radii were never divergent) holds.
- Each rewritten guard is exactly equivalent to the one it replaces for finite inputs, including the ellipse's three-way condition (`minorR > majorR` becomes `minorR <= majorR` under negation).
- The "identical frame" claim is stronger than the PR body states, and correct. `gce_MakeCirc(Center, Norm, R)` does not call `gp_Ax2(P, V)`; it reimplements the same max-component X-axis derivation (`gce_MakeCirc.cxx:192-230` vs `gp_Ax2.cxx:34-70`, the same branches character for character). So parity holds for any normal, not just the `(0,0,1)` the test uses.

Did not build or run the suite as part of this review; the clean 4438-test run plus the revert-and-fail check is the right evidence for this change.

### 1. The NaN contract changed, silently

`radius <= 0` and `!(radius > 0)` are not equivalent for NaN. Before this PR, `Curve3D.circle(center:normal:radius: .nan)` returned a *live* curve: NaN fails `<= 0`, and `Geom_Circle`'s own check is `if (R < 0.0) throw` (`Geom_Circle.cxx:64`), which NaN also fails, so the degenerate curve was constructed and returned. After this PR it returns `nil`, and the same applies to all four types in both families.

That is a genuine improvement and arguably the more valuable half of the fix, but it is beyond the stated zero-boundary scope, is not mentioned anywhere, and is not tested. The doc comments say only "zero and negative radii return `nil`", which is now incomplete. Cheapest fix: add `.nan` to the existing rejection tables (`for radius in [0.0, -1.0, .nan]`, `(Double.nan, 3)`, and so on) and say "non-positive or non-finite" in the docs.

### 2. The identical drift still exists in the 2D family, and the helpers cannot be shared with it

`OCCTCurve2DCreateCircle`/`Ellipse`/`Parabola`/`Hyperbola` guard (`OCCTBridge_Geom2d.mm:5026`, `:5068`, `:5099`, `:5114`), while `OCCTGceMakeCirc2dFromCenterRadius`/`Elips2d`/`Hypr2d`/`Parab2d` (`:2873`, `:2912`, `:2923`, `:2934`) do not. So `Curve2D.circleFromCenterRadius(radius: 0)` still yields a degenerate curve where `Curve2D.circle(radius: 0)` returns `nil`: the exact bug this PR fixes one dimension up.

The framing ("the contract is stated once and cannot drift apart again") is therefore only true within `OCCTBridge_Curve3D.mm`, because `static inline` in a `.mm` is per-translation-unit. Both files already `#import "OCCTBridge_Internal.h"` (`OCCTBridge_Curve3D.mm:26`, `OCCTBridge_Geom2d.mm:20`), so moving the four predicates into that header would make the claim literally true across the bridge and reduce the 2D fix to four one-line guards. If #399 is scoped to 3D that is fine, but the header move is worth doing here and the 2D gap worth filing.

### 3. `#expect` in library doc comments is a new pattern

The eight new snippets use Swift Testing macros; there are currently zero `/// #expect` occurrences anywhere in `Sources/OCCTSwift/`. The established style is plain usage with a trailing comment (`Shape.swift:21-23`: `let box = Shape.box(...)   // -> Shape? (nil on bad input)`). These snippets are consumer-facing and context7-harvested, and `#expect` needs `import Testing`, so as written they do not compile in a consumer's app. Suggested shape:

```swift
let c = Curve3D.circle(center: .zero, normal: SIMD3(0, 0, 1), radius: 5)
let degenerate = Curve3D.circle(center: .zero, normal: SIMD3(0, 0, 1), radius: 0)  // -> nil
```

### 4. Test coverage is good, with two small gaps

Both directions of each pair, zero/negative/inverted, plus sampled geometric agreement, and it closes the direct family's previously untested `hyperbola`/`parabola` zero cases. Gaps: only one center/normal pair is exercised (fine given the frame equivalence confirmed above, but a second, non-axis-aligned normal would make the parity claim self-evident rather than resting on reading OCCT's source), and `validInputsProduceMatchingGeometry` bundles four curve types into one test, so a failure does not localise. `@Test(arguments:)` or four separate tests would read better and matches usage elsewhere in the repo.

### 5. Changelog deferral

Skipping the entry to avoid six-way conflicts on the audit branch is reasonable, but this is a consumer-visible behaviour change (previously-succeeding calls now return `nil`, including the NaN case above), so the branch's merge entry should call it out explicitly rather than filing it under internal deduplication.

### Nits

- `occtValidEllipseRadii` allows `minorR == majorR`; `OCCTBridge_Surface.mm:350` uses `minorRadius >= majorRadius` for its elliptical case, a stricter contract for what looks like the same concept. Probably intentional (degenerate surface vs degenerate curve), but worth a glance while the contracts are being consolidated.
- The `docs/reference/` cross-links point at whole pages (`Curve3D.md`, `Curve3D-Analysis.md`) rather than anchors, so a reader lands at the top of a long page.
- The comment on the helper block is good: it states the invariant and the reason, which is what stops this regressing.

PR #424, opened 2026-07-27
fix(#401): route Surface.normal(u:v:) through the same degeneracy test as normal(atU:v:)

gsdali:
## What & why

`Surface` exposes the same normal twice. `normal(atU:v:)` asks OCCT — `GeomLProp_SLProps`, gated
on `IsNormalDefined()` — and returns `nil` where the tangent plane is degenerate. `normal(u:v:)`
returns a plain `SIMD3` and reports the same condition as `SIMD3(0, 0, 0)`. That difference in
*reporting* is intentional and already documented.

What was not intentional is that the two disagreed about **where** the normal is undefined.
`normal(u:v:)` hand-rolled `Geom_Surface::D1` + `gp_Vec::Crossed` and tested the cross product
against a literal `1e-15` magnitude epsilon that has no relation to the `Precision::Confusion()`
its sibling uses. `OCCTSurfaceNormal` now delegates to `OCCTSurfaceGetNormal` and zeroes its
outputs when that reports undefined, so one computation and one degeneracy test back both entry
points. The public signature and the zero-vector contract are unchanged.

Refs #401. Part of the #380 / #377 duplication audit — targets `refactor/377-segmented-audit`,
not `main`.

## Checklist

      verification)

New suite `SurfaceNormalParityTests` (`Tests/OCCTSurfaceTests`): regular-point agreement across
sphere/plane/cone, the true singular case, the near-singular regression, and the sphere pole.
Before this PR *no* test evaluated either method at or near a singularity in either direction.

## Notes for the reviewer

Two things the finding got wrong, both worth knowing before review:

**1. The divergence runs the opposite way.** The finding predicted `normal(u:v:)` would silently
return zero where `normal(atU:v:)` returns `nil`. Measured on a cone apexed at the origin:

| v | \|d1u × d1v\| | old `normal(u:v:)` | `normal(atU:v:)` |
|---|---|---|---|
| 1e-12 | 5e-13 | unit normal | defined |
| **1e-16** | **5e-17** | **(0,0,0)** | **defined** |
| 0 (apex) | 0 | (0,0,0) | `nil` |

At the true apex the two already agreed. The real failure was at `v = 1e-16`, where the old
epsilon fired on a normal OCCT resolves perfectly well — a false *"undefined"*, not a silent zero
standing in for one. The new near-apex test pins exactly that point, and asserts the recovered
normal matches a well-conditioned point on the same generatrix.

**2. A sphere pole is not a singularity here.** The finding suggested `v = ±π/2` on a sphere as
the case to test. It is a parameterisation singularity but not a normal singularity — OCCT
resolves the tangent plane there and both methods return `(0,0,±1)`. A test pins that so the next
reader does not go looking for a bug in it.

**Docs.** The finding was right that `docs/reference/Surface.md:1255` attributed `normal(u:v:)` to
`GeomLProp_SLProps::Normal` when the code used nothing of the sort. That attribution is now
accurate rather than aspirational, and both entries cross-reference each other with the
singular/non-singular examples above.

**Why not make `normal(u:v:)` optional-returning.** That would be the cleaner API, but it is a
source-breaking change to a public method for a repo that ships semver releases, and the
zero-vector contract is deliberate and documented. The defect worth fixing was the disagreement,
not the signature.

Verified: reverting only `OCCTBridge_Surface.mm` fails the near-apex test with 3 issues. Full
`swift test` (4437 tests) clean.

one.

**Conflict note:** #405 also edits `OCCTBridge_Surface.mm`, in the adjacent `v0.115` block. If
both land, expect a small textual conflict there; the changes are independent in substance.

### DIFF
--- a/Sources/OCCTBridge/src/OCCTBridge_Surface.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Surface.mm
@@ -4967,20 +4967,17 @@ const char* OCCTSurfaceTypeName(OCCTSurfaceRef surface) {
 // MARK: - v0.115: Surface additional (Normal + Curvatures)
 // --- Surface additional (new in v0.115.0) ---
 
+// Non-optional-return counterpart of OCCTSurfaceGetNormal. It keeps its own contract — a
+// zero vector where the normal is undefined, since the Swift wrapper returns a plain
+// SIMD3 — but delegates the computation so the two entry points cannot disagree about
+// *where* the normal is undefined. It used to hand-roll D1 + gp_Vec::Crossed against a
+// literal 1e-15 magnitude epsilon, which classified degeneracy differently from
+// GeomLProp_SLProps::IsNormalDefined() near a singularity (#401).
 void OCCTSurfaceNormal(OCCTSurfaceRef surface, double u, double v,
                          double* nx, double* ny, double* nz) {
+    if (!nx || !ny || !nz) return;
     *nx = *ny = *nz = 0;
-    if (!surface || surface->surface.IsNull()) return;
-    try {
-        gp_Pnt p;
-        gp_Vec d1u, d1v;
-        surface->surface->D1(u, v, p, d1u, d1v);
-        gp_Vec normal = d1u.Crossed(d1v);
-        if (normal.Magnitude() > 1e-15) {
-            normal.Normalize();
-            *nx = normal.X(); *ny = normal.Y(); *nz = normal.Z();
-        }
-    } catch (...) {}
+    OCCTSurfaceGetNormal(surface, u, v, nx, ny, nz);
 }
 
 #include <GeomLProp_SLProps.hxx>
--- a/Sources/OCCTBridge/src/OCCTBridge_Surface.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Surface.mm
@@ -4980,8 +4980,6 @@ void OCCTSurfaceNormal(OCCTSurfaceRef surface, double u, double v,
     OCCTSurfaceGetNormal(surface, u, v, nx, ny, nz);
 }
 
-#include <GeomLProp_SLProps.hxx>
-
 void OCCTSurfaceCurvatures(OCCTSurfaceRef surface, double u, double v,
                              double* gaussian, double* mean) {
     *gaussian = *mean = 0;
--- a/Sources/OCCTSwift/Surface.swift
+++ b/Sources/OCCTSwift/Surface.swift
@@ -2223,7 +2223,33 @@ extension Surface {
         return Surface(handle: ref)
     }
 
-    /// Compute the surface normal at (u, v).
+    /// Compute the surface normal at (u, v), returning the zero vector where it is undefined.
+    ///
+    /// This is the non-optional counterpart of `normal(atU:v:)`. Both evaluate the same
+    /// `GeomLProp_SLProps` normal and use the same degeneracy test, so they always agree on
+    /// *where* the normal exists — they differ only in how they report its absence: this method
+    /// returns `SIMD3(0, 0, 0)`, `normal(atU:v:)` returns `nil`. Prefer `normal(atU:v:)` when you
+    /// need to tell "undefined" apart from a genuine result.
+    ///
+    /// - Parameters:
+    ///   - u: U parameter.
+    ///   - v: V parameter.
+    /// - Returns: The unit normal, or `SIMD3(0, 0, 0)` at a singular point (a sphere pole, a cone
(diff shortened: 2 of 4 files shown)


gsdali (commented):
## Review: `fix(#401): route Surface.normal(u:v:) through the same degeneracy test as normal(atU:v:)`

### Overview

`Surface` exposes two ways to get a surface normal:
- `normal(atU:v:) -> SIMD3<Double>?` — backed by `OCCTSurfaceGetNormal`, which uses `GeomLProp_SLProps::IsNormalDefined()` (order 1, `Precision::Confusion()`) to decide degeneracy.
- `normal(u:v:) -> SIMD3<Double>` — previously hand-rolled `Geom_Surface::D1` + `gp_Vec::Crossed`, testing the cross product magnitude against a literal `1e-15`.

The two used different degeneracy tests, so near a singularity they could classify the same point differently. This PR makes `OCCTSurfaceNormal` (the C bridge function backing `normal(u:v:)`) delegate to `OCCTSurfaceGetNormal`, zeroing its outputs when the shared test reports "undefined." The public Swift signatures and the `nil`-vs-zero-vector contract are unchanged — only the underlying computation is unified.

### Correctness

I checked out the branch, built `OCCTSurfaceTests` and ran the new suite directly rather than relying solely on the PR description:

- `swift build --target OCCTSurfaceTests` — clean (only pre-existing, unrelated warnings elsewhere in the file).
- `swift test --filter SurfaceNormalParityTests` — all 4 new tests pass.

The bridge change itself is sound:
- `OCCTSurfaceGetNormal` already null-checks `s`/`s->surface`/output pointers and wraps `GeomLProp_SLProps` construction in `try/catch`, so the exception-safety and null-surface behavior of the old hand-rolled code is preserved.
- The new `OCCTSurfaceNormal` adds an output-pointer null check before dereferencing (`if (!nx || !ny || !nz) return;`) — the old code unconditionally wrote `*nx = *ny = *nz = 0` with no such guard. This is a genuine (if minor) defensive improvement, not a regression.
- When `IsNormalDefined()` is false, `OCCTSurfaceGetNormal` returns `false` without touching the output pointers, which were already zeroed by the caller — correctly reproducing the old zero-vector-on-degenerate contract.
- `OCCTSurfaceNormal` discards `OCCTSurfaceGetNormal`'s `bool` return value, but that's fine here since all the needed side effects (writing `nx/ny/nz`) already happened.

I also independently verified the "divergence runs the opposite way" claim in the PR description: at `v = 1e-16` on a cone apexed at the origin, `|d1u × d1v| ≈ 5e-17` — under the old `1e-15` threshold (spurious zero) but `IsNormalDefined()` still resolves a real normal there. The new regression test pins exactly this.

### Code quality / style

- Follows project conventions: bridge function null/try-catch pattern, Swift doc comments with fenced `swift` snippets, test suite placed in the correct per-domain target (`OCCTSurfaceTests`).
- The new comment on `OCCTSurfaceNormal` is appropriately concise and explains the *why* (prior epsilon mismatch).
- Doc updates in `docs/reference/Surface.md` correctly cross-link the two entries and fix the previously-inaccurate `GeomLProp_SLProps::Normal` attribution for `normal(u:v:)`.
- Test suite is well-targeted: regular points across plane/sphere/cone, the true apex singularity, the near-apex regression window, and the sphere-pole non-singularity case (correctly identified as *not* a normal singularity, contradicting what the original finding assumed).

### Risks / things worth flagging (minor, non-blocking)

1. **Duplicate `#include <GeomLProp_SLProps.hxx>`** — the file already includes this header at line 130; a second `#include` remains at line ~4986 (unchanged by this PR, just adjacent context in the diff). Harmless due to include guards, and not introduced by this PR, but worth cleaning up opportunistically since this PR is already touching adjacent lines. *(Will push a follow-up commit removing it.)*
2. **Stated conflict with #405** — the PR body flags that #405 touches the same file's adjacent `v0.115` block and a small textual merge conflict is expected. Not a defect, just a sequencing note — will comment on #405 to have it build off this branch/base and wait for merge to `main`.
3. No `CHANGELOG.md` entry — explicitly and correctly deferred per the audit branch's stated convention (avoiding conflicts across six sibling PRs on the same header).

### Test coverage

Solid — this is one of the more thorough test additions for a "make two things agree" fix: it doesn't just test the happy path but pins the exact regression window (`v = 1e-16`) and a case that could be mistaken for a related bug (sphere pole) but isn't. No gaps to ask for.

### Security

None — pure numerical/geometry code, no untrusted input handling changes.

### Verdict

Correct, well-tested, appropriately scoped fix that unifies degeneracy detection without changing the public contract. Approving — will push a small cleanup for item 1 above before merging.

PR #427, opened 2026-07-27
fix(#412): make Curve2D.interpolatePeriodic delegate instead of reimplementing

gsdali:
## What & why

`Curve2D.interpolatePeriodic(points:)` and `Curve2D.interpolate(through:closed:tolerance:)` wrap
the same `Geom2dAPI_Interpolate` constructor with the same `Perform()`/`IsDone()`/`Curve()`
sequence. The former is exactly the latter with `closed: true`. Kept as two independent
implementations they had already drifted twice:

1. **Tolerance unreachable.** The general path threads the caller's tolerance through; the
   periodic path hardcoded `1e-6` with no parameter to change it.
2. **Point-count floor.** The general bridge function rejects `count < 2`, the periodic one
   rejected `count < 3` — so the same 2-point input reached OCCT through one entry point and was
   refused at the bridge boundary by the other.

Both layers now have one implementation: the Swift method delegates, and the bridge's
`OCCTInterpolate2DPeriodic` forwards to `OCCTCurve2DInterpolate` so direct C consumers of the
bridge get the same behaviour rather than a second copy that can drift again.

Refs #412. Part of the #380 / #377 duplication audit — targets `refactor/377-segmented-audit`,
not `main`.

## Checklist

      verification)

New suite `Curve2DInterpolatePeriodicParityTests` (`Tests/OCCTGeom2dTests`): default-tolerance
equivalence, three non-default tolerances (previously unreachable), the 2-point floor, and the
1-point rejection both share. Worth noting how thin the prior coverage was — the issue found that
*no* existing test anywhere called `interpolate(through:closed: true)` with any tolerance, and
`interpolate2DPeriodic` only asserted `!= nil` on one square.

## Notes for the reviewer

**Two behaviour changes, both deliberate:**

`interpolatePeriodic` gains a defaulted `tolerance:` parameter. Source-compatible — existing call
sites keep compiling — and the default is exactly the value it used to hardcode.

`interpolatePeriodic(points:)` with 2 points now succeeds instead of returning `nil`. I unified on
the general entry point's floor rather than the stricter one, because OCCT genuinely accepts a
2-point periodic interpolation: it produces a valid out-and-back loop (verified — `isClosed`,
`isPeriodic`, domain `0...20` for a 10-unit segment, midpoint at the far end). The general
overload has always allowed this, so tightening instead would have removed working behaviour from
the more widely used of the two APIs to preserve an accident in the less used one.

**Delegation is safe, and this was checked rather than assumed.** For a 3-point input the two
implementations produced *bit-identical* curves before the change — same domain, same sampled
points to the last digit — which is what you would expect given they build the same
`Geom2dAPI_Interpolate` the same way. Every path except the two drifted ones is behaviour-
preserving.

**Docs** (the issue's third point): the two were documented on separate pages with no link
between them, while `Curve2D.md` already positioned `closed: true` as the recommended way to make
a periodic loop. Both pages now cross-reference, and the periodic page documents the new parameter
and the corrected minimum.

`Curve3D.interpolatePeriodic` has the same shape but is out of scope here — it is a different
type in a different audit finding.

Full `swift test` (4437 tests) clean. One unrelated intermittent failure appeared in one run —
`"Shape.loadRobust interrupts repair … (#300)"`, the CPU-timing-sensitive cancellation test whose
own source comment documents it as flaky; it also fails in isolation on occasion and passed on
re-run.

one.

**Conflict note:** #411 and #413 also edit `OCCTBridge_Geom2d.mm` and `Curve2D.swift`, in
different regions. The changes are independent in substance.

### DIFF
--- a/Sources/OCCTBridge/src/OCCTBridge_Geom2d.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Geom2d.mm
@@ -4059,20 +4059,14 @@ OCCTCurve2DRef OCCTInterpolate2DWithTangents(const double* points, int32_t count
     } catch (...) { return nullptr; }
 }
 
+// Exactly OCCTCurve2DInterpolate with the periodicity flag pinned. It used to be a second,
+// independent Geom2dAPI_Interpolate call site, which had already drifted: it rejected count < 3
+// where the general entry point rejects only count < 2, so the same 2-point input reached OCCT
+// through one and not the other. Forwarding keeps the C ABI while leaving one implementation
+// (#412). Callers that need a tolerance other than the default should call
+// OCCTCurve2DInterpolate directly with closed = true.
 OCCTCurve2DRef OCCTInterpolate2DPeriodic(const double* points, int32_t count) {
-    if (!points || count < 3) return nullptr;
-    try {
-        Handle(TColgp_HArray1OfPnt2d) pts = new TColgp_HArray1OfPnt2d(1, count);
-        for (int i = 0; i < count; i++) {
-            pts->SetValue(i + 1, gp_Pnt2d(points[i*2], points[i*2+1]));
-        }
-        Geom2dAPI_Interpolate interp(pts, Standard_True, 1e-6);
-        interp.Perform();
-        if (interp.IsDone()) {
-            return (OCCTCurve2DRef)new OCCTCurve2D{interp.Curve()};
-        }
-        return nullptr;
-    } catch (...) { return nullptr; }
+    return OCCTCurve2DInterpolate(points, count, true, 1e-6);
 }
 // --- GeomAPI_PointsToBSpline expansion ---
 // Helper: map continuity int → GeomAbs_Shape (duplicate of Curve3D's mapContinuityV115, ODR-safe)
--- a/Sources/OCCTSwift/Curve2D.swift
+++ b/Sources/OCCTSwift/Curve2D.swift
@@ -260,6 +260,23 @@ public final class Curve2D: @unchecked Sendable {
     }
 
     /// Interpolate a smooth B-spline curve through the given points.
+    ///
+    /// - Parameters:
+    ///   - points: Points to interpolate. At least 2.
+    ///   - closed: Pass `true` for a periodic loop that closes back to `points[0]` — do not repeat
+    ///     the first point at the end. `interpolatePeriodic(points:tolerance:)` is a spelling of
+    ///     this case and delegates here.
+    ///   - tolerance: Interpolation tolerance.
+    /// - Returns: The interpolated curve, or `nil` if interpolation fails.
+    ///
+    /// ```swift
+    /// let open = Curve2D.interpolate(through: [SIMD2(0, 0), SIMD2(5, 3), SIMD2(10, 0)])
+    /// #expect(open?.isPeriodic == false)
+    ///
+    /// let loop = Curve2D.interpolate(through: [SIMD2(0, 0), SIMD2(10, 0), SIMD2(5, 8)],
+    ///                                closed: true, tolerance: 1e-4)
+    /// #expect(loop?.isPeriodic == true)
+    /// ```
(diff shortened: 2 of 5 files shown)


gsdali (commented):
## Review: `fix(#412): make Curve2D.interpolatePeriodic delegate instead of reimplementing`

### Overview

Small, well-scoped fix. `Curve2D.interpolatePeriodic(points:)` previously reimplemented `Geom2dAPI_Interpolate` construction independently of `Curve2D.interpolate(through:closed:tolerance:)`, and the two had drifted:
- the periodic path hardcoded `tolerance = 1e-6` with no way to override it
- the periodic bridge function rejected `count < 3` while the general one only rejects `count < 2`

The fix makes both layers delegate to the single general implementation:
- Bridge: `OCCTInterpolate2DPeriodic` now forwards to `OCCTCurve2DInterpolate(points, count, true, 1e-6)` instead of duplicating the `Geom2dAPI_Interpolate`/`Perform`/`IsDone` sequence.
- Swift: `interpolatePeriodic(points:tolerance:)` gains a defaulted `tolerance` parameter and calls `interpolate(through:closed: true, tolerance:)` directly.
- Docs (`Curve2D.md`, `Curve2D-Analysis.md`) and a new test suite (`Curve2DInterpolatePeriodicParityTests`) are added.

### Verification performed

- Confirmed `OCCTCurve2DInterpolate`'s signature (`points, count, closed, tolerance`) matches the new forwarding call exactly, and that it's declared in `OCCTBridge.h` (included at the top of the `.mm`), so the definition-order (periodic function is defined earlier in the file, at line 4072 vs. 5258) doesn't matter.
- Checked out the branch, ran `swift build --target OCCTGeom2dTests` — clean compile, no new warnings.
- Ran `swift test --filter Curve2DInterpolatePeriodicParityTests` — all 4 new tests pass.
- Spot-checked the doc cross-reference style (`[label](OtherFile.md)`, no anchor) against the existing convention used by sibling PR #411 in the same files — consistent.

### Code quality

- The delegation is exactly the right fix for this class of bug (two independent implementations of one wrapped call, silently drifting). One-line bridge forward + one-line Swift forward is minimal and correct.
- Comments are appropriately concise (one paragraph in the `.mm`, explaining *why* forwarding exists rather than what it does) — matches project convention.
- New `///` doc comments include fenced `swift` snippets on both changed methods, satisfying the context7-harvesting doc standard.
- Test suite is a good parity/regression test: default tolerance, three non-default tolerances, the 2-point floor, and the 1-point rejection shared by both entry points. The `expectSameCurve` helper samples 11 points across the domain and compares closed/periodic flags plus domain bounds — solid for catching future re-divergence.

### Behavior changes (both flagged, both look correct)

1. **New `tolerance:` parameter on `interpolatePeriodic`**, defaulted to the prior hardcoded value — source-compatible.
2. **2-point periodic interpolation now succeeds instead of returning `nil`.** PR body says this was verified directly against OCCT (`isClosed`, `isPeriodic`, correct domain, correct midpoint) rather than assumed. Reasonable call: unifying on the less-restrictive, more-widely-used general entry point's floor, rather than propagating the accidental stricter floor.

### Minor observations (non-blocking)

- No `CHANGELOG.md`/version entry — explicitly and reasonably deferred per the PR body (this merges into an audit branch, not `main`, and 6 sibling PRs touching the same changelog header would conflict).
- The PR correctly flags overlapping-file conflicts with #411/#413 (different regions of `OCCTBridge_Geom2d.mm`/`Curve2D.swift`) — worth a rebase check before merge if those land first.
- Target branch is `refactor/377-segmented-audit`, not `main` — consistent with the rest of the #377/#380 segmented-audit series.

### Risks

None found. This is a pure consolidation with no new OCCT surface, verified bit-identical behavior for the non-drifted case (3+ points, default tolerance), and both intentional behavior changes are narrow, tested, and justified.

**Verdict: solid, low-risk fix — approve.**

PR #444, opened 2026-07-27
fix(#442): fixSolid()/solidFromShellFixed() healed only the first body of a multi-solid shape

gsdali:
Closes #442.

## The defect

`Shape.fixSolid()` and `Shape.solidFromShellFixed()` healed the **first** solid (respectively the
first shell) a `TopExp_Explorer` yielded and discarded every other body without a signal. The return
was a well-formed `Shape` that looked like a healed version of the input, so nothing downstream could
tell that most of the part was gone. The issue's own reproducer, on the pre-fix build:

| call | solids | faces | volume |
| --- | --- | --- | --- |
| input compound | 2 | 12 | 2000.0 |
| `fixSolid()` | 1 | 6 | **1000.0** |
| `solidFromShellFixed()` | 1 | 6 | **1000.0** |

## Settling the design question

#442 listed three defensible answers and did not pick one. Ground truth (`/tmp/occt_v442_test.mm`,
three programs against the pinned kernel) decided it:

- **`ShapeFix_Solid` cannot take a compound.** Its constructor and `Init` want a `TopoDS_Solid`, and
  `TopoDS::Solid` throws on anything else. So "call the other class" — the possibility the issue
  flagged as worth checking first — is not available for this tool; multi-body input has to be driven
  one solid at a time.
- **A compound result is not a new return category.** `ShapeFix_Solid::Shape()` already returns one
  when a single solid's shells resolve into several bodies (measured: one solid holding two disjoint
  shells comes back as a 2-solid compound). Callers that handled `fixSolid()` correctly before
  already handle this, which is what makes option 2 cheap.
- **`ShapeFix_Shape` does handle a multi-solid compound correctly** (2 solids, 2000 mm³) — it is
  already wrapped as `Shape.fixed(tolerance:…)`, and `fixSolid()` now cross-references it for callers
  with mixed content to preserve, since only solids survive `fixSolid()`.

Returning `nil` for multi-body input, which is what #439 did for `outerShell`, was rejected:
`outerShell` answers a question about *one* body and has no meaningful answer for several, whereas
refusing to heal a two-body part is a capability loss with no upside. That asymmetry is why #442 was
split out of #441 rather than folded into it.

## The shell-selection rule

`solidFromShellFixed()` builds one solid per **body-bounding** shell: each solid's outer shell, every
shell belonging to no solid (the usual shape of sewing output), and any further shell of a solid that
lies **outside** it — a disjoint sibling body in a multiconnex solid.

Two naive rules were measured and rejected before landing this one:

- *Every shell the explorer yields* turns a hollow solid's cavity into a second, overlapping positive
  body — a compound reading 9000 mm³ for a 7000 mm³ part.
- *Outer shell per solid* drops a body on a multiconnex solid, i.e. reintroduces the exact first-of-N
  defect this issue is about.

Cavity shells are therefore skipped, and enclosure is decided with `BRepClass3d_SolidClassifier`
rather than shell orientation — measured, **both** a hollow solid's outer and cavity shells are
`FORWARD`, so orientation carries no signal here. Ground truth over the whole input space:

```
single solid           in: shells=1 solids=1  ->  SOLID    1 solid   1000.0
bare shell             in: shells=1 solids=0  ->  SOLID    1 solid   1000.0
2-box compound         in: shells=2 solids=2  ->  COMPOUND 2 solids  2000.0
hollow solid           in: shells=2 solids=1  ->  SOLID    1 solid   8000.0   (cavity skipped)
multiconnex solid      in: shells=2 solids=1  ->  COMPOUND 2 solids  2000.0
2 free shells          in: shells=2 solids=0  ->  COMPOUND 2 solids  2000.0
solid + free shell     in: shells=2 solids=1  ->  COMPOUND 2 solids  2000.0
no shells at all       in: shells=0 solids=0  ->  NULL
```

## Behaviour change

Both calls now return a **compound** where they previously returned one arbitrary body's solid, for
multi-body input only. Single-body input is untouched, down to the returned shape type — pinned by
tests, not assumed. A caller that assumed `.solid` unconditionally should read `.solids`; a caller
that wants one specific body should pick it before healing.

Also fixed: `OCCTShapeFixSolid` gains the `if (!shape) return nullptr;` guard its siblings in the
file use (issue's secondary note). Not reachable through the Swift API, but a null deref is an
uncatchable SIGSEGV rather than something the enclosing `try` would catch.

## Verification

New suite `Issue442FixSolidMultiBody` (10 tests). **Four fail on the pre-fix bridge** with exactly
the issue's numbers — 1 solid / 6 faces / 1000.0 against a 2 solid / 12 face / 2000.0 input, bounds
stopping at x=10 instead of x=30 — confirmed by rebuilding against the pre-fix `OCCTBridge_Healing.mm`
and re-running. The other six pin the single-body, hollow-solid, free-shell and no-solid behaviour
that must **not** change.

Full `swift test`: **4462 tests in 1290 suites, all passing.**

Bridge-only — no OCCT kernel change, no `OCCT.xcframework` rebuild. Operation count unchanged at
4,258 (`Scripts/count-operations.py`): this is behaviour and documentation only.

## The audit note

#442 asked for the first-of-N idiom to be grepped for across the bridge before closing. Done, and
filed as **#443**: 23 candidate sites, 8 false positives, 5 where "first" is the documented contract,
10 undocumented silent picks. Three are confirmed by measurement to lose whole bodies —
`Shape.solid(from:)`, `Shape.upgraded()`, and `AssemblyNode.setTriangulationFromShape` (which stores
one face's triangulation for any input, measured at 4 nodes for both a 6-face box and a 12-face
compound). `Shape.solid(from:)` is the sharpest: its doc names sewing output as the expected input,
and after this PR the two sibling entry points disagree on it — `sewn.solidFromShellFixed()` gives
2 solids / 2000 mm³ where `Shape.solid(from: sewn)` gives 1 / 1000. Left out of this PR deliberately;
each is a separate contract decision, the same reason #442 was split out of #441.

## Docs

Swift doc comments (both now carry a runnable ```swift``` snippet), `OCCTBridge.h`, the generated
reference (`docs/reference/Document-OCAF-Attributes.md`) and `docs/CHANGELOG.md` all state the same
rule. The CHANGELOG entry is deliberately left unversioned for whoever tags.

### DIFF
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -12186,10 +12186,14 @@ bool OCCTDocumentHasRelation(OCCTDocumentRef _Nonnull document, int tag);
 
 // MARK: - ShapeFix_Solid
 
-/// Fix a solid shape (topology and orientation). Returns fixed shape or NULL.
+/// Fix a solid shape (topology and orientation). Heals EVERY solid in the input, not
+/// just the first (#442): a single body comes back as a solid, several as a compound.
+/// Returns fixed shape or NULL when the input holds no solid.
 OCCTShapeRef _Nullable OCCTShapeFixSolid(OCCTShapeRef _Nonnull shape);
 
-/// Create a solid from a shell using ShapeFix_Solid
+/// Create a solid from a shell using ShapeFix_Solid. One solid per body-bounding shell
+/// (#442) — each solid's outer shell plus every free shell; a solid's cavity shells are
+/// skipped. A single body comes back as a solid, several as a compound.
 OCCTShapeRef _Nullable OCCTShapeSolidFromShell(OCCTShapeRef _Nonnull shellShape);
 
 // MARK: - ShapeFix_EdgeConnect
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -12187,13 +12187,16 @@ bool OCCTDocumentHasRelation(OCCTDocumentRef _Nonnull document, int tag);
 // MARK: - ShapeFix_Solid
 
 /// Fix a solid shape (topology and orientation). Heals EVERY solid in the input, not
-/// just the first (#442): a single body comes back as a solid, several as a compound.
-/// Returns fixed shape or NULL when the input holds no solid.
+/// just the first (#442): a single body comes back as a solid, several as a compound of
+/// one result per input body. No body is ever dropped, so a result is not always a solid
+/// — ShapeFix_Solid returns a SHELL it could not close, and a solid it fails to heal
+/// comes back unhealed. Returns NULL only when the input holds no solid.
 OCCTShapeRef _Nullable OCCTShapeFixSolid(OCCTShapeRef _Nonnull shape);
 
 /// Create a solid from a shell using ShapeFix_Solid. One solid per body-bounding shell
-/// (#442) — each solid's outer shell plus every free shell; a solid's cavity shells are
-/// skipped. A single body comes back as a solid, several as a compound.
+/// (#442) — within each solid, every shell an even number of the others enclose, plus
+/// every free shell; a solid's cavity shells are skipped. A single body comes back as a
+/// solid, several as a compound.
 OCCTShapeRef _Nullable OCCTShapeSolidFromShell(OCCTShapeRef _Nonnull shellShape);
 
 // MARK: - ShapeFix_EdgeConnect
--- a/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
@@ -3251,6 +3251,21 @@ static std::vector<TopoDS_Shell> occtBodyBoundingShells(const TopoDS_Shape& shap
         // a body nested inside another body's cavity correctly: enclosed twice, so even.
         std::vector<int> enclosedBy(shells.size(), 0);
         for (size_t i = 0; i < shells.size(); i++) {
+            // An open shell cannot enclose anything, and every shell is a reference under
+            // parity, so letting one classify would add a spurious ±1 to the others and
+            // flip their verdicts. Measured on {A_outer, A_cavity, openShell wrapping both}:
+            // without this the outer shell is DROPPED (count 1) and the cavity emitted as a
+            // body. BRep_Tool::IsClosed on a shell is a real edge-pairing check rather than
+            // the Closed() flag, so a genuine cavity shell still qualifies as a reference.
+            // Open shells reach here routinely — this function accepts them by contract.
+            //
+            // Deliberate trade-off, do not "fix" a double-count back out of this: on a body
+            // whose OUTER shell is the open one, the cavity becomes the only eligible
+            // reference, both end even, and the cavity is emitted as a positive body. Input
+            // that broken has no right answer, and an extra body beats a dropped one for a
+            // call whose whole contract is that bodies do not vanish silently.
+            if (!BRep_Tool::IsClosed(shells[i])) continue;
+
             // Reference built directly rather than through ShapeFix_Solid::SolidFromShell:
             // that call does sh.Free(true), a TShape-level write to state shared with the
             // caller's shape, and nothing here needs the reorientation it pays that for.
--- a/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
@@ -3155,15 +3155,45 @@ OCCTShapeRef _Nullable OCCTShapeFixComposeShell(OCCTShapeRef _Nonnull faceRef, d
 // MARK: - ShapeFix_Solid
 
 #include <ShapeFix_Solid.hxx>
+#include <BRepClass3d.hxx>
+#include <BRepClass3d_SolidClassifier.hxx>
+#include <Precision.hxx>
 
+// Assemble one result shape from healed bodies: the body itself when there is exactly
+// one, a compound when there are several. ShapeFix_Solid::Shape() already returns a
+// compound for a multiconnex input, so a compound result is nothing new for callers.
+static TopoDS_Shape occtSolidBodiesToShape(const std::vector<TopoDS_Shape>& bodies) {
+    if (bodies.empty()) return TopoDS_Shape();
+    if (bodies.size() == 1) return bodies[0];
+    TopoDS_Compound compound;
+    BRep_Builder builder;
+    builder.MakeCompound(compound);
+    for (const TopoDS_Shape& body : bodies) builder.Add(compound, body);
+    return compound;
+}
+
+// Heal EVERY solid, not just the first one an explorer yields (#442). ShapeFix_Solid
+// cannot take a compound — Init/the constructor want a TopoDS_Solid, and TopoDS::Solid
+// throws on anything else — so multi-body input has to be driven one solid at a time.
 OCCTShapeRef OCCTShapeFixSolid(OCCTShapeRef shape) {
+    if (!shape) return nullptr;
     try {
-        TopExp_Explorer exp(shape->shape, TopAbs_SOLID);
-        if (!exp.More()) return nullptr;
-        TopoDS_Solid solid = TopoDS::Solid(exp.Current());
-        ShapeFix_Solid fixer(solid);
-        fixer.Perform();
-        TopoDS_Shape result = fixer.Shape();
+        std::vector<TopoDS_Shape> fixed;
+        for (TopExp_Explorer exp(shape->shape, TopAbs_SOLID); exp.More(); exp.Next()) {
+            ShapeFix_Solid fixer(TopoDS::Solid(exp.Current()));
+            fixer.Perform();
+            TopoDS_Shape result = fixer.Shape();
+            if (result.IsNull()) continue;
+            // Flatten: Shape() hands back a compound when one solid's shells resolve
+            // into several, and nesting those would hide bodies a level down.
+            if (result.ShapeType() == TopAbs_COMPOUND || result.ShapeType() == TopAbs_COMPSOLID) {
+                for (TopExp_Explorer se(result, TopAbs_SOLID); se.More(); se.Next())
+                    fixed.push_back(se.Current());
+            } else {
+                fixed.push_back(result);
+            }
+        }
+        TopoDS_Shape result = occtSolidBodiesToShape(fixed);
         if (result.IsNull()) return nullptr;
         auto* ref = new OCCTShape();
         ref->shape = result;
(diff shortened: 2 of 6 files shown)


gsdali (commented):
## Review

### Overview

`OCCTShapeFixSolid` and `OCCTShapeSolidFromShell` each took the first shape a `TopExp_Explorer` yielded and dropped the rest, returning a well-formed `Shape` that looked healed. The PR iterates all bodies and assembles a compound when there is more than one, adds a `nullptr` guard, and documents the new rule in four places. `solidFromShellFixed()` gains a real selection rule (outer shell per solid, plus disjoint siblings, plus free shells, cavities skipped) decided with `BRepClass3d_SolidClassifier`.

The design work is strong. `ShapeFix_Solid` genuinely cannot take a compound (`Init`/ctor want `TopoDS_Solid`), `BRepClass3d::OuterShell` genuinely returns a shell that classifies the infinite point as `OUT` so the cavity/outer discrimination is sound, and the "both shells are FORWARD" measurement that ruled out orientation checks out against the pinned source. Rejecting the `nil` answer that #439 used is right, and that asymmetry is well argued in the PR body. Behaviour-change documentation and the unversioned CHANGELOG both follow project convention.

### Correctness

- **`fixSolid()` still silently drops a body, on the failure path.** `if (result.IsNull()) continue;` (`OCCTBridge_Healing.mm`) means a solid whose `ShapeFix_Solid::Perform()` produces nothing vanishes from the compound with no signal. That is the exact failure mode this PR exists to eliminate, just relocated. Suggest pushing the original `exp.Current()` as a fallback, or returning `nullptr` so the caller sees the failure. `OCCTShapeSolidFromShell`'s `if (!solid.IsNull()) made.push_back(solid);` has the same shape.

- **Degenerate fallback can double-count.** `if (outer.IsNull()) outer = shells[0];`. `BRepClass3d::OuterShell` returns null when no shell classifies the infinite point as `OUT`. `shells[0]` may then be a cavity; the true outer shell classifies `OUT` of the cavity-derived reference solid and gets added too, giving a compound that double-counts the part. Only reachable on already-invalid input, but a `continue` (or at least a comment saying the guess is deliberate) reads better than silently guessing.

- **Free-shell loop can emit duplicates.** The final `TopExp_Explorer(shape, TopAbs_SHELL)` pass never calls `claimed.Add(...)`, so a compound holding the same free shell twice yields two identical solids. One-line fix.

- **COMPSOLID branch discards what it detects.** `fixSolid` explicitly handles `TopAbs_COMPSOLID` from `fixer.Shape()`, then flattens it into a plain compound, losing the connectivity typing. Either preserve it or drop the branch.

- **`SolidFromShell` writes to the caller's shape.** `ShapeFix_Solid::SolidFromShell` does `TopoDS_Shell sh = shell; if (!sh.Free()) sh.Free(true);`. `Free()` is a *TShape-level* flag, so this mutates state shared with the input `Shape`. Pre-existing, but the PR now calls it on every body-bounding shell plus one extra reference build, widening an unsynchronized write to shared shape state in a bridge that locks carefully elsewhere. Worth a note in the source comment at minimum.

### Test coverage, the main thing I would fix before merge

The suite is well chosen (single-body pinning, hollow, free shells, no-solid, the issue's table), and pre-verifying four tests against the pre-fix bridge is exactly right. But the assertions are guarded in a way that lets them skip:

- **Every volume assertion sits inside `if let volume = ... { #expect(...) }` with no `else`.** `Shape.volume` is `v >= 0 ? v : nil` (`Shape.swift:2408`), so it returns `nil` precisely when the solid comes back **inverted**. Orientation is the entire job of `SolidFromShell`. A regression that returns reversed solids therefore makes `solidFromShellSingleBody`, `solidFromShellMultiBody`, `solidFromShellSkipsCavity` and `reproducerTable` all pass with their core assertion silently skipped. Same for `if let box = shape.boundingBox` in `reproducerTable`.
- **`solidFromShellNoShell`'s only assertion is inside `if let face = ...`**, so an empty face list makes it vacuous. (`fixSolidNoSolid` is protected by its preceding `#expect(faces.count == 6)`.)
- Every fixture `guard let ... else { return }` passes silently.

PR #441, merged one commit before this and in the same lineage, was corrected for exactly this in its third review commit ("a nil binding skipped them and the test still passed... which is the failure mode this whole issue is about") and uses `#expect(Bool(false), "…")` in every `guard`-`else` and if-let-else. `Issue442FixSolidMultiBodyTests.swift` should match that. CLAUDE.md's "never force-unwrap in `#expect`" rule is about avoiding `result!`, not about permitting a skipped assertion.

Missing case worth adding: the **multiconnex solid** (one solid, two disjoint shells). It is in the PR's own ground-truth table and it is the case that rules out the naive "outer shell per solid" rule, so it is the one most likely to regress unnoticed.

### Documentation

- **Wrong link target.** `docs/reference/Document-OCAF-Attributes.md`: `[Shape.solidFromShells(_:)](Shape-Features.md)`. That symbol is documented in `docs/reference/Shape-Measurement.md:739`, not `Shape-Features.md`, and the link carries no anchor. Should be `Shape-Measurement.md#solidfromshells_`. (The `fixed(tolerance:fixSolid:fixShell:fixFace:fixWire:)` link and its anchor do check out, as does the DocC path in the Swift doc comment.)

- **The `nil` contract is overstated.** Both the Swift doc and the reference page now say `solidFromShellFixed()` returns `nil` "if the receiver holds no shell **or none of them closes**". Reading `ShapeFix_Solid.cxx`, `SolidFromShell` does `B.MakeSolid(solid); B.Add(solid, sh);` unconditionally before any classification, and returns that even on its exception path. It never returns a null solid and never rejects an open shell. So the only `nil` is "no shells at all"; an unclosed shell comes back as a bogus "solid". The pre-fix one-liner had the same inaccuracy, but it is being restated and expanded in three places here, so this is the moment to correct it.

Otherwise the docs are good: runnable ```swift``` snippets on both APIs per the context7 rule, header/reference/CHANGELOG all consistent, README correctly untouched since the op count is unchanged.

### Performance

`occtShellIsInsideSolid` constructs a fresh `BRepClass3d_SolidClassifier(reference)` on every call, and it is called once per non-outer shell. The constructor's `Load()` builds a `BRepClass3d_SolidExplorer` over all faces of the outer solid, so a solid with N cavities (a lattice or foam part) pays N full loads of the same solid. Hoisting the classifier out to `occtBodyBoundingShells`, one per solid, passed by reference, turns that into O(N) ray casts instead of O(N) loads.

Separately, `maker.SolidFromShell(outer)` builds the classification reference with a `PerformInfinitePoint`, and then `OCCTShapeSolidFromShell` calls `SolidFromShell` on the same shell again. The reference solid could be built with a plain `BRep_Builder::MakeSolid` + `Add` instead, which also sidesteps the `Free()` mutation noted above.

### Style and conventions

Clean. Helpers follow the file's `occt*` static-helper idiom, `try`/`catch(...)` wrapping matches, includes are all present (`<vector>`, `TopTools_IndexedMapOfShape`, `BRep_Builder`, `TopoDS_Compound` were already in the file), and `TopTools_IndexedMapOfShape` is the orientation-insensitive `IsSame` map, which is the right choice for `claimed`. Minor: the new code uses the file's older `auto* ref = new OCCTShape(); ref->shape = ...` form where #441 used `new OCCTShape(shell)`, consistent with its neighbours, so fine either way.

### Security

No concerns. No I/O, no allocation sized from input, no new global state. The added `if (!shape) return nullptr;` removes an unguarded deref.

---

**Verdict:** ship-worthy fix with sound design. I would hold for two things: the test guards (the volume-is-nil-on-inversion gap is a real hole in exactly the property being fixed) and the `Shape-Features.md` link. The silent-drop-on-heal-failure and the classifier hoist are worth doing in the same pass; the rest are notes.

gsdali (commented):
## Review (round 2), commit `eb6ee99a`

I checked every load-bearing kernel claim in the follow-up commit against the pinned p1 sources. They hold up, including one the commit found that I had missed.

### Previous round: resolved

- **Silent drop on the failure path.** Fixed. An unhealable solid now comes back unhealed via `fixed.push_back(exp.Current())`, and the empty-compound case too.
- **Flatten dropping unclosed shells.** This goes beyond what I reported and the reasoning is correct. `ShapeFix_Solid.cxx:626-632`: when `aResShape` is a `TopAbs_SHELL` and `myCreateOpenSolidMode` is false, it sends `FixAdvSolid.FixShell.MSG10` and then `aB.Add(aComp, aResSh)` adds the **shell** to the compound unchanged. A `TopExp_Explorer(TopAbs_SOLID)` would indeed have dropped exactly that. `TopoDS_Iterator` is right.
- **COMPSOLID branch removed.** Justified. `Shape()` returns `myShape`, which is only ever the input solid, one fixed solid, or a `MakeCompound` compound.
- **Free-shell dedup.** Correct, and the `Add`-returns-an-index catch is a good one. Ground-truthing it rather than trusting the one-liner is what caught the 4000 mm³ double-count.
- **Classifier hoist.** Safe to reuse across `PerformInfinitePoint` then `Perform`. `MARCHEPASSIUNESEULEFACE` is `0`, so `Perform` compiles to a bare `BRepClass3d_SClassifier::Perform(explorer, P, Tol)` and never consults the `isaholeinspace` that `PerformInfinitePoint` sets. The `insideState` inversion is sound, and defaulting `ON`/`UNKNOWN` to `TopAbs_IN` is a sensible read.
- **`SolidFromShell`'s `sh.Free(true)` write.** Avoided for the reference solid, and the remaining unavoidable calls are documented in the source.
- **Tests.** All skippable assertions closed. `expectVolume` with `sourceLocation` threading is the right shape, every `guard`-`else` fails loudly, labels on expectations, multiconnex and duplicate-shell cases added.
- **Docs.** Link now resolves (`Shape-Measurement.md#solidfromshells_` matches the `### \`solidFromShells(_:)\`` heading), and the `nil` contract is corrected.

### Remaining

**1. The `fixSolid()` docs no longer describe what it returns.** This commit introduced two result categories and documented neither:

- The compound can contain an **unclosed shell**. That is the whole point of the `TopoDS_Iterator` change, per its own source comment.
- On heal failure it now contains the **original unhealed solid**.

But the Swift doc still says "a compound of the healed solids", "**Only solids are carried over**", and "- Returns: The repaired solid, a compound of repaired solids...". `OCCTBridge.h` and the reference page say the same. A caller reading `healed.solids.count` under-counts when a shell comes back, with no signal, which is the exact failure class this PR exists to remove. Worth two sentences in all three places, plus a note that a body may be present but unhealed.

**2. The bounding-box rule trades one degenerate case for another.** The inside-out case it fixes is real and well measured. But "only the enclosing shell can have the largest box" assumes exactly one shell encloses all the rest, which a multiconnex solid breaks. Shells `{A_outer, A_cavity, B_outer}` where `B_outer` is the widest:

- `outer` = `B_outer`; reference is built from it.
- `A_outer` classifies OUT of B, added. Correct.
- `A_cavity` also classifies OUT of B, added, so it is **emitted as a positive body**, double-counting.

The old `BRepClass3d::OuterShell` got this one right: whichever body's outer shell it named, that solid's own cavity classified IN of it. So this is a swap, not a strict win. The fully correct rule is enclosure parity: for each shell, count how many *other* shells enclose it; it is a body iff that count is even. That is O(N²) classifications, but N is 1-3 on any real input and you already build one classifier per reference. If that is more than you want in this PR, the assumption should at least be stated in the source comment and the CHANGELOG, since "multiconnex" is otherwise advertised as a supported case.

**3. `occtShellExtent` is called 2N times, N of them on the incumbent.**

```cpp
for (const TopoDS_Shell& candidate : shells)
    if (occtShellExtent(candidate) > occtShellExtent(outer)) outer = candidate;
```

`BRepBndLib::Add` walks every face of the shell, and the incumbent is by construction the widest, so usually the one with the most faces, rebuilt on every iteration. Caching it in a local `double` alongside `outer` makes it N calls. Small, but it undercuts the commit's own perf section.

**4. Nits.**

- `occtShellExtent`'s comment says "Diagonal of a shell's bounding box"; `SquareExtent()` returns the *squared* diagonal, and folds `Gap` in twice per axis. Comparison-only, so no bug, but the comment should match.
- `OCCTBridge.h` still says "each solid's **outer** shell" where the CHANGELOG and reference page were updated to "enclosing shell", and the CHANGELOG's closing line asserts `OCCTBridge.h` "state[s] the same rule". The header was not touched in this commit.

### Not re-verified

I read the sources rather than rebuilding, so the "13 tests, six fail on the pre-fix bridge" and full-suite claims are taken as reported. The new code type-checks on inspection: the `TopoDS_Shape(shell) : TopoDS_Shape(solid)` ternary slices cleanly, includes are present and not duplicated, and `BRepClass3d.hxx` is cleanly removed, its only remaining mention being in a comment.

---

**Verdict:** substantially stronger than round 1. Item 1 is the one I would fix before merge, since it re-opens a silent-undercount path in the documentation of a PR about silent undercounts. Item 2 is a judgement call worth stating explicitly either way; 3 and 4 are cleanup.

gsdali (commented):
## Review (round 3), commit `38cec885`

### Round 2: resolved

- **Enclosure parity** implemented and correct. I traced it over the input space: hollow `{outer, cavity}` (cavity enclosed once, odd, skipped); multiconnex `{A, B}` (both zero, both bodies); the counter-case `{A_outer, A_cavity, B_outer}` (A_cavity enclosed once by A_outer regardless of B being wider, so 8000 + 27000 = 35000); and a body nested inside another body's cavity (enclosed twice, even, kept). The reply is right that this is not a return to `OuterShell` — parity has no reference to get wrong. Good regression test for the exact counter-case.
- **`insideState` per reference** rather than once per solid is required for parity, and the change makes it so.
- **Single-shell early-out** means the overwhelmingly common case still pays zero classifications.
- **`occtShellExtent` gone**, includes cleaned, header now states the parity rule.
- **`claimed` still populated from all of a solid's shells**, so cavities cannot leak back in via the free-shell pass. Worth confirming since the selection changed and the claiming did not.

One thing I checked because it looked dangerous and turns out to be safe: a **face-less reference shell**. `BRepClass3d_SClassifier::Perform` and `PerformInfinitePoint` both set `myState = 3` on `Reject` ("solid without faces => the whole space"), and `State()` maps `3` to `TopAbs_IN`. So `insideState` becomes `OUT`, every subsequent reading is `IN`, and the comparison is false throughout. The degenerate reference contributes **0** to every count rather than **+1 to all of them**, which would have flipped every shell's parity at once. The inversion neutralizes it by accident, but correctly.

### Remaining

**1. The new `fixSolid()` doc advice does not work.** Both the Swift doc and the reference page now say to "check `result.subShapes(ofType: .shell)`" to tell whether a body came back unclosed. `subShapeCount` is `TopExp::MapShapes(shape, TopAbs_SHELL, map)` (`OCCTBridge_Topology.mm:1005-1014`), which maps at every depth, so a compound of two perfectly healed solids reports **2** shells, not 0. The check is non-empty in the healthy case and cannot distinguish it from the failure case. This suite demonstrates the same depth behaviour on itself: `#expect(hollow.shells.count == 2)` for a single hollow *solid*.

The direct-children walk is what is wanted, and the API already exists: `Shape.child(at:)` (`Document.swift:11313`, backed by `TopoDS_Iterator`). So either "iterate `child(at:)` and check `shapeType == .shell`", or the simpler "compare `result.solids.count` against the number of input bodies". Same sentence appears in `Document.swift` and `docs/reference/Document-OCAF-Attributes.md`.

**2. An open shell as a parity reference gives an arbitrary verdict, and now perturbs every other shell.** This is a live path, not hypothetical: this same commit series documents that `solidFromShellFixed()` accepts open shells and returns them as unclosed solids, so open shells reach `occtBodyBoundingShells`. Under the widest-shell rule a bad reference mattered only if it was the one chosen; under parity **every** shell is a reference, so one open shell can add a spurious plus or minus one to other shells' counts and flip a body to a cavity (dropped) or a cavity to a body (double-counted).

Cheap hardening: an open shell cannot enclose anything, so skip it in the reference role only:

```cpp
if (!BRep_Tool::IsClosed(shells[i])) continue;   // cannot enclose anything
```

before building the reference solid. Candidates still get classified normally; it is only the reference that needs closure. One traversal, removes the class.

**3. Nit.** Parity builds N classifiers where the previous rule built 1, so 2 for an ordinary hollow solid. Irrelevant at N of 1-3, and the source comment already flags that each constructor loads every face, so this is just to note the cost is N *loads* plus N(N-1) `Perform` calls, not one load.

### Not re-verified

Read rather than rebuilt, so "4466 tests in 1290 suites" and the ground-truth re-run are taken as reported. New code type-checks on inspection; `Bnd_Box`/`BRepBndLib` are cleanly removed with no orphan references.

---

**Verdict:** the selection rule is now correct on every topology I can construct, and the ordering claim ("exploration order") is finally true of the code. Item 1 is the one to fix before merge: it is the third round in a row where the sentence added to close a silent-undercount gap contains its own, and it ships advice that will read as a failure on healthy output. Item 2 is worth the one line. Item 3 is informational.

PR #447, opened 2026-07-27
fix(#443): three more first-of-N explorer sites dropped most of their input

gsdali:
Closes #443.

#442's audit note asked for the first-of-N `TopExp_Explorer` idiom to be swept across the whole
bridge before closing it. The sweep found 23 candidates: 8 false positives, 5 where "first" is the
documented contract, and 10 undocumented silent picks. Three of those lose most of their input and
are fixed here; the other seven are documented rather than changed, since each is singular by
contract and widening it would change what its arguments mean.

## Measured, against a compound of two disjoint 10 mm boxes (2 solids, 12 faces, 2000 mm³)

| call | before | after |
| --- | --- | --- |
| `Shape.solid(from:)` | 1 solid, 1000 mm³ | 2 solids, 2000 mm³ |
| `Shape.upgraded()` | 1 solid, 1000 mm³ | 2 solids, 2000 mm³ |
| `AssemblyNode.setTriangulationFromShape` | 4 nodes, 2 triangles | 48 nodes, 24 triangles |

## `Shape.solid(from:)` and `Shape.solidWithFullHistory(from:)`

The sharpest of the three: the doc names sewing output as the expected input, and sewing two bodies
yields exactly the two-shell input it mishandled. After #442 the two sibling entry points disagreed
on that same shape, `sewn.solidFromShellFixed()` giving 2 solids / 2000 mm³ and
`Shape.solid(from: sewn)` giving 1 / 1000. Both now go through #442's `occtBodyBoundingShells`,
which moved to `OCCTBridge_Internal.h` for it, so they agree by construction.

The history variant shares **one** `ShapeBuild_ReShape` across the per-body `ShapeFix_Solid` runs.
`BRepTools_ReShape::History()` builds a fresh `BRepTools_History` from the context's whole
replacement map on each call (read in `occt-src`, confirmed on a probe), so the single history still
covers every body rather than only the last one built.

## `Shape.upgraded()`

The call most likely to be pointed at a raw imported mesh, where multi-body input is the norm. Its
solid step now builds one solid per body-bounding shell. Two limits inherent to sewing first are now
documented rather than silent:

- sewing dissolves the input's solids, so a hollow body's **cavity is filled** (8000 mm³ for a
  7000 mm³ hollow cube, unchanged from before but never stated);
- the solid step replaces the sewn shape rather than merging into it, so content sewing could not
  attach to a shell is not carried through.

`fixed(tolerance:)` is the call for either case, and is now cross-referenced.

## `AssemblyNode.setTriangulationFromShape`

Meshed the whole shape and stored only the **first face's** triangulation, so a 6-face box and a
12-face compound both stored 4 nodes and 2 triangles for a doc comment reading "by meshing a shape".
Arguably the worst of the three, since the attribute is what later readers trust as the label's
geometry. It now merges every meshed face into one `Poly_Triangulation` in the shape's own
coordinate system: per-face locations applied to the nodes, reversed faces' winding and node normals
flipped so the result is consistently outward (the same rules `Shape.mesh()` applies), the worst
contributing face's deflection carried over, node normals kept only if every face has them, and
per-face UV nodes dropped since they index parameter spaces that stop meaning anything once the
faces are pooled.

The deflection part was caught by the *existing* `Triangulation deflection` test, not by a new one: a
hand-built `Poly_Triangulation` starts at 0, and a 0 there would read as "exact".

## A latent case in #442's own helper, found and fixed with them

`occtBodyBoundingShells` ran its enclosure-parity pass within each solid but added every shell
belonging to *no* solid unconditionally, on the reasoning that a free shell has no declared cavity
relationship. But **sewing dissolves the solid that carried that declaration**, and sewing is the
ordinary way all of these calls are reached. Measured:

| input | before | after |
| --- | --- | --- |
| sewn hollow box | 2 bodies (cavity emitted) | 1 body, 8000 mm³ |
| `{hollow body, body inside its cavity}`, sewn | 3 bodies, 6784 mm³ | 2 bodies, 8512 mm³ |
| hollow solid, not sewn | 1 body, 8000 mm³ | unchanged |
| two disjoint free shells | 2 bodies, 2000 mm³ | unchanged |
| same free shell twice | 1 body, 1000 mm³ | unchanged |

Free shells are now one further group through the same parity pass, so a hollow body reads the same
whether or not it has been sewn. Containment among free shells is geometric rather than declared, so
a closed shell sitting alone inside another is read as that one's cavity: the same reading OCCT's own
solid convention gives it, and the only one available without a declaration. A body nested inside a
cavity is enclosed twice, so it is still a body. None of #442's shipped cases change.

> **Behaviour change for consumers.** `Shape.solid(from:)`, `Shape.solidWithFullHistory(from:)` and
> `Shape.upgraded()` return a **compound** where they previously returned one arbitrary body's solid,
> for multi-body input only. Single-body input is untouched, down to the returned shape type.
> `Shape.solidFromShellFixed()` returns **one** body where it returned two for a sewn hollow part,
> the second having been the cavity.

## The seven documented, not changed

Each gets a `- Note:` on the Swift wrapper and a `#443 audit:` comment on the bridge function saying
why it stays singular: `MedialAxis.init(of:)` (a medial axis is a property of one face, and the
result type holds one graph), `Shape.halfSpace(face:referencePoint:)` (a half-space is bounded by one
face by definition), `Shape.fillet2D(vertexIndices:radii:)` and `chamfer2D(edgePairs:distances:)`
(the indices are numbered within the chosen face, so covering every face needs per-face index lists
and a different signature), `Shape.splitByWireOnFace(_:faceIndex:)` and `locOpeSplit(wiresOnFaces:)`
(the pair list is already where several wires are named), and `Shape.solidFromShells(_:)` (the
argument order *is* the outer-versus-cavity contract). `OCCTShapeBuildThreadCutter` is internal and
single-body by construction. `Shape.faceRestricted(by:)` and `Wire.offset(by:joinType:)` already
stated what they do and are untouched.

## Review round 2

Rebased onto `4bdc30a` (v1.16.0). The new entry sits above the now-versioned v1.16.0 block and stays
unversioned.

**The free-shell parity group was O(N²) on this PR's own primary input.** Inside a solid N is 1-3,
which is what the inherited comment assumed; sewing a raw imported mesh yields hundreds of disjoint
shells. A conservative bounding-box pre-filter now prunes pairs before any ray cast and skips building
a classifier for a reference that overlaps nothing. Enclosure implies box containment, so no verdict
changes:

| N disjoint shells | without pre-filter | with |
| --- | --- | --- |
| 50 | 15.8 ms | 0.2 ms |
| 100 | 44.7 ms | 0.3 ms |
| 200 | 160.4 ms | 0.7 ms |

Identical results on `{outer, cavity}` (1), `{outer, cavity, nested body}` (2) and `{two touching
boxes}` (2). Two new tests cover it: 100 disjoint free shells, and two touching bodies whose boxes
overlap without enclosure, so the cheap test cannot decide them and the ray cast must still run.

**`setTriangulationFromShape` also changed coordinate frames, and that was undisclosed.** The old code
fetched each face's `TopLoc_Location` and discarded it. Measured: a box moved to (100, 200, 300) stored
node 1 at the origin, and now stores it at (100, 200, 300). Now stated in the doc comment, the
CHANGELOG behaviour-change block, `docs/reference/Document.md` and above.

**The `OCCTShapeCreateMesh` parity claim was wrong for normals.** That function pushes
`triangulation->Normal(i)` raw, with neither the location transform nor the reversal, so only its
winding swap matches. The comment is corrected, and the consequence (`Mesh.normals` disagrees with this
attribute for a located or reversed face) is recorded as `Shape.mesh()`'s question rather than papered
over.

**The normals branch is unreachable via `BRepMesh_IncrementalMesh`**, which produces no node normals at
all: 0 of 6 box faces, 0 of 1 sphere face. It *is* reachable through glTF import, so it is now tested
rather than only reasoned about: a sphere written to GLB and read back stores unit normals that point
outward, and the same sphere meshed from B-Rep stores none.

**Two accessors added, because none of the above was checkable.** The attribute exposed only counts and
deflection, so nothing in the Swift API could read the coordinates it stored, which is part of why both
defects survived. `AssemblyNode.triangulationNode(at:)` and `triangulationNormal(at:)`: 2 operations,
counts refreshed with `Scripts/count-operations.py --fix` (4,258 to 4,260).

New fixtures for the coverage gap: a located box (nodes must fall inside the shape's own bounds) and a
mirrored box (body at negative x, every face still stored).

Both minor items applied: the shared-`ReShape` comment now notes that each body's `Perform` applies
against a map already holding the earlier bodies' replacements, and the parity comment carries the cost
note for the unbounded free-shell group.

## Review round 4

`OCCTShapeUpgrade`'s own per-shell `MakeSolid` loop — touched by this same PR — had the identical
plain-`continue` drop that round 3 fixed on `Shape.solid(from:)`/`solidWithFullHistory(from:)`, missed
because it wasn't one of the two functions round 3's finding was about. Fixed the same way
(`push_back` the unfixed shell rather than dropping it), same "dead code today, fixed for symmetry"
status. New test: `upgraded() keeps an unclosable shell's faces rather than dropping them` — asserts
on face count, not solid count, since `upgraded()`'s later `ShapeFix_Shape` pass reclassifies the kept
body so it no longer counts as a `TopAbs_SOLID`, unlike the round-3 siblings.

Also: three existing `docs/reference/` pages (`Shape-Measurement.md`, `Shape-Builders-1.md`,
`Shape-Builders-2.md`) covering methods this PR documents-not-changes had the new caveat in the Swift
doc comment but not mirrored into the page; now they match. And `upgraded()`'s doc comment said "free
shells each become a body," which the free-shell parity fix in this same PR made untrue — reworded to
match the correct rule the very next line already stated.

## Verification

Ground truth in C++ first for all three defects and all three fixes, and again for every claim in the
review round (the pre-filter cost and verdicts, the frame change, the normals branch reachability),
then `swift test` green at **4,511 tests** (1294 suites). 31 new tests: 22 in
`Tests/OCCTShapeHealingTests/Issue443FirstOfNTests.swift`, 9 plus one extended existing test in
`Tests/OCCTXCAFTests`.

Bridge-only: no OCCT kernel change and no `OCCT.xcframework` rebuild. `Sources/OCCTBridge/src/*.mm`
did change, so whoever tags this needs to rebuild and re-attach `OCCTBridge.xcframework.zip` and bump
`Package.swift`'s URL and checksum, per the usual release step.

CHANGELOG entry is deliberately unversioned, matching the #430/#439/#442 precedent.

Related: #439 (`outerShell`/`innerShells`, PR #441), #442 (`fixSolid`/`solidFromShellFixed`, PR #444).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 A comprehensive Swift wrapper for [OpenCASCADE Technology (OCCT)](https://www.opencascade.com/) 8.0.0p1, providing B-Rep solid modeling for macOS and iOS. **v1.0.0 — SemVer-stable as of 2026-05-07.**
 
-**4,258 wrapped operations** | macOS 12+ / iOS 15+ / visionOS 1+ / tvOS 15+ (arm64) | OCCT 8.0.0p1
+**4,260 wrapped operations** | macOS 12+ / iOS 15+ / visionOS 1+ / tvOS 15+ (arm64) | OCCT 8.0.0p1
 
 ## Quick Start
 
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -7588,6 +7588,23 @@ int32_t OCCTDocumentTriangulationNbTriangles(OCCTDocumentRef doc, int64_t labelI
 /// Get the deflection of a triangulation attribute.
 double OCCTDocumentTriangulationDeflection(OCCTDocumentRef doc, int64_t labelId);
 
+/// Read one node of a triangulation attribute, in the coordinate frame it was stored in.
+/// @param index 1-based node index, as Poly_Triangulation numbers them.
+/// @param outXYZ receives x, y, z. Untouched when the call returns false.
+/// @return false if the label has no triangulation attribute or the index is out of range.
+bool OCCTDocumentTriangulationNode(OCCTDocumentRef doc, int64_t labelId, int32_t index,
+                                   double* outXYZ);
+
+/// Read one node normal of a triangulation attribute, in the frame it was stored in.
+/// Node normals exist only when the meshed faces carried them; BRepMesh_IncrementalMesh does
+/// not produce any, so this is false for anything meshed from B-Rep and true for e.g. an
+/// imported glTF mesh.
+/// @param index 1-based node index.
+/// @param outXYZ receives the normal. Untouched when the call returns false.
+/// @return false if there is no attribute, no node normals, or the index is out of range.
+bool OCCTDocumentTriangulationNormal(OCCTDocumentRef doc, int64_t labelId, int32_t index,
+                                     double* outXYZ);
+
 // MARK: - TDataXtd Point/Axis/Plane Attributes (v0.56.0)
 
 /// Set a point attribute on a label.
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -1766,9 +1766,10 @@ OCCTShapeRef OCCTShapeCreateFaceFromWire(OCCTWireRef wire, bool planar);
 /// @return Face shape with holes, or NULL on failure
 OCCTShapeRef OCCTShapeCreateFaceWithHoles(OCCTWireRef outer, const OCCTWireRef* holes, int32_t holeCount);
 
-/// Create a solid from a closed shell
-/// @param shell Shell shape (must be closed)
-/// @return Solid shape, or NULL on failure
+/// Create a solid from every body-bounding shell of a shape, not just the first (#443).
+/// Cavity shells are skipped; see occtBodyBoundingShells in OCCTBridge_Internal.h.
+/// @param shell Shell shape (must be closed), or any shape holding shells
+/// @return One solid, a compound of solids for multi-body input, or NULL if there is no shell
 OCCTShapeRef OCCTShapeCreateSolidFromShell(OCCTShapeRef shell);
 
 /// Sew multiple faces/shapes into a shell or solid
@@ -2647,7 +2648,8 @@ OCCTShapeRef OCCTShapeConvertToBSpline(OCCTShapeRef shape);
 /// Sew a single shape (reconnect disconnected faces)
 OCCTShapeRef OCCTShapeSewSingle(OCCTShapeRef shape, double tolerance);
 
-/// Upgrade shape: sew + make solid + heal (pipeline)
+/// Upgrade shape: sew + make solid + heal (pipeline). One solid per body-bounding shell of
+/// the sewn result, so multi-body input stays multi-body (#443).
 OCCTShapeRef OCCTShapeUpgrade(OCCTShapeRef shape, double tolerance);
 
 
@@ -4869,7 +4871,9 @@ OCCTBooleanHistoryRef _Nullable OCCTShapeHealWithHistory(OCCTShapeRef _Nonnull s
                                                             OCCTShapeRef _Nullable * _Nullable outResult);
 
 /// Create a solid from a closed shell (BRepBuilderAPI_MakeSolid + ShapeFix_Solid
-/// orientation fix), with full history.
+/// orientation fix), with full history. Same body selection as
+/// OCCTShapeCreateSolidFromShell; one shared ReShape context, so the single history
+/// covers every body (#443).
 OCCTBooleanHistoryRef _Nullable OCCTShapeCreateSolidFromShellWithHistory(OCCTShapeRef _Nonnull shell,
                                                                            OCCTShapeRef _Nullable * _Nullable outResult);
 
@@ -7570,7 +7574,9 @@ bool OCCTDocumentHasGeometryAttr(OCCTDocumentRef doc, int64_t labelId);
 
 // MARK: - TDataXtd Triangulation Attribute (v0.56.0)
 
-/// Set a triangulation attribute on a label by meshing a shape.
+/// Set a triangulation attribute on a label by meshing a shape. Stores EVERY face's
+/// triangulation merged into one Poly_Triangulation, not just the first face's (#443).
+/// Returns false if the shape has no face, or if nothing in it meshed.
 bool OCCTDocumentSetTriangulationFromShape(OCCTDocumentRef doc, int64_t labelId, OCCTShapeRef shape, double deflection);
 
 /// Get the number of nodes in a triangulation attribute.
--- a/Sources/OCCTBridge/src/OCCTBridge_Document.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Document.mm
@@ -2144,9 +2144,25 @@ bool OCCTDocumentHasGeometryAttr(OCCTDocumentRef doc, int64_t labelId) {
 // Merge every meshed face of `shape` into one triangulation, in the coordinate system the
 // shape itself is expressed in: each face's per-face triangulation is stored in its own
 // TopLoc_Location, so the nodes are transformed on the way in, and a REVERSED face has its
-// winding (and node normals) flipped so the merged result is consistently outward, by the
-// same rules OCCTShapeCreateMesh applies. UV nodes are dropped: they index per-face parameter
-// spaces that no longer mean anything once the faces are pooled.
+// winding flipped so the merged result is consistently outward. UV nodes are dropped: they
+// index per-face parameter spaces that no longer mean anything once the faces are pooled.
+//
+// NOTE the frame. Before #443 this function fetched the location and DISCARDED it, storing
+// one face's nodes in that face's own local frame. For a located face (an assembly component,
+// anything from Shape.transformed()) the stored numbers therefore change, independently of
+// the "every face, not the first" fix. The shape's frame is the right one for an attribute
+// nothing else carries a location for, but it is a behaviour change; see the CHANGELOG.
+//
+// Node normals are handled MORE strictly than OCCTShapeCreateMesh, which pushes
+// triangulation->Normal(i) raw with neither the location transform nor the reversal
+// (OCCTBridge_Mesh.mm); only its winding swap matches this. That divergence means
+// Mesh.normals and this attribute disagree for a located or reversed face; settling which is
+// wrong is Shape.mesh()'s question, not this function's.
+//
+// The normal branch never fires on the ordinary path: BRepMesh_IncrementalMesh does not
+// populate node normals (measured, 0 of 6 box faces and 0 of 1 sphere face; there is no
+// SetNormal anywhere under occt-src TKMesh). It fires only for a face that arrived carrying
+// a normal-bearing triangulation from elsewhere, which glTF import does produce.
 //
 // Returns a null handle when the shape has no face, or when nothing in it meshed.
 static Handle(Poly_Triangulation) occtMergedTriangulation(const TopoDS_Shape& shape,
@@ -2241,6 +2257,47 @@ int32_t OCCTDocumentTriangulationNbTriangles(OCCTDocumentRef doc, int64_t labelI
         if (!label.FindAttribute(TDataXtd_Triangulation::GetID(), attr)) return 0;
         return attr->NbTriangles();
     } catch (...) { return 0; }
+}
+
+// Reading a stored node back was not possible before #443: the attribute exposed only its
+// counts and deflection, so nothing in the Swift API could see the coordinates it holds. That
+// is part of why the first-face bug, and the frame it stored in, went unnoticed.
+bool OCCTDocumentTriangulationNode(OCCTDocumentRef doc, int64_t labelId, int32_t index,
+                                   double* outXYZ) {
+    if (!doc || doc->doc.IsNull() || !outXYZ) return false;
+    try {
+        TDF_Label label = doc->getLabel(labelId);
+        if (label.IsNull()) return false;
+        Handle(TDataXtd_Triangulation) attr;
+        if (!label.FindAttribute(TDataXtd_Triangulation::GetID(), attr)) return false;
+        Handle(Poly_Triangulation) tri = attr->Get();
+        if (tri.IsNull() || index < 1 || index > tri->NbNodes()) return false;
+        const gp_Pnt p = tri->Node(index);
+        outXYZ[0] = p.X();
+        outXYZ[1] = p.Y();
+        outXYZ[2] = p.Z();
+        return true;
+    } catch (...) { return false; }
+}
+
+// HasNormals() first: Poly_Triangulation::Normal reads myNormals unconditionally, and that
+// array is empty for anything BRepMesh produced.
+bool OCCTDocumentTriangulationNormal(OCCTDocumentRef doc, int64_t labelId, int32_t index,
+                                     double* outXYZ) {
+    if (!doc || doc->doc.IsNull() || !outXYZ) return false;
+    try {
+        TDF_Label label = doc->getLabel(labelId);
+        if (label.IsNull()) return false;
+        Handle(TDataXtd_Triangulation) attr;
+        if (!label.FindAttribute(TDataXtd_Triangulation::GetID(), attr)) return false;
+        Handle(Poly_Triangulation) tri = attr->Get();
+        if (tri.IsNull() || !tri->HasNormals() || index < 1 || index > tri->NbNodes()) return false;
+        const gp_Dir n = tri->Normal(index);
+        outXYZ[0] = n.X();
+        outXYZ[1] = n.Y();
+        outXYZ[2] = n.Z();
+        return true;
+    } catch (...) { return false; }
 }
 
 double OCCTDocumentTriangulationDeflection(OCCTDocumentRef doc, int64_t labelId) {
--- a/Sources/OCCTBridge/src/OCCTBridge_Document.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Document.mm
@@ -18,6 +18,8 @@
 
 // === Area-specific OCCT headers ===
 
+#include <algorithm>
+
 #include <STEPCAFControl_Reader.hxx>
 #include <STEPCAFControl_Writer.hxx>
 #include <STEPControl_StepModelType.hxx>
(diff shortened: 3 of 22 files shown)


gsdali (commented):
## Review

### Overview

Closes out #442's audit note. A 23-site sweep of the first-of-N `TopExp_Explorer` idiom: three sites that lose most of their input are fixed (`Shape.solid(from:)` plus its history variant, `Shape.upgraded()`, `AssemblyNode.setTriangulationFromShape`), seven singular-by-contract sites get a `#443 audit:` bridge comment plus a `- Note:` on the Swift wrapper, and #442's `occtBodyBoundingShells` is promoted to `OCCTBridge_Internal.h` so the two solid-from-shell families answer the same question the same way. A latent case in that helper is also changed: free shells now go through enclosure parity as one group instead of being emitted unconditionally.

The audit-and-document-the-rest approach is right, and the `#443 audit:` marker at each unchanged site is a good pattern. It makes the next sweep cheap and records the reasoning where someone will actually hit it.

**Verified against the pinned sources** (the load-bearing claim for the history variant): `BRepTools_ReShape::History()` does rebuild a fresh `BRepTools_History` from the whole `myShapeToReplacement` map on each call, and `ShapeFix_Solid.cxx` never clears its context, only `Apply`/`Replace`. So one shared `ShapeBuild_ReShape` across the per-body fixers genuinely covers every body. That claim holds.

### Blocking

**The PR does not merge.** `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. The branch is two commits behind `main`: v1.16.0 shipped (`a3e7f51`, `4bdc30a`) and stamped the `### Unreleased: ...` headings for #442/#439/#430 into `### v1.16.0 (July 2026): ...`, which is exactly the CHANGELOG context this diff carries. Needs a rebase onto `4bdc30a`; the new entry then belongs above the now-versioned v1.16.0 block, still unversioned per the established convention.

### Correctness and performance

**1. The free-shell parity group is O(F²) on the input this PR documents as the primary one.**

Free shells were previously appended unconditionally, which is O(F). They now form a single parity group, which costs F `BRepClass3d_SolidClassifier` constructions (each `Load`s every face of its reference shell and builds a UB-tree) plus F(F-1) ray-cast `Perform` calls.

Inside a solid, F was 1-3, which is what the inherited comment still assumes. Free shells are a different scale: sewing a raw imported mesh routinely yields hundreds of disjoint shells, and this PR's own text calls `upgraded()` "the call most likely to be pointed at a raw imported mesh, where multi-body input is the norm." At F=200 that is 200 classifier loads and roughly 40,000 ray casts where the old code did nothing. No new test exceeds three bodies, so nothing here would surface it.

Concrete fix, a conservative bounding-box pre-filter that prunes disjoint pairs before any ray cast:

```cpp
std::vector<Bnd_Box> boxes(shells.size());
for (size_t k = 0; k < shells.size(); k++) BRepBndLib::Add(shells[k], boxes[k]);
// ... inside the i loop, before building the reference:
//     skip i entirely if no box is inside boxes[i]
// ... inside the j loop:
if (boxes[i].IsOut(boxes[j])) continue;   // cannot be enclosed
```

Enclosure implies box containment, so this only removes pairs that could never have been enclosed; parity is unchanged. Worth noting this is not a return to the rule #442 round 3 rejected: `Bnd_Box` failed as the *decision* rule, but as a *conservative pre-filter* it is sound.

**2. `setTriangulationFromShape` changes coordinate frames, and that is not flagged anywhere.**

The old code did `BRep_Tool::Triangulation(face, loc)` and **discarded `loc`**, storing the first face's nodes in that face's own local frame. The new one transforms every node into the shape's frame. For a single-face shape with a non-identity `TopLoc_Location` (an assembly component, anything from `Shape.transformed()`) the stored attribute now holds different numbers, independent of the "more faces" fix.

That is a second behaviour change and it is absent from the PR body, the CHANGELOG behaviour-change block, and the doc comment, which states the new frame descriptively without saying it moved. Anyone with a persisted OCAF document and a reader comparing before and after gets silently different geometry. It should be in the behaviour-change note alongside the node-count change.

**3. "The same rules `OCCTShapeCreateMesh` applies" is not accurate for normals.**

Reading `OCCTBridge_Mesh.mm:88-176`, `OCCTShapeCreateMesh` pushes `triangulation->Normal(i)` **raw**: no location transform, no reversal for a REVERSED face. Only the winding swap matches. `occtMergedTriangulation` does both (`normal.Reverse()` then `normal.Transformed(transformation)`), so it is stricter than the function it claims parity with.

Either the comment overstates, or `Shape.mesh()`'s node normals are wrong for located/reversed faces. Its per-triangle normals are computed from transformed nodes with swapped winding, so those are fine; it is only `Mesh.normals` in question. Worth settling, because as it stands a consumer reading `label.triangulation` and `shape.mesh()` for the same shape gets disagreeing normals. If `Shape.mesh()` is the defective one that is its own issue, not this PR's, but the claim of parity should not ship either way.

**4. The normal branch is unreachable on the ordinary path and untested on the other.**

`BRepMesh_IncrementalMesh` does not populate node normals; there is no `SetNormal` anywhere under `occt-src/src/ModelingAlgorithms/TKMesh/`. So `tri->HasNormals()` is false for every face this function meshes, `hasNormals` is always false, and the flip/transform code never runs. It fires only when a face arrives carrying a normal-bearing triangulation from elsewhere (glTF/OBJ import, `Mesh.toShape`), and no test covers that. Combined with (3), the normal handling is both described inaccurately and unverified.

**5. Test coverage gap: no located, transformed or mirrored shape anywhere in the triangulation tests.** Every fixture is an untransformed box, sphere or edge, so the `Transformed(transformation)` path that finding (2) is about is never exercised with a non-identity location. Given #375's history in this repo with mirrored-shape normalization, a mirrored fixture is worth one test on its own.

### Minor

- The shared `ShapeBuild_ReShape` in `OCCTShapeCreateSolidFromShellWithHistory` means body B's `Perform` calls `Context()->Apply()` against a map already holding body A's replacements. Harmless for disjoint sewn bodies (`Apply` returns unmapped shapes unchanged), but if two bodies share sub-shapes, A's replacement reaches B. Worth a clause in the comment that already explains the sharing.
- The parity comment inherited into `occtSelectBodyShells` still carries #442's framing ("one classifier per reference shell"), which read fine when a group was one solid's 1-3 shells. Now that the same function serves the unbounded free-shell group, the cost note should say so.

### Good

- Behaviour changes are stated up front, per call, with measured numbers, and the free-shell regrouping is direct about changing a *just-released* result (`solidFromShellFixed()` on a sewn hollow part). The sewn/un-sewn agreement it buys is the right trade.
- The nested-body-in-a-cavity case (enclosed twice, still a body) is both documented and tested. That is the case the parity rule exists for.
- Tests follow the loud-guard convention settled in #444: `expectVolume` records on nil, every `guard`-`else` fails, labelled expectations.
- The deflection detail being caught by the *existing* test rather than a new one, and said so in the PR body, is the kind of thing worth writing down.

---

**Verdict:** the shell-selection consolidation is correct and the three fixes are real. Rebase is required before anything else. Of the substantive items, the O(F²) free-shell pass is the one I would fix in this PR: it lands a quadratic cost on `upgraded()`'s documented primary input with no test that would notice. Items 2-5 are all in `setTriangulationFromShape` — the frame change needs disclosing, and the normal handling needs either a correct description or a test.

gsdali (commented):
## Overview

This continues the "first-of-N `TopExp_Explorer`" audit started in #439/#442. It:

1. **Fixes 3 silent-data-loss sites**: `Shape.solid(from:)` / `solidWithFullHistory(from:)`, `Shape.upgraded()`, and `AssemblyNode.setTriangulationFromShape` — each previously took only the first shell/face found and dropped the rest.
2. **Shares** `occtBodyBoundingShells`/`occtSolidBodiesToShape` (from #442) between `OCCTBridge_Healing.mm` and `OCCTBridge_Modeling.mm` via `OCCTBridge_Internal.h`.
3. **Extends** the enclosure-parity algorithm to a second free-shell group (previously free shells were added unconditionally, which double-counted a sewn hollow body's cavity), with a bounding-box pre-filter to avoid O(N²) ray casts on large free-shell sets.
4. **Documents** 7 more first-of-N sites that stay singular by contract.
5. Adds 2 new triangulation-attribute read accessors, 27 new tests, and updates README/API_REFERENCE/CHANGELOG/cookbook/reference docs.

Well-scoped and well-evidenced — every claim is backed by a measured before/after table. I spot-checked the "documented, not changed" claims against the actual bridge bodies (`OCCTFace2DFillet`, `OCCTMedialAxisCompute`, `OCCTLocOpeSplitByWireOnFace`, `OCCTLocOpeSplitByWires`, `OCCTSolidFromShells`) and they're all accurate.

## Findings

### 1. `OCCTShapeUpgrade`'s per-shell loop reintroduces the exact silent-drop pattern this PR just fixed elsewhere (medium)

In `OCCTBridge_Healing.mm`, the new `OCCTShapeUpgrade` body:

```cpp
std::vector<TopoDS_Shape> made;
for (const TopoDS_Shell& shell : occtBodyBoundingShells(sewedShape)) {
    BRepBuilderAPI_MakeSolid makeSolid(shell);
    if (makeSolid.IsDone()) made.push_back(makeSolid.Solid());
}
```

drops the shell entirely from `made` if `IsDone()` is false — no fallback push. Compare this to the *sibling* loops touched in this same diff:

- `OCCTShapeCreateSolidFromShell` (Modeling.mm): `if (solid.IsNull()) { made.push_back(topoShell); continue; }`
- `OCCTShapeCreateSolidFromShellWithHistory` (Modeling.mm): identical push-not-drop fallback
- `OCCTShapeSolidFromShell` (Healing.mm, from #442, unchanged here): `made.push_back(solid.IsNull() ? TopoDS_Shape(shell) : TopoDS_Shape(solid));`

Three of the four per-body solid-construction call sites explicitly push the un-fixed shell as a fallback rather than dropping the body. `OCCTShapeUpgrade` — touched in this very diff — is the one holdout. Review round 3 argues (convincingly, via `occt-src`) that `BRepBuilderAPI_MakeSolid`'s single-shell constructor always succeeds, so this is very likely dead code today — but that argument applies equally to `OCCTShapeUpgrade`'s identical constructor call, which makes the omission an inconsistency rather than a deliberate exception. Given the whole point of #443 is closing exactly this class of "quietly drops part of the input" bug, please close it for symmetry/defense-in-depth:
```cpp
if (makeSolid.IsDone()) made.push_back(makeSolid.Solid());
else made.push_back(shell);
```

### 2. Existing `docs/reference/` pages weren't updated for 4 of the "documented, not changed" sites (low-medium)

CLAUDE.md's release process requires updating *"any `docs/reference/<Type>.md` page covering a changed type"* in the same PR. This PR correctly updated `Shape-Features.md`, `Shape-Healing.md`, `Document.md`, and `Document-OCAF-Attributes.md` for the 3 fixed sites — but the new `- Note:` caveats added to the Swift doc comments for the "documented, not changed" sites weren't mirrored into the existing reference pages that already cover those exact methods:

- `docs/reference/Shape-Measurement.md` — `fillet2D(vertexIndices:radii:)`, `chamfer2D(edgePairs:distances:)`, `solidFromShells(_:)` all have existing entries, none updated
- `docs/reference/Shape-Builders-1.md` — `splitByWireOnFace(_:faceIndex:)`, not updated
- `docs/reference/Shape-Builders-2.md` — `locOpeSplit(wiresOnFaces:)`, not updated

(`MedialAxis` and `halfSpace` have no pre-existing reference page, so those are a pre-existing gap, not a regression from this PR.) Please propagate the new caveats into these three pages before merge — context7 harvests them, and this project already treats doc/source drift as a real defect (see the #289-driven desync rule in API_REFERENCE.md).

### 3. Imprecise wording: "free shells each become a body" (low)

`Shape.upgraded()`'s new doc comment (Shape.swift) and its mirror in `docs/reference/Shape-Healing.md` both say:

> Body selection is the same rule as `Shape.solid(from:)`: cavity shells are skipped, free shells each become a body.

This isn't quite what the code does after this PR's own fix — free shells now go through the *same enclosure-parity pass* as a solid's own shells, so a free shell that's an even-enclosed cavity (the sewn-hollow-box case this PR specifically fixes, and covers with `sewnHollowIsOneBody`/`upgradedHollow`) is *skipped*, not turned into a body. The `- Note:` two lines below correctly describes this, but the topic sentence on its own reads as if it contradicts the Note that follows it. `Shape.solid(from:)`'s own doc comment states the same concept more accurately ("every shell that an even number of the others in its group enclose, where a group is ... all the shells belonging to no solid") — worth reusing that phrasing here instead.

## Code Quality

- The `occtSelectBodyShells` extraction is a faithful refactor of the pre-existing per-solid logic plus an additive, correctness-preserving bounding-box pre-filter — enclosure implies box containment, so pruning on `Bnd_Box::IsOut` cannot flip a verdict. Good instinct calling out that this is *not* the `Bnd_Box`-as-decision-rule #442 rejected.
- `occtMergedTriangulation`'s two-pass (count, then build) approach, offset bookkeeping (`base`/`nodeAt`/`triangleAt`), and per-face location/reversal handling are correct on inspection — node index math (`base + n1`) and pre-increment array-fill are consistent with `Poly_Triangulation`'s 1-based indexing.
- Comments are unusually rigorous throughout — each non-obvious decision is backed by a measured example, matching the project's stated preference for WHY-comments over WHAT-comments.
- Test suite (`Issue443FirstOfNTests.swift`, new `OCCTXCAFTests` suite) is thorough: covers the fix, the sibling-agreement regression, the free-shell parity change, the pre-filter (both the pruned and non-prunable "touching bodies" case), the frame-change behavior change, and the glTF-normals branch specifically because it's otherwise unreachable. Good use of `Issue.record` + guard pattern (no force-unwraps in `#expect`).

## Risk / Behavior-Change Assessment

The PR is upfront about the breaking change: `Shape.solid(from:)`, `solidWithFullHistory(from:)`, and `upgraded()` now return a compound instead of one arbitrary body's solid for multi-body input, and `solidFromShellFixed()` now returns one body instead of two for a sewn hollow part. Both are flagged in the PR body and CHANGELOG's behaviour-change block. No concerns there.

No OCCT kernel changes, no thread-safety-sensitive globals touched, no security-relevant surface (pure geometry/topology on already-in-memory shapes).

## Requesting changes

Findings #1 and #2 are small, mechanical fixes — please address before merge. #3 is a nit, fold in if convenient.

gsdali (commented):
## Re-review note

This is an updated revision. All three findings from my previous review comment on this PR have been addressed, confirmed by re-reading the current diff:

1. **`OCCTShapeUpgrade`'s silent-drop inconsistency** — fixed. Its per-shell loop now matches the other three per-body solid-construction call sites:
   ```cpp
   made.push_back(makeSolid.IsDone() ? TopoDS_Shape(makeSolid.Solid())
                                      : TopoDS_Shape(shell));
   ```
   with a comment explicitly citing the review finding. A new, well-considered test (`upgradedKeepsOpenBody`) backs it — and its doc comment shows real independent verification rather than a copy-paste of the sibling test: it correctly notes that `upgraded()`'s later `ShapeFix_Shape` pass reclassifies the kept-but-unclosable body so `solids.count` isn't the right assertion here (unlike `solidFromKeepsOpenBody`), and pins the guarantee on face count instead. Good catch on their part.
2. **Stale `docs/reference/` pages** — fixed. `Shape-Builders-1.md`, `Shape-Builders-2.md`, and `Shape-Measurement.md` all now carry the `(#443 audit)` caveat notes matching the Swift doc comments.
3. **Imprecise "free shells each become a body" wording** — fixed, in both `Shape.swift` and `docs/reference/Shape-Healing.md`: now correctly states free shells go through the same enclosure-parity pass, so an even-enclosed free shell is skipped rather than always becoming a body.

## New (minor) finding: this round's own fixes aren't reflected in CHANGELOG.md or the PR description

The CHANGELOG's `#443` entry still ends at "Review round 3" (the `Shape.solid(from:)`/`solidWithFullHistory(from:)` `MakeSolid`-failure fix) — the diff for `docs/CHANGELOG.md` is byte-identical to the previous revision's. This round fixed a different, sibling instance of the same defect class (`OCCTShapeUpgrade`) plus the doc-wording/reference-page gaps, but there's no "Review round 4" paragraph documenting it, unlike rounds 2 and 3 which each got one. Given this project's own consistent practice — and CLAUDE.md's "no stale docs" / same-PR-doc-parity rule that I'd flag if it were a source doc comment — it's worth a short paragraph for consistency, even though it's a small, defense-in-depth-only fix.

Relatedly, the top-level PR description's verification section is now stale: it says "27 new tests: 18 in `Issue443FirstOfNTests.swift`, 9 plus one extended... in `OCCTXCAFTests`" and "swift test green at 4,495 tests" — the file now has 22 `@Test`s (confirmed by counting the current diff), so the counts predate this round's additions. Low priority (it's the GitHub PR description, not a repo-tracked file), but worth a refresh before merge so the final PR record matches what actually shipped.

## Everything else from the original review still stands

- The `occtSelectBodyShells` extraction/bounding-box pre-filter is a faithful, correctness-preserving refactor (enclosure implies box containment, so pruning can't flip a verdict).
- `occtMergedTriangulation`'s offset bookkeeping and per-face location/reversal handling for the triangulation-attribute fix are correct on inspection.
- Test coverage is thorough and specific (numeric volume/face-count assertions rather than just "didn't crash"), and the comments throughout continue to justify non-obvious decisions with measurements rather than assertions.
- No OCCT kernel change, no thread-safety-sensitive surface, consistent with the "bridge-only" framing.

Nothing here blocks merge — the CHANGELOG/PR-description gap is cosmetic. Nice iteration; the fixes read as genuinely re-derived (the `upgradedKeepsOpenBody` face-count-vs-solid-count distinction in particular) rather than mechanically copied from the sibling fix.

Issue #274, opened 2026-07-04
Shape.face(outer:holes:) unconditionally reverses hole wires — wrong when a hole is already wound opposite the outer


gsdali:
## Summary

`Shape.face(outer:holes:)` — `OCCTShapeCreateFaceWithHoles` in `Sources/OCCTBridge/src/OCCTBridge_Modeling.mm` (~line 2628) — unconditionally reverses **every** hole wire it is given:

```cpp
OCCTShapeRef OCCTShapeCreateFaceWithHoles(OCCTWireRef outer, const OCCTWireRef* holes, int32_t holeCount) {
    ...
    BRepBuilderAPI_MakeFace makeFace(outer->wire, true);  // planar
    ...
    for (int32_t i = 0; i < holeCount; i++) {
        if (holes[i]) {
            // Inner wires must be reversed to represent holes
            TopoDS_Wire reversed = TopoDS::Wire(holes[i]->wire.Reversed());
            makeFace.Add(reversed);
        }
    }
    ...
}
```

This is only correct when the caller's hole wire already shares the **same** raw winding sense as the outer wire (so the forced `.Reversed()` lands it opposite, which is what OCCT's `BRepBuilderAPI_MakeFace` requires for a hole). The doc comment / example on the Swift side (`Shape.swift` ~line 1282) doesn't state this precondition at all:

```swift
/// Create a face with holes from outer and inner wires
/// - Parameters:
///   - outer: The outer boundary wire (closed)
///   - holes: Array of inner boundary wires defining holes
public static func face(outer: Wire, holes: [Wire]) -> Shape? { ... }
```

If a caller's hole wire is natively wound **opposite** the outer (which is, in fact, the *correct* configuration for a hole — the caller did nothing wrong), this function silently reverses it into the *wrong* sense (same as the outer), and the face build fails or the resulting solid is topologically invalid.

## How we hit this (OCCTDesignLoop issue #51)

[...]

PR #448, opened 2026-07-27
fix(#397): faceAddHole rejected every circular hole wire, and never oriented the ones it kept

gsdali:
Fixes #397.

## What was wrong

`Shape.faceAddHole(face:wire:)` returned `nil` for **every** hole wire built from circular geometry —
`Wire.circle(origin:normal:radius:)` and a hand-joined two-arc circle alike, at any radius, in either
winding — while a polygonal hole on the same face worked.

The cause was not in OCCT. It was this wrapper's own degenerate-wire guard, added for #234 (which
declines a zero-area hole because the invalid face it produces goes on to SIGSEGV `ShapeFix`
downstream). The guard counted the wire's **vertices**: a circle has one (`Wire.circle`) or two (two
joined arcs), so every curved hole tripped the "fewer than 3 distinct vertices" rejection meant for
out-and-back line segments. These wires never reached `BRepBuilderAPI_MakeFace::Add` at all — the
issue's suspicion that `Add` was rejecting them was a red herring.

## And a second half, exposed by fixing the first

Once the wires got through, they came out wrong. `MakeFace::Add` does no reorienting of its own, and
this wrapper never oriented the wire either, so a hole wound the same way as the face's outer boundary
was added as a second **outer** loop. Measured on a 20×20 face with a 2×2 hole:

```
square hole CCW: area=404.0  extruded solid valid=false   <- silently broken, pre-existing
square hole CW:  area=396.0  extruded solid valid=true
```

Only callers who happened to hand in an opposite-wound wire ever got a hole. This is pre-existing and
predates #397, but shipping "circles are accepted now" without it would just move the silent failure
one step later, so both halves are fixed here.

After:

```
square hole CCW: area=396.0  valid=true      circle CCW: area=371.73  valid=true
square hole CW:  area=396.0  valid=true      circle CW:  area=371.73  valid=true
```

## The fix

- The degenerate guard samples points **along the wire's curves** instead of at its vertices.
- Sampling alone would weaken #234's protection — an arc traversed out and back spreads its samples
  over a curve and clears the collinearity test that catches a straight out-and-back — so the loop's
  own vector area is checked too: mean width (area ÷ longest chord) below `Precision::Confusion()` is
  still declined.
- Winding is decided by comparing the hole's signed area against the face's outer boundary in the
  face's plane, the same rule `OCCTShapeCreateFaceWithHoles` has used since #274, with a
  validity-checked retry of the other orientation for non-planar hosts where no plane can be fitted.
- Deduplication: the wire sampler both tests need is now one helper (`occtSampleWirePoints`) rather
  than two copies of the same `BRepTools_WireExplorer` traversal. `occtSignedWireAreaInPlane` is
  rewritten on top of it with identical sample count, order and semantics.

Bridge-only: no OCCT kernel change, no xcframework rebuild, no new operations (count unchanged at
4,258).

## Tests

New `Issue397CircularHoleTests` (`OCCTModelingTests`), each failing before this change:

- `Wire.circle` hole accepted, and the hole is really cut (area = 400 − πr²)
- two-arc circle hole accepted in both windings
- the holed face extrudes into a valid solid of the right volume, with its circular edges intact
- a same-winding polygon hole cuts instead of adding area (the second half above)
- a zero-area curved wire (an arc traversed out and back) is **still** declined — the #234 guard rail

`Issue234DegenerateHoleTests` passes unchanged. Full `swift test`: 4473 tests in 1291 suites, all
passing.

### DIFF
--- a/Sources/OCCTBridge/src/OCCTBridge_Modeling.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Modeling.mm
@@ -98,6 +98,7 @@
 #include <BRepAlgo_AsDes.hxx>
 #include <BRepCheck_Analyzer.hxx>
 #include <BRepBuilderAPI_FindPlane.hxx>
+#include <BRepTools.hxx>
 #include <BRepTools_WireExplorer.hxx>
 #include <BRepAdaptor_Curve.hxx>
 #include <ShapeUpgrade_UnifySameDomain.hxx>
@@ -2717,10 +2718,33 @@ OCCTShapeRef OCCTShapeCreateFaceFromWire(OCCTWireRef wire, bool planar) {
     }
 }
 
+// 3D points sampled along a wire in traversal order, `samplesPerEdge + 1` per edge (both
+// endpoints included, so consecutive edges repeat their shared point). Arc-aware: sampling the
+// parametric range is what lets a curved edge contribute its true bulge rather than just its two
+// vertices — the distinction #397 turned on. Degenerate edges, and edges carrying no 3D curve, are
+// skipped rather than throwing.
+static std::vector<gp_Pnt> occtSampleWirePoints(const TopoDS_Wire& wire, int samplesPerEdge) {
+    std::vector<gp_Pnt> pts;
+    for (BRepTools_WireExplorer explorer(wire); explorer.More(); explorer.Next()) {
+        const TopoDS_Edge& edge = explorer.Current();
+        if (BRep_Tool::Degenerated(edge)) continue;
+        try {
+            BRepAdaptor_Curve curve(edge);
+            const double f = curve.FirstParameter();
+            const double l = curve.LastParameter();
+            // Walk the curve in the edge's traversal direction within the wire.
+            const bool reversed = (edge.Orientation() == TopAbs_REVERSED);
+            for (int s = 0; s <= samplesPerEdge; s++) {
+                const double t = (double)s / (double)samplesPerEdge;
+                pts.push_back(curve.Value(reversed ? (l + (f - l) * t) : (f + (l - f) * t)));
+            }
+        } catch (...) { continue; }   // no 3D curve on this edge — nothing to sample
+    }
+    return pts;
+}
+
 // Signed area of a (nominally planar) wire measured in the plane spanned by pln's
-// X/Y directions. Arc-aware: each edge is sampled along its parametric range in
-// wire-traversal order and the shoelace sum is accumulated over the sampled polyline,
-// so curved edges contribute their true bulge to the sign. The magnitude is only an
+// X/Y directions, as a shoelace sum over the sampled polyline. The magnitude is only an
 // approximation, but the SIGN — all we use it for — is robust for simple loops.
 // A positive result means the wire winds counter-clockwise about pln's normal.
 static double occtSignedWireAreaInPlane(const TopoDS_Wire& wire, const gp_Pln& pln) {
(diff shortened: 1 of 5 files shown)


gsdali (commented):
## Review: `fix(#397): faceAddHole rejected every circular hole wire, and never oriented the ones it kept`

**Verification performed:** built `OCCTModelingTests` on the PR's head commit (`1651130`, via a temporary worktree) and ran `Issue397CircularHoleTests`, `Issue234DegenerateHoleTests`, and `Issue274HoleWireOrientationTests`. All 11 tests pass, confirming the PR's own test-run claims and confirming no regression to the two prior issues this touches.

### Overview

`Shape.faceAddHole` rejected every hole wire built from circular geometry because its #234 degenerate-wire guard counted **vertices** (a circle has 1–2), not enclosed area. Fixing that exposed a second, pre-existing bug: `MakeFace::Add` never reorients hole wires, so a same-winding hole was silently added as a second *outer* loop instead of subtracting area. This PR fixes both:

1. `occtSampleWirePoints` — new shared helper sampling points along a wire's curves (arc-aware), replacing the vertex-count degeneracy check and de-duplicating what was two copies of the same `BRepTools_WireExplorer` traversal.
2. A vector-area (mean-width) check added on top of the point-count/collinearity checks, to keep rejecting a curved out-and-back wire that would otherwise clear the old checks by spreading samples over a curve.
3. `occtFacePlane` — new helper to get a face's plane (direct `Geom_Plane` cast, or `BRepBuilderAPI_FindPlane` fallback), used to compare hole/outer winding and conditionally reverse the hole wire, mirroring the rule `OCCTShapeCreateFaceWithHoles` established for #274.
4. A validity-checked retry: if the winding decision produces an invalid face, the opposite orientation is tried and kept if it validates.

### Code quality / conventions

- Matches project conventions well: try/catch around throwing OCCT calls, degenerate-edge skip in the sampler, no force-unwraps in the new `#expect`s, Swift doc comment with a runnable fenced snippet on the changed public API (context7-harvestable), CHANGELOG entry correctly leaves version/date unset per this project's standing convention.
- Good reuse: the winding-comparison logic explicitly follows the established `OCCTShapeCreateFaceWithHoles` (#274) pattern rather than inventing a new one, and the new `occtSampleWirePoints` helper collapses duplicate traversal code that predated this PR.
- Comments are appropriately dense for this codebase's existing style in this file (long-running rationale blocks are the local norm, e.g. the #234 comment this replaces).

### Potential issue (minor, pre-existing shape)

In the final fallback:
```cpp
TopoDS_Face holed = build(reverse);
if (holed.IsNull() || !BRepCheck_Analyzer(holed).IsValid()) {
    TopoDS_Face alt = build(!reverse);
    if (!alt.IsNull() && BRepCheck_Analyzer(alt).IsValid()) holed = alt;
    else if (holed.IsNull()) holed = alt;
}
if (holed.IsNull()) return nullptr;
```
If **neither** orientation produces a `BRepCheck_Analyzer`-valid face (e.g., a hole that lies partly outside the outer boundary, or self-intersects it — not a winding problem at all), the function falls through and returns the first, *invalid* `holed` face rather than `nil`. This isn't a regression — the pre-PR code never validated the result either, so an out-of-bounds hole was always silently accepted — but the new code introduces a validity check specifically to arbitrate between orientations, and it would read more consistently if "both invalid" also returned `nil`, matching the function's now-implied contract ("nil signals a bad hole") and the naming convention that fallible ops return optionals. Low severity given it doesn't make anything worse than before.

### Test coverage

Strong. New `Issue397CircularHoleTests` covers: single-edge `Wire.circle`, hand-joined two-arc circle in both windings, extrusion into a valid solid with genuine circular edges, the same-winding-polygon "adds area instead of cutting" regression, and a zero-area curved (out-and-back arc) wire that must still be rejected. Confirmed all pass, and confirmed no regression on the adjacent #234/#274 suites.

### Risk assessment

Bridge-only change (no OCCT kernel patch, no xcframework rebuild) — low blast radius, confined to `OCCTMakeFaceAddHole` plus a shared-helper extraction that other call sites (`OCCTShapeCreateFaceWithHoles`) don't consume, so no cross-function coupling risk. No concurrency, no new public API surface, op count correctly unchanged (verified 4,258 in README/API_REFERENCE match the PR's claim).

**Recommendation:** approve. The one note above (both-orientations-invalid silently returning a non-nil invalid face) is worth a follow-up comment but isn't blocking, since it's not a new failure mode this PR introduces.

PR #449, opened 2026-07-27
fix(#446): unify consumed the shape it was given, so a declined merge still damaged the caller's solid

gsdali:
Fixes #446.

## What was wrong

`ShapeUpgrade_UnifySameDomain` rewrites sub-shapes of the shape it is handed, and those rewrites reach
the `TShape`s the caller's `Shape` still shares. So the idiom every consumer writes — take the merge if
it is valid, otherwise keep what you had — silently damaged what you had. OCCT documents the class as
producing a new shape and says nothing about the input being consumed.

## Root cause, traced in the kernel

`TransformPCurves` (`ShapeUpgrade_UnifySameDomain.cxx:1228`, plus two sibling sites) writes temporary
pcurves onto the **input's** edges, against a scratch reference face the algorithm builds for itself,
and only ever removes them again if that reference face is later replaced. `SetSafeInputMode` does not
cover that path — it is unguarded — and safe mode is OCCT's default anyway, so the reporter was
already running it.

**Minimal in-repo reproducer**: two stacked coaxial cylinders. Their cylindrical faces are same-domain
but differently parameterised, which is exactly what drives that path. The input's serialized BREP
grows across a call whose result the caller never uses:

```
Shape.unified()           input BREP 1676 -> 1778 bytes
UnifySameDomainBuilder    input BREP 1676 -> 1778 bytes
Shape.simplified()        input BREP 1676 -> 1778 bytes
```

After this change all three are byte-identical, with the merge result unchanged (4 faces → 3).

## The fix

Every unify entry point works on a private copy (`BRepBuilderAPI_Copy`), so the caller's shape is
untouched whatever the algorithm does to its own input.

No API change and no new parameter. The issue floated `copyInput: Bool = true` as an escape hatch;
the copy is unconditional instead, because "the input survives" is what every existing call site
already assumed, and the copy costs a fraction of the merge it precedes. It can be made optional later
if a caller ever measures it as a problem.

## Deduplication (the cleanup half)

Three bridge call sites constructed `ShapeUpgrade_UnifySameDomain` independently — `OCCTShapeUnifySameDomain`
and `OCCTShapeSimplify` in `OCCTBridge_Healing.mm`, the builder in `OCCTBridge_Modeling.mm` — each with
its own copy of the construct/`Build()`/null-check sequence. That triplication is precisely why this fix
would otherwise have had to be written three times. They now share
`occtUnifySameDomain` / `occtUnifySameDomainInput` / `occtUnifySameDomainMapped` (`OCCTBridge_Internal.h`).

The two **public Swift** entry points were checked for redundancy and both are justified, so both stay:

| | role |
|---|---|
| `Shape.unified(unifyEdges:unifyFaces:concatBSplines:)` | one-shot |
| `UnifySameDomainBuilder` | adds angular/linear tolerance, `keepShape`, internal-edge control |
| `Shape.simplified(tolerance:)` | `unified()` + heal, now over the shared copy-protected path |

Two papercuts found while surveying them:

- their `concatBSplines` defaults disagree (`true` for `Shape.unified()`, `false` for the builder).
  Left as-is rather than silently changed under existing callers, but now cross-documented on both.
- `setSafeInputMode(_:)`'s doc comment claimed it "copies input shape to preserve original". It copies
  nothing, and does not preserve the original either — corrected.

`keepShape(_:)` names a sub-shape of the **caller's** shape, so working on a copy means mapping it to
its counterpart there; without that mapping every `keepShape` would have quietly kept nothing. A test
covers it.

Bridge-only: no OCCT kernel change, no xcframework rebuild, no new operations (count unchanged at 4,258).

## Tests

New `Issue446UnifyInputMutationTests` (`OCCTShapeHealingTests`):

- input BREP byte-identical across `Shape.unified()`, `UnifySameDomainBuilder`, `Shape.simplified()` —
  these three **fail on the pre-fix bridge** (1676 vs 1778 bytes), verified by reverting `Sources/OCCTBridge`
  under the new tests
- a declined merge leaves the caller's validity, self-intersection and volume unchanged (the reported
  symptom's shape; on this small fixture it holds either way, the byte check is the discriminating one)
- `keepShape` still blocks a merge through the copy

Full `swift test`: 4473 tests in 1291 suites, all passing.

### DIFF
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -18857,6 +18857,17 @@ void OCCTPipeShellSetBuildHistory(OCCTPipeShellRef _Nonnull ps, bool enabled);
 typedef void* OCCTUnifySameDomainRef;
 
 /// Create a UnifySameDomain builder.
+///
+/// The builder unifies a private COPY of `shape` — the algorithm rewrites its input, and those
+/// rewrites used to reach the caller's shape (#446). `OCCTUnifySameDomainKeepShape` takes the
+/// CALLER's sub-shapes and maps them onto the copy.
+///
+/// NOTE: despite the `_Nonnull` annotation this can return null — on a construction failure (which
+/// predates #446) and now also if the input cannot be copied. Every accessor below tolerates a null
+/// builder, and `OCCTUnifySameDomainRelease` accepts one, so a null degrades to "does nothing,
+/// answers null" rather than crashing. The annotation is kept for source compatibility: correcting
+/// it to `_Nullable` would make the Swift `UnifySameDomainBuilder.init` failable, which is a
+/// breaking API change for a path no caller can hit with a valid shape.
 OCCTUnifySameDomainRef _Nonnull OCCTUnifySameDomainCreate(OCCTShapeRef _Nonnull shape,
                                                            bool unifyEdges, bool unifyFaces, bool concatBSplines);
 
--- a/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Healing.mm
@@ -353,16 +353,49 @@ OCCTShapeRef OCCTShapeFixDetailed(OCCTShapeRef shape, double tolerance,
     }
 }
 
+// #446: the three shared helpers declared in OCCTBridge_Internal.h — see the block comment there
+// for why every unify entry point works on a copy.
+TopoDS_Shape occtUnifySameDomainInput(const TopoDS_Shape& shape, BRepBuilderAPI_Copy& copier) {
+    try {
+        copier.Perform(shape);
+        return copier.Shape();
+    } catch (...) {
+        return TopoDS_Shape();
+    }
+}
+
+TopoDS_Shape occtUnifySameDomainMapped(const TopoDS_Shape& sub, BRepBuilderAPI_Copy& copier) {
+    try {
+        // ModifiedShape raises Standard_NoSuchObject for a shape that was not part of the copy.
+        TopoDS_Shape mapped = copier.ModifiedShape(sub);
+        return mapped.IsNull() ? sub : mapped;
+    } catch (...) {
+        return sub;
+    }
+}
+
+TopoDS_Shape occtUnifySameDomain(const TopoDS_Shape& shape,
+                                 bool unifyEdges, bool unifyFaces, bool concatBSplines) {
+    try {
+        BRepBuilderAPI_Copy copier;
+        TopoDS_Shape work = occtUnifySameDomainInput(shape, copier);
+        if (work.IsNull()) return TopoDS_Shape();
+
+        ShapeUpgrade_UnifySameDomain unifier(work, unifyEdges, unifyFaces, concatBSplines);
+        unifier.Build();
+        return unifier.Shape();
+    } catch (...) {
+        return TopoDS_Shape();
+    }
+}
+
 OCCTShapeRef OCCTShapeUnifySameDomain(OCCTShapeRef shape,
                                        bool unifyEdges, bool unifyFaces,
                                        bool concatBSplines) {
     if (!shape) return nullptr;
 
     try {
-        ShapeUpgrade_UnifySameDomain unifier(shape->shape, unifyEdges, unifyFaces, concatBSplines);
-        unifier.Build();
-
-        TopoDS_Shape result = unifier.Shape();
+        TopoDS_Shape result = occtUnifySameDomain(shape->shape, unifyEdges, unifyFaces, concatBSplines);
         if (result.IsNull()) return nullptr;
 
         return new OCCTShape(result);
@@ -418,10 +451,10 @@ OCCTShapeRef OCCTShapeSimplify(OCCTShapeRef shape, double tolerance) {
     if (!shape) return nullptr;
 
     try {
-        // First unify same domain
-        ShapeUpgrade_UnifySameDomain unifier(shape->shape, true, true, true);
-        unifier.Build();
-        TopoDS_Shape unified = unifier.Shape();
+        // First unify same domain (#446: on a private copy — the caller's shape is not an input
+        // this algorithm may consume)
+        TopoDS_Shape unified = occtUnifySameDomain(shape->shape, true, true, true);
+        if (unified.IsNull()) return nullptr;
 
         // Then heal the shape
         Handle(ShapeFix_Shape) fixer = new ShapeFix_Shape(unified);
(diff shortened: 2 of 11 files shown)


gsdali (commented):
## Overview

Fixes #446: `ShapeUpgrade_UnifySameDomain` rewrites sub-shapes of the shape it is handed, and because the caller's `Shape` shares those `TShape`s, the standard "take the merge if valid, else keep what I had" idiom silently damaged what you had. All three bridge entry points (`OCCTShapeUnifySameDomain`, `OCCTShapeSimplify`, the `OCCTUnifySameDomain` builder) now go through shared helpers that unify a private `BRepBuilderAPI_Copy`, with `keepShape` sub-shapes mapped into that copy.

## What I verified independently

Compiled ground-truth C++ probes against the pinned p1 xcframework rather than taking the PR body at its word:

- **The bug and the fix both reproduce.** Two stacked coaxial cylinders: input BREP `1656 -> 1758` bytes across a direct unify (MUTATED), `1656 -> 1656` through the copy, merge result unchanged at 4 faces -> 3. Premise and remedy both hold.
- **Root-cause citation is accurate.** `ShapeUpgrade_UnifySameDomain.cxx:1228` is indeed `BB.UpdateEdge(anEdge, NewPCurves[0], theRefFace, 0.)` followed by `MapEdgesWithTemporaryPCurves.Add(anEdge)`, and the only two sites that null those pcurves out again (`:3657`, `:3790`) sit inside branches that first do `RefFace.Nullify()`. `mySafeInputMode` is `true` in both constructors and never guards `TransformPCurves`. The `OCCTBridge_Internal.h` comment is correct as written.
- **The `keepShape` mapping is load-bearing.** `copier.ModifiedShape(callerEdge)` returns a shape that is *not* `IsSame` as the caller's edge, so without the mapping every `keepShape` would keep nothing. `ModifiedShape` does throw for a foreign shape, so the `catch (...)` in `occtUnifySameDomainMapped` is required, not defensive noise.
- **`copyMesh` defaults to false, but that costs nothing here.** I expected the copy to drop triangulation from meshed inputs. Measured: the unify result has 0/3 triangulated faces both with and without the copy, because the algorithm rebuilds faces regardless. No regression, no reason to pass `copyMesh=true`.
- **Coverage is complete.** Only three `ShapeUpgrade_UnifySameDomain` construction sites exist in the bridge and all three now route through the helpers. `OCCTBridge_Healing.mm` already includes `BRepBuilderAPI_Copy.hxx` (line 129), so the new definitions compile; the added include in `OCCTBridge_Modeling.mm` is needed for the by-value member.

## Findings

**1. Undocumented identity change (medium, docs).** For an input where unify finds nothing to merge, the old result *was* the input: `result.IsSame(box) == 1`, 6/6 faces shared. Now it is an independent copy: `IsSame == 0`, 0/6 faces shared. That is the unavoidable price of the fix, but docs and changelog only promise "the input is not modified" and never say the result no longer shares sub-shapes with it. `Shape.isSame(as:)`, `isPartner(with:)` and `isEqual(to:)` are public, and downstream consumers map selections and attributes by sub-shape identity. Worth one sentence in `Shape.unified()`'s doc comment and the changelog, plus a test pinning it (`#expect(box.unified()?.isSame(as: box) == false)`).

**2. The "fraction of the merge" claim is optimistic for the no-op case (low, changelog wording).** Measured over a 600-face compound: copy 2.4 ms vs unify 13.7 ms, so 18% when there is real merging work. On a nothing-to-merge input: copy 5.6 ms vs unify 8.7 ms, so 64%. Since `unified()` is the standard post-boolean cleanup and frequently finds nothing, and peak memory now doubles, I would soften that line rather than add the `copyInput:` flag (agreed with leaving it out). For the builder, the copier plus its modifier map stays alive for the object's lifetime, which is the correct trade for `keepShape` but is worth a word in the struct comment.

**3. Leak on the throwing path, now much bigger (low, `OCCTBridge_Modeling.mm:8749`).** `auto result = new OCCTUnifySameDomain();` is scoped inside the `try`, so if `new ShapeUpgrade_UnifySameDomain(...)` throws, `result` leaks. Pre-existing, but the leak used to be one pointer and is now a full shape copy plus its map. `std::unique_ptr<OCCTUnifySameDomain> result(new ...); ... return (OCCTUnifySameDomainRef)result.release();` closes it in one line.

**4. One more `nullptr` through a `_Nonnull` contract (low).** `OCCTUnifySameDomainCreate` is declared `_Nonnull` and gains a new null-return path (copy failure); Swift stores it into a non-optional `let ref: OCCTUnifySameDomainRef`. It degrades safely in practice because every accessor null-checks and `Release` tolerates null, so this is not a crash. But #345 set the opposite precedent for exactly this shape of call. Either mark it `_Nullable` and make the Swift `ref` optional, or leave it and note in the header that null is possible.

**5. Docs discipline gap (low).** `docs/reference/Shape-Features.md` is the reference page for the two changed `Shape` methods (`unified(...)` at line 1385, `simplified(...)` at 1423) and was not touched, while the builder's page in `Document-Completions.md` was. CLAUDE.md requires the reference page for a changed type to move in the same PR. `docs/guides/cookbook/healing-and-validity.md` is the other natural home for the guarantee, since it is where the post-boolean `unified()` idiom is taught. While in that file: it credits `withoutSmallFaces` to `ShapeAnalysis_CheckSmallFace` + `ShapeUpgrade_UnifySameDomain`, but `OCCTShapeRemoveSmallFaces` actually uses `BRepAlgoAPI_Defeaturing`. Pre-existing, cheap to fix in passing.

**6. Test notes (low).**

- `declinedMergeLeavesInputIntact` is self-admittedly non-discriminating on this fixture and pays for two `isSelfIntersecting(hardTimeout: 20)` calls to prove nothing the byte check does not already prove.
- `keepShapeSurvivesTheCopy` asserts `blockedCount > 0`, which would also pass if some unrelated mechanism blocked a merge. Asserting the count matches the seam-edge expectation would tighten it.
- Nothing asserts the copy leaves the *merged result* geometrically identical. `#expect(abs(merged.volume - body.volume) < 1e-9)` is one line and guards the thing a copy could plausibly perturb.
- The byte-fingerprint helper is a good tool; consider hoisting it somewhere shared so sibling ops can be audited the same way.

**7. Sibling audit, for the issue rather than this PR (info).** I spot-checked the same input-consumption class on the same fixture for `ShapeFix_Shape` (`healed()`/`fixed()`), `BRepAlgoAPI_Defeaturing` (`withoutSmallFaces`) and `ShapeUpgrade_ShapeDivideClosed`: all three left the input byte-identical. No obvious follow-up hiding next door, worth recording in #446 so nobody files a speculative sweep.

## Conventions and quality

Helper naming (`occtUnifySameDomain*`) matches the file's existing `occt*` internal convention, the `OCCTBridge_Internal.h` block comment matches the style set by the #430 block, and the tests avoid force-unwrapping inside `#expect` as required. The changelog entry is correctly filed as "Unreleased" with no predicted version. The deduplication is the right call: three copies of construct/`Build()`/null-check is exactly why this fix would otherwise have been written three times, and the surviving `Shape.unified()` / builder / `simplified()` split is justified in the PR body rather than assumed.

## Verdict

Correct fix, correctly root-caused, and the mechanism holds up under independent verification. Nothing here blocks merge. I would land items 1 and 5 (one sentence each, plus the reference page) before tagging, since they are the parts a downstream consumer will actually trip over, and fold 3 in while you are in the file.

PR #458, opened 2026-07-28
fix(#409): Curve2D.arcLength(from:to:) returns nil on failure, not 0.0

gsdali:
## What & why

`Curve2D` exposes the same arc-length computation twice. `length(from:to:)` (backed by
`OCCTCurve2DGetLengthBetween`) already returns `Double?`, mapping a `-1.0` bridge sentinel to
`nil` on failure. `arcLength(from:to:)` (backed by `OCCTCurve2DLength`) returned a plain,
non-optional `Double`, mapping the exact same failure conditions to a sentinel `0.0` — silently
indistinguishable from a curve whose arc length is genuinely zero.

Reading both bridge implementations directly (not just the auditor's "cosmetic duplication"
summary) turned up a real, previously-undocumented divergence beyond the sentinel: `arcLength`'s
bridge call constructs a range-checked `Geom2dAdaptor_Curve(curve, u1, u2)`, which raises
`Standard_ConstructionError` when `u1 > u2`; `length`'s bridge call builds an *unrestricted*
adaptor and calls the `Length(adaptor, u1, u2)` overload, which tolerates either order and
returns the same magnitude regardless (verified with a ground-truth C++ probe against the pinned
xcframework). So the two functions are not fully redundant — `arcLength` enforces an ordering
contract `length` does not — and consolidating them onto one call would silently change that
contract. `OCCTCurve2DLength` keeps its own bridge call; only its failure sentinel changes, from
`0` to `-1.0`, matching every other length-reporting bridge function in this file.

Refs #409. Part of the #380 / #377 duplication audit — targets `refactor/377-segmented-audit`,
not `main`.

## Checklist

      verification)

New suite `Curve2DArcLengthFailureTests` (`Tests/OCCTGeom2dTests`): a reversed range
(`u1 > u2`) is a genuine failure (reported as `-1.0`, not `0.0`), an equal-bounds range is a
genuine zero-length result (not the failure sentinel), and `length(from:to:)` is shown tolerating
the exact reversed range `arcLength(from:to:)` rejects. The existing happy-path test
(`curve2DArcLength`, `Tests/OCCTCurveTests`) is unchanged — `arcLength(from:to:)` keeps its
original non-optional signature (see "Update after review" below).

## Notes for the reviewer

**Verified the bug was live**, not just theoretically reachable: temporarily reverting only the
bridge sentinel (`-1.0` back to `0`) and re-running the new suite reproduces the exact failure —
`arcLength(from: 0.8, to: 0.2) → 0.0` where `nil` was expected — then restored.

**Why not delegate `arcLength` to `length`.** That was my first instinct reading the issue (both
ultimately call `GCPnts_AbscissaPoint::Length` over a `Geom2dAdaptor_Curve`), but a ground-truth
probe against the real xcframework showed `length(from:to:)`'s unrestricted-adaptor idiom
happily computes a positive length for a reversed range — it does not fail at all for the input
that makes `arcLength(from:to:)` fail. Delegating would have silently dropped `arcLength`'s order
validation rather than just fixing its sentinel. Keeping the bridge calls separate, with a shared
failure contract (`nil`), preserves both computations' actual behavior.

**Relationship to #408 (now #461, merged).** #461 is the mirror fix for `Curve3D`
(`totalArcLength`, `arcLength(from:to:)`, `arcLengthBetween`). It deliberately kept those three
non-optional and collapsed failure to a `-1.0` sentinel, specifically to avoid a source-breaking
signature change. This PR originally converted `Curve2D.arcLength(from:to:)` to `Double?` instead
— a real inconsistency, corrected below after review. `Curve2D.length`/`length(from:to:)` remain
the optional, failure-distinguishing entry points on both types; `arcLength(from:to:)` stays a
non-optional `-1.0`-sentinel convenience wrapper on both types, though (unlike `Curve3D`)
`Curve2D`'s wrapper keeps its own bridge call rather than delegating to `length(from:to:)`, per
the ordering-contract divergence above.

Full `swift test` (4515 tests) clean except one pre-existing flake documented in
`CLAUDE.md`/the test's own source comment: `"Shape.loadRobust interrupts repair, not just the
transfer (#300)"`, a CPU-timing-sensitive cancellation test that fails only when running beside
its IGES sibling under parallel load — unrelated to this change.

unit, and parallel child PRs editing the same changelog header would conflict on every one.

## Update after review

Per [review feedback](https://github.com/SecondMouseAU/OCCTSwift/pull/458#pullrequestreview-4802255752):
reworked `arcLength(from:to:)` to stay **non-optional**, matching `Curve3D.arcLength(from:to:)`'s
shape from #461 exactly — the `Double?` return type was a MAJOR-triggering breaking change
(`docs/SEMVER.md`) and broke parity with the already-merged mirror fix. It now returns `-1.0` on
failure instead of the original `0.0`. The range-check-divergence finding (this PR's real
contribution) is unchanged: `arcLength(from:to:)` still keeps its own bridge call rather than
delegating to `length(from:to:)`, since the two use genuinely different adaptor constructions
(range-checked vs. unrestricted) and delegating would silently drop the order validation. Updated
the new test suite and `docs/reference/Curve2D-Analysis.md` to match; reverted the happy-path test
in `Tests/OCCTCurveTests` to its original (pre-PR) form since the signature no longer changed.
Rebased onto the current `refactor/377-segmented-audit` (picks up #461) so the diff is directly
comparable. `swift test --filter OCCTGeom2dTests` (398 tests) and `swift test --filter
OCCTCurveTests` (328 tests) both clean.

### DIFF
--- a/Sources/OCCTSwift/Curve2D.swift
+++ b/Sources/OCCTSwift/Curve2D.swift
@@ -2354,13 +2354,17 @@ extension Curve2D {
         return Curve2D(handle: ref)
     }
 
-    /// Compute the arc length of this curve between parameters `u1` and `u2`. Unlike
-    /// `length(from:to:)`, `u1` must not exceed `u2`. Returns `nil` on failure (a reversed
-    /// range, a released curve, or any other computation error) rather than a sentinel `0.0`
-    /// that could be mistaken for a genuine zero-length segment.
-    public func arcLength(from u1: Double, to u2: Double) -> Double? {
+    /// Compute the arc length of this curve between parameters `u1` and `u2` (non-optional).
+    ///
+    /// Unlike `length(from:to:)`, `u1` must not exceed `u2` — the underlying computation uses a
+    /// range-checked adaptor that fails on a reversed range. Returns `-1.0` on failure (a
+    /// reversed range, a released curve, or any other computation error) — arc length is
+    /// otherwise always non-negative, so this is an unambiguous failure sentinel, never
+    /// confusable with a genuine zero-length segment (e.g. `u1 == u2`). Use `length(from:to:)`
+    /// directly if you need an optional rather than a sentinel value (and order tolerance).
+    public func arcLength(from u1: Double, to u2: Double) -> Double {
         let l = OCCTCurve2DLength(handle, u1, u2)
-        return l >= 0 ? l : nil
+        return l >= 0 ? l : -1.0
     }
 
     /// Split this curve at C1 discontinuities.
--- a/Sources/OCCTSwift/Curve2D.swift
+++ b/Sources/OCCTSwift/Curve2D.swift
@@ -450,7 +450,10 @@ public final class Curve2D: @unchecked Sendable {
         return l >= 0 ? l : nil
     }
 
-    /// Arc length between two parameter values.
+    /// Arc length between two parameter values. Unlike `arcLength(from:to:)`, `u1` may be
+    /// greater than `u2` (order doesn't affect the result). Returns `nil` on failure (e.g. a
+    /// released curve), never a sentinel value that could be mistaken for a real zero-length
+    /// segment.
     public func length(from u1: Double, to u2: Double) -> Double? {
         let l = OCCTCurve2DGetLengthBetween(handle, u1, u2)
         return l >= 0 ? l : nil
@@ -2351,9 +2354,13 @@ extension Curve2D {
         return Curve2D(handle: ref)
     }
 
-    /// Compute the arc length of this curve between parameters u1 and u2 (non-optional).
-    public func arcLength(from u1: Double, to u2: Double) -> Double {
-        OCCTCurve2DLength(handle, u1, u2)
+    /// Compute the arc length of this curve between parameters `u1` and `u2`. Unlike
+    /// `length(from:to:)`, `u1` must not exceed `u2`. Returns `nil` on failure (a reversed
+    /// range, a released curve, or any other computation error) rather than a sentinel `0.0`
+    /// that could be mistaken for a genuine zero-length segment.
+    public func arcLength(from u1: Double, to u2: Double) -> Double? {
+        let l = OCCTCurve2DLength(handle, u1, u2)
+        return l >= 0 ? l : nil
     }
 
     /// Split this curve at C1 discontinuities.
--- a/Sources/OCCTBridge/include/OCCTBridge.h
+++ b/Sources/OCCTBridge/include/OCCTBridge.h
@@ -17478,7 +17478,8 @@ double OCCTCurve3DClosestParameter(OCCTCurve3DRef _Nonnull curve, double px, dou
 /// Create a trimmed copy of a 2D curve between parameters u1 and u2.
 OCCTCurve2DRef _Nullable OCCTCurve2DTrimmed(OCCTCurve2DRef _Nonnull curve, double u1, double u2);
 
-/// Compute the length of a 2D curve between parameters u1 and u2.
+/// Compute the length of a 2D curve between parameters u1 and u2 (u1 must not exceed u2).
+/// Returns -1.0 on failure (null curve, reversed range, or any other computation error).
 double OCCTCurve2DLength(OCCTCurve2DRef _Nonnull curve, double u1, double u2);
 
 // --- Surface additional (v0.115.0) ---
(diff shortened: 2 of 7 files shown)


Sources/OCCTSwift/Curve2D.swift
```diff
@@ -2272,9 +2275,13 @@ extension Curve2D {
...
-    public func arcLength(from u1: Double, to u2: Double) -> Double {
-        OCCTCurve2DLength(handle, u1, u2)
+    /// Compute the arc length of this curve between parameters `u1` and `u2`. Unlike
+    /// `length(from:to:)`, `u1` must not exceed `u2`. Returns `nil` on failure (a reversed
+    /// range, a released curve, or any other computation error) rather than a sentinel `0.0`
+    /// that could be mistaken for a genuine zero-length segment.
+    public func arcLength(from u1: Double, to u2: Double) -> Double? {
+        let l = OCCTCurve2DLength(handle, u1, u2)
+        return l >= 0 ? l : nil
     }
```
gsdali: **Blocking: please rework this to follow #461's pattern before merge.**

This changes `arcLength(from:to:)`'s return type from `Double` to `Double?` — a genuine, unambiguous source-breaking signature change. Per `docs/SEMVER.md`, "a changed return type" is explicitly listed as a MAJOR-bump trigger, no interpretation required (unlike the more debatable "same signature, different behavior" cases in #460/#461).

This is also inconsistent with the mirror fix in #461 (`Curve3D`, already merged), which deliberately kept `arcLength(from:to:)` **non-optional** and delegated to `length(from:to:)` with a `-1.0` sentinel specifically to avoid breaking existing callers — #461's own PR body states: *"Converting `totalArcLength`/`arcLength(from:to:)`/`arcLengthBetween(_:_:)` to optionals would be the 'correct' shape... but it's a source-breaking signature change to three existing public non-optional methods."*

This PR's description claims the two fixes will "read as the same shape in both types," but they don't: `Curve3D.arcLength(from:to:)` returns `Double` (merged), `Curve2D.arcLength(from:to:)` returns `Double?` (this PR). Please rework this to match #461's shape:

```swift
public func arcLength(from u1: Double, to u2: Double) -> Double {
    let l = OCCTCurve2DLength(handle, u1, u2)
    return l >= 0 ? l : -1.0
}
```

(or equivalent — keep the non-optional signature, `-1.0` sentinel on failure, same as `Curve3D`). The range-check-divergence finding (this PR's real contribution — `arcLength`'s adaptor rejects `u1 > u2`, `length`'s doesn't) is independent of this and should be kept as-is; only the return-type change needs to be reverted. Please also correct the PR description, which currently asserts parity with #408/#461 that doesn't actually hold.

Holding this PR until the signature is reverted to non-optional — please ping for another pass once updated.

gsdali (commented):
Reviewed. The underlying investigation is solid — the finding that `arcLength(from:to:)`'s adaptor is range-checked (rejects `u1 > u2`) while `length(from:to:)`'s isn't, verified with a ground-truth C++ probe, is genuinely good diligence and correctly stopped this from being a naive "just delegate" fix that would have silently dropped that validation. The bridge sentinel fix (`0` → `-1.0`) is scoped exactly as claimed, no other in-repo call sites were missed, and the new `Curve2DArcLengthFailureTests` suite exercises real behavior rather than hardcoded literals — ran it independently, 6/6 pass.

**🛑 Holding this PR** on one issue, flagged inline on the `arcLength(from:to:)` signature: this PR changes the return type to `Double?`, which is a MAJOR-triggering breaking change per this repo's own `docs/SEMVER.md` ("a changed return type"), and it's inconsistent with the already-merged #461 (`Curve3D`'s mirror fix), which deliberately kept `arcLength(from:to:)` non-optional specifically to avoid that break. The PR description claims the two fixes read as "the same shape" — they don't, as landed.

Please rework `arcLength(from:to:)` to stay non-optional and delegate to a `-1.0` sentinel on failure, matching `Curve3D.arcLength(from:to:)`'s exact shape from #461 (see the inline comment for the concrete diff). The range-check-divergence discovery itself is correct and should stay — only the return-type change needs reverting. Also please correct the PR description's parity claim once that's done. Ping me for another pass after the update.

Sources/OCCTSwift/Curve2D.swift
gsdali: Reworked as requested: `arcLength(from:to:)` is back to non-optional `Double`, matching `Curve3D.arcLength(from:to:)` from #461 exactly, with `-1.0` as the failure sentinel instead of the original ambiguous `0.0`. The order-validation finding (the actual bug this PR fixes) is untouched — `arcLength` still keeps its own bridge call rather than delegating to `length(from:to:)`, since the two adaptors have genuinely different range-tolerance behavior.

Updated `Curve2DArcLengthFailureTests` and `docs/reference/Curve2D-Analysis.md` to match, reverted the `Tests/OCCTCurveTests` happy-path test to its original form (no longer needs an `if let`), rebased onto current `refactor/377-segmented-audit` (picks up #461) so the diff is directly comparable, and corrected the PR description's parity claim. `swift test --filter OCCTGeom2dTests` (398 tests) and `swift test --filter OCCTCurveTests` (328 tests) both clean. Ready for another pass.

PR #473, opened 2026-07-29
fix(#433, #434): FillingSurface mis-mapped continuity, converge onto Shape.fill's implementation

gsdali:
## What & why

`FillingSurface.add(edge:continuity:)`/`add(freeEdge:continuity:)` hand-mapped the plate
constraint order onto `GeomAbs_Shape` locally instead of sharing `occtFillingContinuityToGeomAbs`
(the helper #430 introduced for `Shape.fill`): order 1 requested `GeomAbs_C1` (curvature) instead
of `GeomAbs_G1` (tangency), and order 2 requested `GeomAbs_C2` (ordinal 4), which every
constraint class rejects — failing the whole `build()` rather than just that one constraint
(#433). `add` returned `true` regardless, since `Add` only appends and never validates the order.

Root cause was architectural, not a typo: `FillingSurface` held its own `BRepFill_Filling`
instead of the `BRepOffsetAPI_MakeFilling` `Shape.fill` already used — the same class,
forwarding to a private `BRepFill_Filling` it never exposes. #434 converges the two onto one
implementation: `FillingSurface` now builds the same `BRepOffsetAPI_MakeFilling` through a
shared `occtFillingMakeBuilder`, and both entry points share `occtFillingAddConstraint`
outright — no longer templated, since both callers now hold the same concrete filler type —
fixing #433 as a consequence of the convergence rather than as a separate patch.

Also adds `FillingSurface.add(edge:support:continuity:)`, matching `FillConstraint`'s
support-face semantics (used or the constraint fails, never silently substituted) — the one
capability `FillingSurface` was missing relative to `Shape.fill(constraints:)`.

Bridge-only change: no OCCT kernel patch, no `OCCT.xcframework` rebuild.

Closes #433
Closes #434

## Checklist

- [x] New or changed behavior is covered by a unit test in the same PR — two new tests in
      `FillingSupportFaceTests` (`OCCTSurfaceTests.swift`): one guards the `.g1`/`.g2` mapping
      fix directly (mirrors `Shape.fill`'s own equivalent regression test), the other covers
      `add(edge:support:continuity:)`'s nominated-support-face-must-work-or-fail contract.

## Notes for the reviewer

- Targets `refactor/377-segmented-audit`, not `main` — same branch #430/#431/#432/#398 landed
  on, per #377's ordering rule.
- Deliberately did **not** add per-constraint `G0Error(index)`/`G1Error(index)`/`G2Error(index)`
  accessors or a `Generated()`-based full-history variant, even though #434 notes
  `FillingSurface` "gains" both structurally from this refactor. Neither was asked for
  explicitly, both are non-trivial new public surface (the error accessors need the constraint
  index `Add()` returns, which nothing currently threads through), and #327's own
  `*WithFullHistory` family is the natural home for history integration if that's wanted later.
  Flagging so it isn't mistaken for an oversight.
- Docs updated in the same PR: `docs/reference/GeometrySolvers.md` (rewrote the `FillingSurface`
  section), `docs/reference/Shape-Features.md` (cross-reference + removed a now-stale note),
  `docs/CHANGELOG.md`, `Scripts/repro/430-fill-untrimmed-pcurve/README.md` (status table — also
  caught that its #432 row had been stale as "open" since the moment it was written, same PR
  that fixed it), and `CLAUDE.md`. Operation count bumped via `Scripts/count-operations.py --fix`
  (4,262 → 4,266); README/API_REFERENCE were already 4 behind before this PR's own +1, from
  drift elsewhere on this branch — `--fix` corrects both.
- Verified: `swift build` clean (bridge recompiles from source — this worktree had
  `OCCTSWIFT_BRIDGE_PREBUILT=1` set, which would have silently skipped the `.mm` changes if left
  on), full `swift test` — **4585 tests in 1312 suites, 0 failures**.


gsdali (commented):
## Overview

This converges `FillingSurface`'s bridge implementation onto the same `BRepOffsetAPI_MakeFilling` that `Shape.fill` already uses (#434), fixing `FillingSurface`'s continuity mis-mapping as a side effect (#433: `.g1` requested `GeomAbs_C1`/curvature instead of `GeomAbs_G1`/tangency, and `.g2` requested `GeomAbs_C2`, ordinal 4, which every constraint class rejects, failing the whole `build()`). Concretely:

- `OCCTFilling` now holds a `BRepOffsetAPI_MakeFilling` instead of its own `BRepFill_Filling`.
- `occtFillingMakeBuilder` is extracted from `OCCTShapeFillMakeBuilder` (Healing.mm) and shared by both entry points, with `Shape.fill` continuing to hard-code `degree=3, nbPtsOnCur=15` (not exposed on `FillingParameters`) while `FillingSurface`'s own constructor still allows callers to set them.
- `occtFillingAddConstraint` drops its template parameter now that both callers hold the same concrete filler type.
- `OCCTFillingGetFace` switches from `BRepFill_Filling::Face()` to `BRepBuilderAPI_MakeShape::Shape()` (`BRepOffsetAPI_MakeFilling` doesn't expose `Face()`).
- New `FillingSurface.add(edge:support:continuity:)` / `OCCTFillingAddEdgeWithSupport`, giving `FillingSurface` `FillConstraint`'s explicit-support-face capability.
- Docs (README, API_REFERENCE, CHANGELOG, `GeometrySolvers.md`, `Shape-Features.md`, `Continuity.swift`, the #430 repro README, `CLAUDE.md`) all updated in the same PR — good discipline, and the repro README fix (a stale "#432 open" row corrected) is a nice catch.

The convergence itself is clean: `occtFillingContinuityToGeomAbs`/`occtFillingAddConstraint`/`occtFillingMakeBuilder` are lifted unchanged (only de-templated/generalized), the `#431` argument-binding fix is correctly threaded through the new shared builder in the same parameter order, and the guaranteed-copy-elision return-by-value pattern is preserved.

## Finding: `add(edge:support:continuity:)`'s default makes its own headline guarantee false (medium-high confidence)

```swift
// FillingSurface.swift
public func add(edge: Edge, support: Face, continuity: SurfaceContinuity = .g0) -> Bool {
    OCCTFillingAddEdgeWithSupport(handle, edge.handle, support.handle, continuity.rawValue)
}
```

and, in `occtFillingAddConstraint` (unchanged logic, shared by both entry points):

```cpp
if (order != GeomAbs_C0) {
    // ... support is read, validated against the edge's pcurve, used or rejected ...
}
filling.Add(edge, order, isBound);   // GeomAbs_C0: support is never even looked at
return true;
```

At `.g0` (`GeomAbs_C0`), `support` is not read at all — not validated, not used, not even null-checked. Since `continuity` defaults to `.g0`, `filling.add(edge: e, support: someUnrelatedFace)` — the natural way to call a method whose only new parameter has a default — silently succeeds and silently ignores `someUnrelatedFace` entirely. That directly contradicts the doc comment's headline claim:

> Mirrors `FillConstraint`'s support-face semantics: a face named here is used or the constraint fails, rather than silently falling back to another surface.

`FillConstraint` itself (`Shape.swift:3371`) avoids exactly this trap by defaulting `continuity` to `.g1`, not `.g0` — presumably deliberately, since a support-face parameter is meaningless at G0. `add(edge:support:continuity:)` claims to mirror that type's semantics but not its default, so it inherits the meaninglessness without inheriting the guard against it.

Neither new test exercises this path — both explicitly pass `continuity: .g1`/`.g2`, so the default-parameter case is untested as well as undocumented:

```swift
#expect(tangentFilling.add(edge: rim, support: wall, continuity: .g1))     // explicit .g1
#expect(!filling.add(edge: rim, support: strangerFace, continuity: .g1))   // explicit .g1
```

Suggest changing the default to `.g1` (matching `FillConstraint`'s precedent) — that alone makes the "used or fails" claim true for the common zero-argument call, and a caller who genuinely wants G0-with-a-support-face still gets it by passing `continuity: .g0` explicitly, at which point silently ignoring `support` is at least *visible* in the call site rather than implied by an unrelated default.

## Minor notes

- **Parity claim is slightly overstated.** The CHANGELOG frames `add(edge:support:continuity:)` as "the one capability `FillingSurface` was missing relative to `Shape.fill(constraints:)`" — but `FillConstraint.isBoundary` covers free edges too, while this PR only adds the support-face overload for boundary edges (no `add(freeEdge:support:continuity:)`). Not a blocker, just worth scoping the claim accurately or filing the free-edge variant as a follow-up if it's wanted.
- `new OCCTFilling{ occtFillingMakeBuilder(...) }` in `OCCTFillingCreate` relies on the same C++17 guaranteed-copy-elision guarantee that `occtFillingMakeBuilder`'s own return-by-value already documents at length a few lines away — worth a one-line pointer back to that comment at this call site, purely for the next reader, since the codebase otherwise never leaves a non-obvious initialization trick unexplained.

## Risk / Test Coverage

No OCCT kernel change, no thread-safety-sensitive state, matches the project's "bridge-only" pattern for this line of work (#430–#434). Existing filling tests continue to exercise the shared helpers unchanged. The two new tests are well-targeted at #433's core defect (curvature measurably differs from tangency, not just "didn't crash") and at the nominated-support-must-work-or-fail contract — the gap is specifically the untested default-parameter path noted above.

gsdali (commented):
## Re-review note

This is an updated revision (commit `283bbc39`, "address review — `add(edge:support:continuity:)` defau..."). All three points from my previous review comment are addressed, confirmed by reading the current diff:

1. **Default-`.g0`-makes-`support`-inert.** Fixed by changing the default to `.g1`:
   ```swift
   public func add(edge: Edge, support: Face, continuity: SurfaceContinuity = .g1) -> Bool {
   ```
   matching `FillConstraint`'s own default and its underlying reason. A new, well-targeted test (`fillingSurfaceAddWithSupportDefaultsToG1`) exercises exactly the previously-untested path: calling `add(edge:support:)` with no `continuity` argument now both accepts a real support face (and produces a measurably tangent, non-flat cap) and rejects an unrelated one — pinning the "used or fails" claim at the actual default, not just at an explicitly-passed `.g1`/`.g2`. The doc comment on the method itself now explains the "why," and `docs/reference/GeometrySolvers.md`'s signature block was updated to match (`= .g1`), so code/tests/both doc locations are consistent.
2. **Overstated parity claim.** Fixed — the CHANGELOG entry now explicitly scopes it: "Covers the boundary-edge case; `FillConstraint.isBoundary` also covers free edges with a named support face, which this PR does not add an equivalent for."
3. **Missing copy-elision comment pointer.** Fixed — `OCCTFillingCreate`'s aggregate-init call site now has its own one-line comment pointing back to `occtFillingMakeBuilder`'s copy-elision explanation.

## Everything else from the original review still stands

- The convergence itself (shared `occtFillingMakeBuilder`/`occtFillingAddConstraint`/de-templating) is unchanged and was already clean.
- `OCCTFillingGetFace`'s switch from `Face()` to `Shape()` is still correct and exercised by the full existing filling test surface.
- No OCCT kernel change, no thread-safety-sensitive state.

## Minor, unconfirmed note

The PR description's "Verified" line still says `swift test` — **4585 tests in 1312 suites** — but this round added one more test (`fillingSurfaceAddWithSupportDefaultsToG1`) after that line was presumably last written, and nothing in this round's commit touches the PR description itself. Worth a quick recount before merge so the description matches what actually ships — I haven't independently re-run the suite for this pass since the branch is already checked out in another active worktree and this was a diff-only review.

Nothing here blocks merge. Good iteration — the new test targets the exact gap the previous review identified rather than just re-asserting the already-covered explicit-continuity cases.

PR #475, opened 2026-07-29
chore(#440): silence the unhandled-file warning on every root build

gsdali:
Closes #440.

## The warning, and what actually caused it

`swift build` emitted:

```
warning: 'OCCTSwift': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
```

The file is `Tests/OCCTStressTests/Fixtures/unify-crash-mmd-kiha10-body5.brep`, the 600 KB
mesh-sewn solid backing #348's null-pcurve regression test. It is read straight from the source
tree via `#filePath`, never through `Bundle.module`, so it is neither a build input nor a resource
to copy. `OCCTStressTests` now declares `exclude: ["Fixtures"]`, which is the accurate description
of what it is. Declaring it as a resource instead would have embedded 600 KB in the test bundle
that nothing reads.

## Who actually saw it (corrected in review)

An earlier revision of this PR claimed the warning fired for every downstream consumer building
OCCTSwift from source, and made that the main reason to fix it. **That was wrong**, and it repeated
a claim from #440.

The fixture is in a *test* target, and SwiftPM constructs no test targets for a non-root package,
so a consumer's build never had a target owning `Tests/` to diagnose. Measured on Swift 6.3.3
against a synthetic package:

| stray unhandled file in a dependency | dependency built as root | path-dep consumer (`swift build` / `swift test`) |
|---|---|---|
| `Tests/DepTests/Fixtures/fixture.brep` | warns | **silent** |
| `Sources/Dep/stray.brep` | warns | warns |

So the real audience was root builds: our own dev loop and CI, plus anyone who clones OCCTSwift and
builds it directly. That is still worth fixing on its own, and the fix is unchanged; only the
justification was overstated.

## Correcting #440's diagnosis of the cause

#440 attributed the warning to `Tests/occt_parallel_crash_portable.cpp` sitting unclaimed at the
`Tests/` root. That file is not the cause and never was: no target declares `path: "Tests"`, so
SwiftPM does not scan the directory root at all and never diagnosed it. Verified by moving the
`.cpp` out of the tree and rebuilding, which leaves the same warning naming the same single `.brep`.
The fixture landed 2026-07-22 (e10eef3) and the issue was filed 2026-07-27, so the misattribution
was present from filing.

## The move, to a different destination than suggested

Still worth doing on its own merit, since a reproducer does not belong in the package's test tree,
but `Scripts/repro/342-boolean-ops/` would have mislabelled it. That file is the
[OCCT#1179](https://github.com/Open-Cascade-SAS/OCCT/issues/1179) parallel-crash sweep: 11
operation groups built around `Extrema_ExtElCS` and `ShapeUpgrade_FaceDivide`, committed April 2026,
three months before #342, with one boolean group out of eleven. It is not the same family of work.

It now has `Scripts/repro/occt1179-parallel-crash/`, matching the convention the other 11 repro
directories use, with a README covering what it sweeps, why it is kept now that OCCT#1179 is fixed
(our own [OCCT#1203](https://github.com/Open-Cascade-SAS/OCCT/pull/1203), shipped in `V8_0_0`), and
the property that distinguishes it from its boolean-concurrency siblings: it is the only reproducer
here driven by CI, on both Windows and macOS. The two paths in
`.github/workflows/occt-parallel-crash-test.yml` follow it. The `.cpp` itself is unchanged (100%
rename).

**Also corrected in review:** that README's history section claimed the `.cpp` arrived in `e32c8ba`
(#379, an unrelated `Edge.circleProperties` fix), and inferred from that its `Tests/` placement was
incidental. `e32c8ba` does not touch the file. It was added in `d239bd7` (2026-04-14 12:35:50), 13
seconds after the workflow that references it (`e174ae6`, 12:35:37), which already named
`tests\occt_parallel_crash_portable.cpp` on both platforms, and was restructured in `103ed61`
two days later. `git log --follow` shows only those two commits, and the only rename in its history
is this PR's. So the location was deliberate, and the README now says so.

## Note on the diff size

`Package.swift` also loses its 11 pre-existing em-dashes, per OKF-STANDARD.md's writing-style rule
("clear them from any file you are already editing"). Comments only, no semantic change.

## Verification

Re-run after rebasing onto `main` at v1.17.0:

- `swift build`: clean, zero warnings, 3 consecutive runs.
- `swift build --build-tests`: all 18 test targets, zero warnings.
- Negative control: removing `exclude: ["Fixtures"]` reproduces the warning naming exactly that
  `.brep`; restoring it clears it. So the exclude is load-bearing, not incidentally silent.
- `swift test --filter StressUnifySameDomainNullPCurveTests`: passes, so `exclude:` does not break
  the `#filePath` fixture read.
- `swift package dump-package`: manifest parses.
- Workflow YAML still parses, both jobs intact, both paths updated.

No API change, no behaviour change, no kernel change, no xcframework rebuild.


gsdali (commented):
## Overview

Fixes #440: a SwiftPM "found 1 file(s) which are unhandled" warning that fires during package
*planning*, so it hits every downstream consumer building OCCTSwift from source, not just this
repo's own test builds. The PR:

- Correctly re-diagnoses the actual cause as `Tests/OCCTStressTests/Fixtures/unify-crash-mmd-kiha10-body5.brep`
  (a `#filePath`-read fixture, never a `Bundle.module` resource) — not `Tests/occt_parallel_crash_portable.cpp`,
  which #440 originally blamed.
- Adds `exclude: ["Fixtures"]` to the `OCCTStressTests` target.
- Moves the `.cpp` (100% rename) from `Tests/` root to a new `Scripts/repro/occt1179-parallel-crash/`,
  with a README, and updates both CI workflow path references.
- Drive-by: strips 11 em-dashes from `Package.swift` comments per house style, explicitly
  flagged in the PR body so it isn't a surprise in review.

## Verification of the PR's own claims

- **Root-cause re-diagnosis**: verified independently — every `.testTarget` in `Package.swift`
  declares an explicit subdirectory `path:` (`Tests/OCCT*Tests`), none declares bare
  `path: "Tests"`, so SwiftPM genuinely never scans the `Tests/` root directly. The "moving the
  `.cpp` out and rebuilding reproduces the identical single-file warning" test in the PR body is
  exactly the right way to prove this, and the target-path check confirms the mechanism.
- **CI workflow update**: fetched the PR branch's full `occt-parallel-crash-test.yml` — exactly
  two references to the old path exist in the whole file, both updated to the new one, no stray
  leftovers.
- **`exclude: ["Fixtures"]` is the minimal correct fix** — a `resources:` declaration would have
  copied the 600 KB `.brep` into the test bundle for nothing, since the fixture is read via
  `#filePath` directly from the source tree, not `Bundle.module`.

## Finding: the ".cpp's placement was incidental" claim is factually wrong (repeated 3×)

The PR body, the new `README.md`, and the `CHANGELOG.md` entry all say:

> The `.cpp` itself was committed later, under `Tests/` at the package root (e32c8ba, via #379, an
> unrelated `Edge.circleProperties` fix), so its original location was incidental rather than chosen.

Checked against actual git history:

```
e174ae6  2026-04-14 12:35:37  "Add cross-platform OCCT parallel crash reproducer and CI workflow"
d239bd7  2026-04-14 12:35:50  "Add portable C++17 parallel crash reproducer source"   <- adds the .cpp
103ed61  2026-04-16           "Restructure parallel crash reproducer: isolated group testing (v2)"
e32c8ba  2026-07-26           "fix(#378): Edge.circleProperties..." (#379)             <- does NOT touch the .cpp
```

`e32c8ba`/#379 never touches `occt_parallel_crash_portable.cpp` at all (`git show e32c8ba --stat`
has no mention of it) — it's the unrelated `Edge.circleProperties` fix, three months later. The
`.cpp` was actually added by `d239bd7`, thirteen *seconds* after the CI workflow commit
(`e174ae6`) on the same day — i.e. added together with the workflow in the same initial batch,
which is about as deliberate a placement as it gets, not "incidental."

This doesn't undermine the PR's actual justification for moving the file — "a reproducer does not
belong in the package's test tree" stands on its own regardless of how it got there — but the
specific commit-attribution claim is wrong and repeated in three places that will outlive this PR
(README, CHANGELOG). Worth fixing before merge, particularly since this PR's own stated purpose is
correcting a *previous* misdiagnosis (#440's) through careful verification — the same rigor should
extend to this claim. Suggest either dropping the "incidental rather than chosen" framing or
correcting the commit references to `d239bd7`/`e174ae6` and removing the `e32c8ba`/#379 mention.

## Minor notes

- `Scripts/repro/occt1179-parallel-crash/` breaks from this repo's usual repro-directory
  convention of prefixing with *this project's own* issue number (`298-*`, `319-selfintersection`,
  `341-meshcaf`, `374-resource-manager-storage-schema-race`, etc.) — here it's prefixed with the
  upstream OCCT issue number instead. Given this reproducer apparently never got its own OCCTSwift
  issue number, that's a reasonable one-off call rather than an oversight, just flagging the
  deviation for awareness.
- Spot-checked several of the em-dash replacements in `Package.swift` — all preserve meaning; one
  or two (e.g. the `OwnAutoNamingScope` patch-list entry) read marginally less crisp with a bare
  comma than the original appositive dash, but nothing incorrect.

## Risk / Test Coverage

No API, behavior, or kernel change — purely build-manifest and file-location. The verification
list in the PR body (clean `swift build`, `--build-tests` across all 18 targets, `dump-package`,
targeted + full `swift test`, standalone recompile of the moved file, workflow YAML parse check)
is thorough for what is otherwise a low-risk chore. Low risk overall; the one finding above is a
documentation-accuracy fix, not a functional concern.

PR #515, opened 2026-07-30
fix(#486): one batch grid-evaluation family, not three generations per type

gsdali:
Closes #486.

Part of the #381 / #377 duplication audit (pass 1b). Targets `refactor/381-pass1b`, not `main`.

## The duplication

`Curve3D`, `Curve2D` and `Surface` each had **three** generations of "evaluate at N parameters"
bridge functions, 15 in total, with no shared helper anywhere between them: every one hand-rolled
its own parameter-pack loop and its own result-unpack loop. The two Surface entry points had drifted
onto opposite UV layouts as a direct result:

| | wrote | its header comment said |
|---|---|---|
| `OCCTSurfaceEvaluateGrid` (v0.29.0) | `outXYZ[(iv * uCount + iu) * 3]`, V-major | "row-major (u varies fastest)" |
| `OCCTGridEvalSurfaceD0` (v0.111.0) | `xs[iu * vCount + iv]`, U-major | "row-major" |

"Row-major" says nothing about a UV grid, where either parameter can be the row.

## Ground truth first

Compiled a standalone C++ test against the pinned headers before touching any code
(`GeomGridEval_Curve`, `Geom2dGridEval_Curve`, `GeomGridEval_Surface`):

- `EvaluateGrid(uArr, vArr)`'s `Array2` is `(iu, iv)`, 1-based, `NbRows() == uCount`. **Both existing
  implementations read it correctly**: the divergence was purely in the output buffer index.
- `GeomGridEval_Curve::EvaluateGrid` agrees with per-point `Geom_Curve::EvalD0`/`EvalD1` to ~4e-13 on
  a BSpline, and exactly on analytic curves. Same for the 2D sibling.
- Empty params return an empty array, no throw. Out-of-domain params silently extrapolate.

**One claim in the issue does not hold.** The two Swift surface spellings were *not* returning
transposed grids relative to each other: `Surface.evaluateGrid` transposed the V-major bridge buffer
while unpacking, so both ended up U-major at the Swift layer. The real hazards were the bridge-layer
divergence and the flat array's undocumented index formula, not a mismatch between the two Swift
results.

## What changed

**Bridge: 15 functions to 6.** Removed nine duplicates (`OCCTCurve3DEvalBatchD0`/`D1`,
`OCCTCurve2DEvalBatchD0`/`D1`, `OCCTGridEvalCurveD0`/`D1`, `OCCTGridEvalCurve2dD0`/`D1`,
`OCCTGridEvalSurfaceD0`) and gave the survivors one contract, documented above
`OCCTCurve3DEvaluateGrid`: one OCCT batch evaluator per job, interleaved output buffers, U-major
surface grids, and an `int32_t` count return.

Two shared helpers in `OCCTBridge_Internal.h` now carry what each function used to re-derive:
`occtGridEvalParams` for the pack loop, and `occtSurfaceGridIndex` as the single definition of the
surface grid layout. That is the part that stops this recurring: the index convention lives in one
place, in code, rather than being re-spelled in prose at each call site.

`OCCTGridEvalSurfaceD1` had no duplicate, so it survives as `OCCTSurfaceEvaluateGridD1`, moved beside
its D0 sibling and reshaped to the family's buffers. `OCCTCurve2DEvalBatchD0`/`D1` had been defined in
`OCCTBridge_Curve3D.mm` despite operating on `Curve2D`; their replacement lives in
`OCCTBridge_Geom2d.mm`.

**No Swift source break.** All 15 spellings still exist; nine are deprecated and forward to their
canonical sibling. Per [`SEMVER.md`](docs/SEMVER.md), a signature or return-type change here would
need another recorded exception, so there is none.

**New: `Surface.evaluateGridD1(uParameters:vParameters:)` + `SurfaceGridD1`.** Finishes for the D1
path what #404 did for D0: results are indexed `.at(u:v:)` instead of arriving as a flat array whose
major order the caller has to know. Both grid types index through one internal `surfaceGridIndex`,
mirroring the bridge helper.

## Behaviour changes

1. **A failed evaluation returns an empty result, not zeroes.** The `void` bridge functions gave
   their Swift callers no way to detect failure, so the wrappers returned a full-length array of
   `SIMD3(0, 0, 0)` as if evaluation had succeeded. All six survivors return the count written.
2. **`evalBatchD0`/`evalBatchD1` now reach the batch evaluator.** Those four had regressed to
   looping `Geom_Curve::EvalD0`/`EvalD1` (and the 2D equivalents) per point, bypassing the
   `GeomGridEval_Curve` path `evaluateGrid` had used since v0.29.0. Results can move by ~1e-13 on a
   BSpline; analytic curves are unchanged.
3. **C API only: `OCCTSurfaceEvaluateGrid` is now U-major.** `OCCTBridge` is not an SPM product, so
   this reaches no consumer through the package. Its Swift caller dropped the compensating
   transpose, so `SurfaceGrid` output is byte-for-byte unchanged. U-major was already the majority
   convention (`OCCTSurfaceDrawMesh`, `OCCTGridEvalSurfaceD0`/`D1`, and `SurfaceGrid` from #404).

## Tests

Three new suites, 11 tests: `Issue486SurfaceGridTests`, `Issue486Curve3DBatchTests`,
`Issue486Curve2DBatchTests`.

The existing `GridEvalSurfaceTests` used asymmetric grids but only asserted counts and rough
magnitudes, so it could never have caught the layout divergence. The new surface tests cross-check
every `(u, v)` against the independent `point(atU:v:)` / `evalD1(u:v:)` evaluators, which is the
check `evaluateGrid` has had since #404, and they pin the deprecated flat arrays against their newly
documented index formula.

Verified the new tests actually catch the defect class: with `occtSurfaceGridIndex` deliberately
inverted to V-major, 3 of the 5 surface tests fail with 83 issues. Restored before committing.

Full suite green.

Also swept the sibling ecosystem repos for callers of the nine deprecated spellings: the only
non-Package.swift hits are `OCCTSwiftViewport`'s MetalDemo gallery, and every one of those calls the
canonical `Curve3D.evaluateGrid(_:)`, which is unchanged. No downstream deprecation warnings.

## Note for whoever releases this

`Sources/OCCTBridge/src/*.mm` changed, so the prebuilt `OCCTBridge.xcframework` needs rebuilding and
its release-asset URL + checksum bumping in `Package.swift`. A stale prebuilt bridge is worse than
usual here: `OCCTSWIFT_BRIDGE_PREBUILT=1` against the old binary would give new Swift sources a
V-major buffer they no longer transpose, silently transposing every `evaluateGrid` result. Everything
in this PR was built and tested from source with that env var unset.


gsdali:
## Review

### Overview

Targets `refactor/381-pass1b` correctly. Consolidates three generations of "evaluate at N parameters" bridge functions across `Curve3D`/`Curve2D`/`Surface` (15 total) down to 6 canonical ones, after finding the two `Surface` grid-eval spellings had drifted onto **opposite UV layouts** while both header comments called themselves "row-major." Also fixes a correctness regression (`evalBatchD0`/`D1` had silently stopped using the batch evaluator, falling back to a per-point loop) and a silent-failure bug (`void`-returning bridge functions gave Swift no way to detect a failed evaluation, so callers got a zero-filled array indistinguishable from a real result).

I independently verified the two claims most load-bearing for correctness, rather than trusting the description:

- **The U-major/V-major transpose-removal math checks out.** I traced both the old bridge index formula + old Swift transpose and the new bridge index formula + no-transpose path by hand: both reduce to `SurfaceGrid[u*vCount+v] = point(u[u], v[v])`. The claim that `SurfaceGrid`'s Swift-level output is byte-for-byte unchanged despite the bridge layout flip is correct.
- **Deprecation-annotation coverage is complete.** Unlike a sibling PR in this same series (#509), I grepped the whole branch for every remaining call site of the four newly-deprecated method families (`gridEvalD0/D1`, `evalBatchD0/D1`) and confirmed **all of them** — including pre-existing tests this PR didn't otherwise touch — are wrapped in `@available(*, deprecated, ...)` at the enclosing `@Test func`. No leftover references to any of the 9 removed C bridge symbols anywhere in `Sources/`.
- CI green (macOS build+test, iOS simulator smoke).

### Code quality & style

- Strong adherence to project conventions: the two new shared helpers (`occtGridEvalParams`, `occtSurfaceGridIndex`) go exactly where CLAUDE.md's own architecture doc would put them (`OCCTBridge_Internal.h`), with a comment explaining *why* they exist (to stop the layout from re-diverging), not just what they do.
- The "extend this family; do not start a seventh spelling" comment block above `OCCTCurve3DEvaluateGrid` is a genuinely useful piece of institutional memory for future contributors — names the exact defect class and points at the contract to extend instead.
- Every deprecated method's message is specific and actionable (names the tuple-label change from `d1` to `tangent`, not just "use X instead").
- `SurfaceGridD1` mirrors `SurfaceGrid`'s existing design exactly (shared `surfaceGridIndex` helper, same `.at(u:v:)` shape, same `isEmpty`/`.empty` pattern) — good consistency rather than inventing a parallel convention.

### Specific findings

**1. A latent buffer-overflow risk appears to be fixed here, but isn't called out as such.** The new `std::min(paramCount, static_cast<int32_t>(results.Size()))` clamp in `OCCTCurve3DEvaluateGrid`/`OCCTCurve2DEvaluateGrid` (`D1` variants too) is described in-code only as guarding the *empty-result* case ("clamp rather than trusting paramCount"). But the old code (`int32_t n = static_cast<int32_t>(results.Size());` then looping `n` times into a buffer sized `3*paramCount`) would have been a real out-of-bounds write into the caller's buffer if `results.Size()` had ever exceeded `paramCount`. Worth confirming whether that direction was ever actually reachable (vs. purely defensive) — if it was, this is a more significant fix than the PR credits it as.

**2. Minor asymmetry, unexplained:** the curve family's fix clamps-and-partially-fills on a size mismatch, while the surface family's fix (`if (results.NbRows() < uCount || results.NbColumns() < vCount) return 0;`) rejects entirely. Both are defensible (a 2D partial-fill is harder to reason about safely), but the PR doesn't explain why the two families took different strategies — worth a one-line comment if intentional.

**3. Verified, not just trusted:** the "no Swift source break" and "zero remaining ecosystem callers of the deprecated spellings" claims both check out from what's visible in this repo. I did not independently verify the sibling-repo sweep (`OCCTSwiftViewport`), since that's outside this diff's scope.

### Potential issues or risks

- **None blocking.** The `~1e-13` precision change in `evalBatchD0/D1` (now reaching the real batch evaluator) is a genuine behavior change but is exactly the point of the fix, bounded, and measured against ground truth rather than asserted.
- **Operationally important, correctly flagged**: the "Note for whoever releases this" section is exactly right to call out — a stale prebuilt `OCCTBridge.xcframework` combined with new Swift sources would silently transpose every `evaluateGrid` result, which is the worst possible failure mode (wrong answer, no crash, no warning). Good that the PR explicitly confirms it was built and tested with `OCCTSWIFT_BRIDGE_PREBUILT` unset.
- No thread-safety implications (pure computation, no shared/global state).
- Test coverage is strong and targeted: the new `Issue486SurfaceGridTests` suite specifically cross-checks every `(u,v)` against an independent evaluator (`point(atU:v:)`/`evalD1`) — the exact check missing before — and the PR's own verification note (deliberately inverting `occtSurfaceGridIndex` to confirm 3/5 tests fail with 83 issues, then reverting) is a good example of proving the tests actually catch the defect class rather than just adding green checkmarks.

**Overall**: a thorough, well-verified consolidation with good deprecation hygiene and a mathematically-checked no-behavior-change claim on the tricky part (the transpose removal). The one thing worth a follow-up comment (not a blocker) is confirming whether the new curve-side buffer clamp was fixing a reachable overflow or a purely defensive one, since that changes how the fix should be framed.

Issue #726, opened 2026-08-06
Hunt for values that were never computed but are returned as measurements (hardcoded constants, pinned fixture counts, name-prefix censuses)


gsdali:
## The class

A value that was never computed, returned through an API whose shape says it was measured. The
caller cannot tell the difference, because a fabricated answer and a real one have the same type,
the same plausible magnitude, and no error channel between them.

This keeps recurring, has been found by accident every time, and nothing in the tree looks for it.
Opening this to make it something we hunt rather than something we stumble into.

## Confirmed instances

**Production code returning a constant that reads as a measurement:**

- `OCCTBridge_Healing.mm:248` — `result.selfIntersectionCount = 0;  // Would require more expensive
  computation`. Every `Shape.analyze()` result has reported zero self-intersections since the field
  existed. Not "none found": never looked. Documented as such only just now, in #702's PR.
- #605 — `Shape.centerOfMass` returned the **bounding-box centre**. A number of the right type and a
  believable magnitude, for a different quantity.
- #609 — zero-mass `BRepGProp` results returned as successful answers, with the fabricated magnitude
  an open shell produces.
- #522 — `MaxError()` reported 4.08e-15 unchanged from tolerance 1e-1 down to 1e-7, because the
  quantity it summed had been zeroed. A confidence number that was structurally incapable of moving.
- #725 (just filed) — `joinBezierPatches` reports `IsDone() == true` while having silently dropped
  the weights.
- #640 / PR #716 — `math_GaussSetIntegration` returns `IsDone() == true` and a wrong value for any
  input with more than one variable.

**Tests pinning a hardcoded number that encodes the bug rather than the answer:**

- #703 / PR #720 — the "real pocket" fixture asserted 1 pocket for eight months. Measured: it removed
  `0.0` volume. It never overlapped the box. The `1` was entirely the defect it was meant to guard.
- #640 — a test asserted the 2-variable integral equals `0.5`. The correct value is `1.0`. The pin
  had frozen a wrong answer as the expected one, and "fix the literal" would have frozen a second
  wrong answer, since the class cannot produce `1.0` at all.
- #702 / PR #717 — a total was about to be pinned as a constant when the fixture turned out to carry
  a pre-existing nonzero `gapCount`. Computing the expected value from the real fields is what
  stopped the test pinning a coincidence.

**Censuses reporting a count nobody measured:** #558 said 14, measured 28. #571 said 3, measured 6.
[...]

PR #751, opened 2026-08-07
investigate(#597): the two remaining bridge-side sites cannot be fixed by rejecting on error

gsdali:
## What & why

#597's own comment (verified by #741) narrowed the remaining scope to bridge-side sites in
`OCCTBridge_Modeling.mm`/`OCCTBridge_Healing.mm` that run an approximation and never read the error
it reports. Census (per `docs/v2.0.0-plan.md`'s census-once rule) found two candidates:
`occtPlateApproxSurface` (shared helper backing `OCCTShapePlatePoints`/`OCCTShapePlateCurves` in
`OCCTBridge_Healing.mm`, plus four more entry points in `OCCTBridge_ProjLib_NLPlate.mm`) and
`OCCTShapeFillBuildResult` (backing `OCCTShapeFill`/`OCCTShapeFillWithSupport`/
`OCCTShapeFillConstraints` in `OCCTBridge_Healing.mm`), plus several sites that turned out not to
need anything: `ShapeUpgrade_UnifySameDomain`'s two bridge sites confirm #741's own finding (the
OCCT class exposes no error API at all); `PipeShellBuilder`/`FillingSurface` already expose their
underlying class's error as an opt-in accessor (the deliberate manual-builder design, not an
oversight); `ShapeCustom_BSplineRestriction` self-polices in the kernel (declines a face's
conversion rather than ever accepting one out of tolerance).

Both real candidates were built and measured against the actual bridge, not just reasoned about.
**Both regress real, already-shipped, already-tested behaviour**, for two different reasons:

- `GeomPlate_MakeApprox::ApproxError()` measures fidelity to an *intermediate* object
  (`GeomPlate_Surface`) the caller never sees, not fidelity to the caller's own input points/curves,
  which is what the six entry points' actual contract (and their existing tests) check. On the
  #571 fixture it exceeds `tolerance` by up to 5.8x while the deviation from the caller's own input
  points stays inside it throughout. Gating on it broke all 6 `Issue571PlateApproxTests`.
- `BRepOffsetAPI_MakeFilling::G0Error()` *is* the right kind of number (distance to the caller's own
  constraints), but the bridge's effective `Tol3d` is always `1e-4`, `FillingParameters`'s own
  Swift default, not a fallback for an unset value, and legitimate, correct, higher-continuity or
  heavily-constrained fills routinely exceed it. Gating on it broke 2/17 `FillingSupportFaceTests`
  whose surfaces are provably correct by their own geometric assertions.

Both changes were reverted. What ships is the measurement: two doc comments recording exactly why
(so the same fix is not re-attempted and re-reverted), a durable `CLAUDE.md` Known OCCT Bugs entry
recording the same investigation and conclusion (see "Review response" below), and the full
investigation with five standalone reproducers under
[`Scripts/repro/597-bridge-modeling-healing-approx-error/`](https://github.com/SecondMouseAU/OCCTSwift/tree/refactor/381-pass1b/Scripts/repro/597-bridge-modeling-healing-approx-error).

**No functional change.** This does not close #597. It fully accounts for the bridge-side scope
the issue's own comment left open, and the conclusion is that none of it is safely fixable the way
#741 fixed `OCCTGeomFillSweep`. The remaining, deliberately out-of-scope kernel defect
(`GeomFill_Sweep.cxx:325`, where `SError` is overwritten with the requested tolerance instead of the
measured one inside the `myForceApproxC1` branch) is unchanged; it needs a carried patch and its own
reproducer, tracked on #597 and #726, not a bridge-side check.

Refs #597, #668, #726, #741

## Review response

Two reviews landed on the original commit: one automated pass (4 findings, all in the investigation's
supporting files) and a follow-up pass (a 5th, more consequential finding, added as an inline
comment). All five are addressed here.

1. **Hand-typed census table conflicts with the census-once rule.** Valid. Cluster E (#668) already
   has an executable census artifact at
   [`Scripts/repro/572-approx-consumer-sweep/`](https://github.com/SecondMouseAU/OCCTSwift/tree/refactor/381-pass1b/Scripts/repro/572-approx-consumer-sweep),
   which is #597's own origin. `Scripts/repro/597-bridge-modeling-healing-approx-error/README.md`'s
   `## Census` section now points there as the entry point instead of restating its own table, and
   the wider family #597 established (every approximation-error-reporting class in
   `OCCTBridge_Modeling.mm`/`OCCTBridge_Healing.mm`, not just `GeomConvert_ApproxSurface` consumers)
   has been added to the 572 artifact's own README (its "#597: the wider family" section) rather
   than duplicated here.
2. **Dead multi-span knot vector in `wavyEdge()`, plus a comment mislabeling a degree-13 curve as
   "cubic".** Moot: `wavyEdge()` was rewritten wholesale to fix finding 5 below, and the rewrite has
   no dead code and no stale comment. Nothing left to fix separately.
3. **The "gating on `ApproxError()`/`G0Error()` regresses shipped behaviour" conclusion lived only
   as inline comments.** Added a `CLAUDE.md` `## Known OCCT Bugs` entry (matching the length and
   detail of its neighbours) recording the investigation, both candidate fixes, why each was
   rejected, and the corrected finding-5 measurement below. The two inline comments
   (`OCCTBridge_Internal.h` on `occtPlateApproxSurface`, `OCCTBridge_Healing.mm` on
   `OCCTShapeFillBuildResult`) are unchanged; the new entry is the durable, grep-from-one-place
   record that survives the pending file breakdowns (#393-#395).
4. **Dead `nKnots` variable in `occt_597_bsplinerestriction_selfpolice.mm`, off by one from what is
   actually used.** Removed. Rebuilt and re-ran; output is byte-identical to before the removal.
5. **The fixture does not build what it claims, the important one.** `wavyEdge()` in
   `occt_597_fill_g0_realistic.mm` built its sine-sampled points as poles of a single-span,
   full-multiplicity `Geom_BSplineCurve`, i.e. a degree-13 Bezier, not an interpolant. Independently
   verified the review's own Bernstein-basis measurement before touching anything: for this
   function's exact parameters (`amp=4.0, waves=3, npts=14`), control-point peak 3.971, realized
   curve peak 1.4895 (rounds to 1.490), so 62.8% of the intended amplitude is gone, matching the
   review comment exactly. Fixed by rebuilding `wavyEdge()` on `GeomAPI_Interpolate`, which actually
   interpolates the sampled points, matching this repo's own naming precedent
   (`OCCTCurve3DInterpolate` in `OCCTBridge_Curve3D.mm` is `GeomAPI_Interpolate`-based;
   `GeomAPI_PointsToBSpline`, this repo's "FitPoints", approximates rather than interpolates and
   would have the same problem here).

   **Re-measured, and the numbers moved dramatically:**

   ```
   tol3d=0.0001     isDone=true  G0Error=0.20369      bboxDiag=1044.12    EXCEEDS TOLERANCE, ACCEPTED ANYWAY
   tol3d=1e-05      isDone=true  G0Error=0.203757     bboxDiag=1043.96    EXCEEDS TOLERANCE, ACCEPTED ANYWAY
   tol3d=1e-06      isDone=true  G0Error=0.226344     bboxDiag=1247.59    EXCEEDS TOLERANCE, ACCEPTED ANYWAY
   ```

   versus the old (damped) fixture's `G0Error=5.89516e-05, within tol` at the bridge's own default.
   The corrected boundary exceeds even the loosest tested tolerance, the bridge's own `1e-4`
   default, by roughly 2000x. Added a bounding-box and control-pole diagnostic to the script while
   investigating why the error looked so large: the accepted surface's poles land up to ~530 units
   from a boundary whose own extent is about 10 units across. `IsDone()` is still true, and
   `G0Error()` alone does not communicate how far the fit has actually diverged.

   **Does the PR's conclusion survive? Yes, on the part that was measured. A follow-on sentence
   that was not measured has been corrected, narrower, rather than left standing.** The load-bearing
   evidence for "gating on the caller's own requested tolerance regresses shipped behaviour" was
   always the confirmed regression against `FillingSupportFaceTests` (2/17 failures on correct,
   already-tested surfaces), independent of this fixture, and that conclusion is unchanged.

   A second sentence went further, claiming no single `G0Error()` threshold at all could separate a
   genuinely diverged fit from the two regressing tests, without ever measuring those two tests' own
   `G0Error()` values. A second coordinator review caught this as an unmeasured, load-bearing claim.
   Measured it: instrumented `OCCTShapeFillBuildResult`/`OCCTFillingBuild` with a temporary debug
   print of `G0Error()`, ran each of the 17 `FillingSupportFaceTests` individually so each value maps
   to one test unambiguously, reverted afterward (no production code changed). The two regressing
   values are 5.295e-4 (`curvatureContinuityIsAccepted`'s `.g2` branch, 5.3x the `1e-4` tolerance
   every test uses) and 1.231e-3 (`internalConstraintIsNotABoundary`'s interior-pull branch, 12.3x
   tolerance), both two to three orders of magnitude below the corrected fixture's ~2000x-2260x. A
   fixed absolute threshold placed between them (e.g. `0.01`) would in fact separate today's known
   cases. The honest conclusion is narrower than "no threshold can work": gating on the caller's
   *requested* tolerance is unsafe and proven so; a *fixed* threshold is unjustified rather than
   impossible, since nobody has measured what it should be, which is exactly the failure mode #726
   exists to catch. Full detail, including the per-test measurement table, is in
   [`Scripts/repro/597-bridge-modeling-healing-approx-error/README.md`](https://github.com/SecondMouseAU/OCCTSwift/tree/refactor/381-pass1b/Scripts/repro/597-bridge-modeling-healing-approx-error).

### None, investigation only, no functional change (#597)

Measured whether the two remaining bridge-side approximation sites in `OCCTBridge_Modeling.mm`/
`OCCTBridge_Healing.mm` could be fixed the way #741 fixed `OCCTGeomFillSweep` (reject a result whose
reported error exceeds the requested tolerance). Both regressed real, already-shipped, already-
tested behaviour when actually built and tested, for two different reasons (see the PR description
and `Scripts/repro/597-bridge-modeling-healing-approx-error/README.md`), so neither change shipped.
Two comments were added recording the measurement, plus a `CLAUDE.md` Known OCCT Bugs entry making
the same record durable across the pending file breakdowns. A PR review found that one of the
investigation's own supporting fixtures understated its boundary's difficulty; corrected and
re-measured, the underlying conclusion holds and is now backed by a harder, more convincing case.

## SemVer impact

NONE. No production code behaviour changed; only comments, a `CLAUDE.md` entry, and
`Scripts/repro/` content were added or corrected.

## Checklist

- [x] New or changed behavior is covered by a unit test in the same PR: N/A, no behavior changed.
      The investigation's evidence is the *existing* test suites (`Issue571PlateApproxTests`,
      `FillingSupportFaceTests`) failing against the candidate fix and passing again once reverted;
      see "Notes for the reviewer".
- [x] Every new test and every new `--self-test` case was run once with its subject broken: N/A,
      no new test was added (see above); the policy's spirit is satisfied by running the *existing*
      suites both with the candidate fix present (fails, 6/6 and 2/17) and reverted (passes, full
      suite green).

## Notes for the reviewer

**Why this PR has no `Closes` line but also isn't purely a no-op.** #597, as re-scoped by its own
comment, asked specifically for the bridge-side sites in these two files. That scope is now fully
investigated: the census is complete, and every candidate is accounted for (fixed by #741, not
reachable, already correct by design, self-policing in the kernel, or unsafe to fix the obvious way,
measured). There is no remaining bridge-side work in `OCCTBridge_Modeling.mm`/`OCCTBridge_Healing.mm`
to do for this issue. Recommend #597 stays open only for the kernel-side `GeomFill_Sweep.cxx:325`
fix already noted in its own comment thread, which this PR does not touch.

**Both regressions were confirmed against the real bridge, not inferred.** For each candidate: (1)
implement the reject-on-error-exceeds-tolerance check, (2) `swift build` from source
(`env -u OCCTSWIFT_LOCAL`), (3) run the relevant existing suite and observe it fail, (4) revert, (5)
re-run and observe it pass again. This is the same "prove the test fails" discipline applied in
reverse: proving a *proposed fix* fails before ruling it out, not just a new test.

All 5 gate scripts and their `--self-tests` green: `check-bridge-index`, `check-null-handle-guards`,
`check-docs-defaults`, `derive-bridge-header-split --verify`, `count-operations`; the census and
merge-history audit self-tests also pass. Full suite (`swift test`): all tests pass, no failures
introduced by this review response (no production code changed).


gsdali:
The fixture rebuild is right and the re-measurement is the valuable part. One thing sent back.

## The corrected fixture falsifies the original measurement, and that is fine

`GeomAPI_Interpolate` in place of the single-span full-multiplicity Bezier, and the boundary that
previously read `G0Error=5.9e-05` at `tol3d=1e-4` now reads **0.204**, exceeding the bridge's own
default by roughly 2000x. The added bounding-box diagnostic is what makes it interpretable: the
accepted surface's poles sit around 530 units from a 10-unit boundary, so this is a diverged fit
rather than a demanding one.

That is the opposite of what the README used to claim, and saying so plainly rather than softening
it is the right call.

## What I sent back

The README now argues:

> there is no single threshold on `G0Error()`'s raw value that rejects this case while keeping those
> two passing

**That claim is unmeasured.** `curvatureContinuityIsAccepted` and `internalConstraintIsNotABoundary`
are named, but their own `G0Error` values appear nowhere on the branch. The nearby figures,
6.22713e-05 and 9.03801e-07, belong to `occt_597_fill_g0_mechanism.mm`, a different fixture, and
cannot stand in for them.

It is also load-bearing. It is the sentence that turns "gating on the caller's tolerance is unsafe",
which is proven by the 2-of-17 regression, into "no threshold can work at all", which is not. If the
two regressing tests sit four orders of magnitude below 0.204, as the mechanism fixture's numbers
hint they might, then a fixed absolute threshold **would** separate them and the honest conclusion
is narrower: unjustified rather than impossible.

Narrower is still sufficient. "The obvious gate is proven unsafe, and any other threshold would be a
number nobody measured" is a complete argument for not shipping one, and it is #726's own reasoning
turned on ourselves. The overstatement is what I want gone, not the conclusion.

The measurement is cheap: the gating patch that produced the 2-of-17 regression was already written
and run, so printing `G0Error()` alongside the effective tolerance per test is a small edit to
existing work.

## Noting the pattern

This is the third time today an unmeasured claim has turned up inside work whose subject is
unmeasured claims: the census script citing `hasExtent` as its model of correct code, my own stray
-modification parser missing `0020` because it keyed on one diff header spelling, and now this. The
common shape is not carelessness, it is that **a number adjacent to the one you need reads as the
one you need**.

Everything else in the response is good. Finding 2 is correctly moot, the census now points at
Cluster E's own artifact rather than re-deriving, and the CLAUDE.md entry is ordered against the
kernel half's entry that landed mid-task.

Issue #763, opened 2026-08-07
Unmeasured values, production half: adjudicate the census's 62 bridge candidates


gsdali:
Part of the unmeasured-values phase (#726). **Gates Pass 2a (#382) and Pass 3 (#384).**

`python3 Scripts/census-unmeasured-values.py` reports **62 production candidates**. Each is a field
assigned only a literal while a sibling field of the same aggregate, at the same conditional depth,
is assigned from a real computation. Every one needs a human verdict; the script is a census, not a
verdict, and it says so.

Where they are, which is why this gates those two passes specifically:

| File | Candidates | Audited by |
|---|---|---|
| `OCCTBridge_Document.mm` | 25 | Pass 3 (#384) |
| `OCCTBridge_Topology.mm` | 9 | Pass 2a (#382) |
| `OCCTBridge_Healing.mm` | 9 | Pass 2a |
| `OCCTBridge_Properties.mm` | 7 | Pass 2a |
| `OCCTBridge_BRepGraph.mm` | 4 | Pass 2a |
| six others | 8 | various |

**54 of the 62 sit in files Pass 2a and Pass 3 own.** That is the whole argument for doing this
first rather than folding it in.

## The verdict, per candidate

One of four, and record which:

1. **Not an instance.** Several will be the *opposite*: a value already using the
   "make absence representable" fix #583/#595/#609 established, like `OCCTShapeAxis.hasExtent`
   gating `extentMin`/`extentMax` into a Swift-side `nil`. Say so and move on.
2. **Compute it.** Non-breaking. Can land any time.
3. **Make the absence representable** (`nil`, or a sentinel that cannot collide with a real answer).
   Usually breaking.
4. **Remove the field.** Breaking.

## The v2.0.0 line

**Verdicts 3 and 4 are release-gating; 1 and 2 are not.** v2.0.0 is the last chance to change a
public field's type or remove it before v3.0.0, and a wrong value that a consumer cannot distinguish
from a right one is worth a break. Computing a value properly changes no signature and can ship in
any patch.

So the release-gating task is **not** to fix all 62. It is to triage all 62 to the depth needed to
answer "does the right fix here require a breaking change?", then make only those breaks. The rest
defers with the phase.

`selfIntersectionCount` (`OCCTBridge_Healing.mm:248`) is the seed instance and looks like verdict 4:
`Shape.isSelfIntersecting(timeout:)` already answers the question properly, so the field has no
[...]

Issue #382, opened 2026-07-26
Pass 2a — Shape/Topology core duplication audit


gsdali:
Part of the segmented audit plan in #377. Branch: `refactor/382-pass2a`, cut fresh from `main` on 2026-08-11 — not the shared `refactor/377-segmented-audit`, which turned out to be dead (frozen 2026-07-29, 38 commits behind `main`, predates Pass 1b/v2.0.0/the Shape.swift split). See #377's branch-policy note.

**Scope note:** behavior/bug fixes found alongside duplication are in scope here, not just structural dedup — see #377's "Scope: behavior fixes are in scope" section for why and how.

## The premise this issue used to carry is now stale in a way that isn't just numbers drifting

The original scope below assumed this pass would run against `Shape.swift` as a 13,661-line monolith,
and that its own findings would inform the mechanical breakout later (see the old "Rule": *fix or
triage findings here before ... #394 (Shape.swift breakout) start*).

That ordering already happened in reverse. `Shape.swift`'s breakout (tracked as #394, executed as
its two real sub-issues **#659** eviction and **#660** domain-split) merged **2026-08-03/04**, as
part of the v2.0.0 file-breakdown work, before this pass ever started. It was a pure verbatim move
(#659/#660's own acceptance criteria forbid any behavior or logic change) — it decided *where* code
lives, not whether any of it duplicates anything else. So this pass now inherits an already-split
file set rather than one monolith.

## Re-derived file list (measured today, not from the original issue body)

`Shape.swift`'s eleven `extension Shape` descendants plus its evicted standalone-type siblings
(excluding `ShapeAxis.swift`/`ShapeMeasurements.swift`, which are #383's declared scope, unrelated
to this split), plus the original `Edge`/`Face`/`Wire`/`WireOrder`/`TopologyRef`:

```
2378  Shape.swift
2140  Shape+Analysis.swift
 324  Shape+Curve.swift
 101  Shape+Drawing.swift
 762  Shape+Geom2d.swift
 159  Shape+Math.swift
 217  Shape+Mesh.swift
3218  Shape+Modeling.swift
1474  Shape+ShapeHealing.swift
1148  Shape+Surface.swift
2858  Shape+Topology.swift
  98  ShapeAnalysisResult.swift
  20  ShapeContents.swift
  38  ShapeContentsExtended.swift
  68  ShapeDistance.swift
  49  ShapeFixer.swift
 170  ShapeFixers.swift
[...]

PR #768, opened 2026-08-07
chore(#763): triage Document.mm + IO.mm's 26 unmeasured-value candidates

gsdali:
## What & why

Triages the `OCCTBridge_Document.mm` (25) + `OCCTBridge_IO.mm` (1) half of #763's 62
unmeasured-value candidates (parent: #726) — 26 of 62. Each candidate got a verdict (1: not an
instance / 2: compute it / 3: make absence representable / 4: remove the field) established by
reading the function and the OCCT call it wraps, not by the census line or the identifier alone,
per `okf/policies/measure-dont-assume.md`.

**Result: 25 Verdict 1 (not an instance), 1 Verdict 2 (compute it), 0 Verdict 3, 0 Verdict 4.**
No breaking changes in this PR.

Full per-candidate table with evidence: [`Scripts/repro/726-unmeasured-values/triage-document.md`](https://github.com/SecondMouseAU/OCCTSwift/blob/chore/763-triage-document/Scripts/repro/726-unmeasured-values/triage-document.md).

### The one fix (Verdict 2, non-breaking)

`OCCTDocumentGetShapeColor` hardcoded `result.a = 1.0`, reading through
`XCAFDoc_ColorTool::GetColor(shape, type, Quantity_Color&)` — the RGB-only overload. Reading
`XCAFDoc_ColorTool.cxx` directly shows that overload *internally* fetches the full RGBA value and
then discards alpha (`col = aCol.GetRGB();`); the underlying `XCAFDoc_Color` attribute always
stores real RGBA. A real, non-1.0 alpha reaches this exact storage today via STEP import
(`STEPCAFControl_Reader.cxx:2341` sets shape colors with `Quantity_ColorRGBA` carrying the file's
transparency). `OCCTDocumentGetLabelColor`, the label-keyed sibling three functions up in the same
file, already reads the RGBA overload correctly — this was the one inconsistent site.

**This overrides this repro directory's own prior README verdict for that exact site** (`ALPHA`,
"`Quantity_Color` has no alpha channel; `1.0` is the domain default"). That verdict conflated "the
C++ value type this call reads into has no alpha field" with "no alpha data exists to read" — not
the same claim, and the second is false here. Recorded as its own section in the triage table: this
is the #605-shaped case #763 warns about, a plausible-looking constant that was actually a wrong
choice of accessor, caught only by reading the OCCT source underneath the call.

**Companion change, not itself a census candidate:** `OCCTDocumentSetShapeColor` (the writer) only
ever stored through the RGB-only `SetColor` overload, and `Document.setShapeColor` was silently
dropping `color.alpha` before it reached the bridge at all, despite `Color` already carrying an
`alpha` field. Left alone, the getter fix would have been untestable from Swift except via a file
import. Added an additive `OCCTDocumentSetShapeColorRGBA` bridge function and pointed
`Document.setShapeColor` at it, forwarding `color.alpha`. No existing signature changed.

### Why 25 of 26 were "not an instance"

Most of this file's candidates are exactly the false-positive shapes the census's own header
comment documents: 6 are default-then-flip validity/presence flags (a literal `false`/`true` at the
top of a function, flipped only after a real OCCT call succeeds — the "make absence representable"
idiom #583/#595/#609 already established, not an instance of the bug those fixed), 1 is a genuine
structural constant (`XCAFDoc_VisMaterialPBR::EmissiveFactor` is `NCollection_Vec3<float>` per the
OCCT header — no alpha field exists anywhere upstream for emissive color), 18 are three plain
value-type constructors (`OCCTXCAFPrsStyleCreate*`) that build a request from Swift-supplied
parameters and are never used to read a style back out of a document (traced every call site: the
only caller is `PresentationStyle.toOCCT()`), and 1 is the `sewingApplied`-shaped "flag flipped true
because a real count was positive" idiom. Each is verified against the OCCT header/source or a
call-site trace, recorded in the table — not asserted from the field name.

## Tests

`Tests/OCCTXCAFTests/OCCTXCAFTests.swift`, suite `XDE ColorTool by Shape`:
- `shapeColorPreservesAlpha` — sets alpha 0.5 on a shape color, asserts it reads back as 0.5.
- `shapeColorOpaqueUnaffected` — regression guard: a fully-opaque color still round-trips to 1.0.

### Removal matrix (prove the test fails)

| Step | `shapeColorPreservesAlpha` | `shapeColorOpaqueUnaffected` | pre-existing `setAndGetColor` |
|---|---|---|---|
| Defect injected (reverted `OCCTDocumentGetShapeColor` to `Quantity_Color`/hardcoded `1.0`; kept the RGBA setter) | **FAILED** — `abs(got.alpha - 0.5) → 0.5` not `< 1e-5` | passed (bug invisible at alpha=1.0) | passed (never asserted on alpha) |
| Fix restored | passed | passed | passed |

## Gates and full suite

- All 5 gate scripts: clean (`check-bridge-index`, `check-null-handle-guards`,
  `check-docs-defaults`, `derive-bridge-header-split --verify`, `count-operations`).
- All 5 gates' `--self-test`, plus `census-unmeasured-values.py --self-test`: all green
  (18/18, 24/24, 13/13, 8/8, 10/10 respectively).
- `census-unmeasured-values.py` re-run after the fix: `OCCTDocumentGetShapeColor`'s `result.a` line
  is gone from the `OCCTBridge_Document.mm` candidate list; the new `OCCTDocumentSetShapeColorRGBA`
  introduced no new candidates (void return, no aggregate-result shape for the census to flag).
- Full `swift test`: **5483 tests, 1428 suites, 0 failures.**

### `Document.shapeColor`/`setShapeColor` now round-trip alpha correctly (#763)

`Document.shapeColor(_:type:)` previously always reported `alpha == 1.0` regardless of what was
actually stored, because the bridge read a shape's color through `XCAFDoc_ColorTool`'s RGB-only
`GetColor` overload. It now reads the RGBA overload and reports the real value — for example, a
STEP import with a transparent surface style. `Document.setShapeColor(_:color:type:)` previously
silently dropped `color.alpha` on write; it is now stored and round-trips through `shapeColor`.

## SemVer impact

PATCH. No public signature changed. `Document.setShapeColor`/`shapeColor` keep their exact
signatures; a new additive bridge function (`OCCTDocumentSetShapeColorRGBA`) was added to the C
surface, which is not part of the Swift-facing API surface `count-operations.py` tracks. Consumers
see a corrected value from `shapeColor` (previously always `1.0`) and a previously-dropped `alpha`
on `setShapeColor` now actually taking effect — a bug fix, not a behavior a correct caller could
have been relying on to be otherwise.

## Checklist

- [x] New or changed behavior is covered by a unit test in the same PR.
- [x] Every new test was run once with its subject broken, and the failure is reported above
      (removal matrix).

## Notes for the reviewer

- This is the `OCCTBridge_Document.mm` + `OCCTBridge_IO.mm` half of #763 (26 of 62 candidates).
  The other 36 (Topology/Healing/Properties/BRepGraph + six others) are separate work per #763's
  own table (Pass 2a / #382).
- `docs/reference/Document-Persistence-IO.md`'s `setShapeColor`/`shapeColor` entries were updated
  to describe the corrected alpha handling, per `docs/` staying current with behavior changes.
- Flagging explicitly for review: the override of this repro directory's own prior `README.md`
  verdict on `OCCTDocumentGetShapeColor`'s `result.a` (was `ALPHA`, now Verdict 2/fixed). I did not
  edit that README in this PR — it documents the *first census run's* first-pass analysis, and
  correcting it retroactively felt like it belonged in a PR about the README itself rather than one
  fixing the bug it missed. Happy to add a note there too if preferred.
- No verdict-3/4 (breaking) candidates in this half — nothing to flag against the v2.0.0 line here.

### DIFF
--- a/Scripts/repro/726-unmeasured-values/triage-document.md
+++ b/Scripts/repro/726-unmeasured-values/triage-document.md
@@ -1,0 +1,117 @@
+# #763 triage: `OCCTBridge_Document.mm` + `OCCTBridge_IO.mm` (26 of 62 candidates)
+
+Parent: #726. Sibling issue: #763 ("Unmeasured values, production half: adjudicate the census's 62
+bridge candidates"), the `OCCTBridge_Document.mm` (25) + `OCCTBridge_IO.mm` (1) half specifically.
+Candidate list reproduced with:
+
+```
+python3 Scripts/census-unmeasured-values.py | grep -E "^  OCCTBridge_(Document|IO)\.mm:"
+```
+
+Line numbers below are as reported **before** this pass's fix (`OCCTDocumentGetShapeColor` gained
+5 lines, shifting everything after it in the file by that amount; re-running the census today
+reports `OCCTDocumentGetShapeColor`'s own `result.a` line gone and every later line number +5).
+
+## Verdicts, per #763
+
+1. **Not an instance.** The literal is correct.
+2. **Compute it.** Non-breaking.
+3. **Make the absence representable** (`nil`/sentinel). Usually breaking.
+4. **Remove the field.** Breaking.
+
+Finer tags (reused from `Scripts/repro/726-unmeasured-values/README.md`'s own taxonomy, all of
+which collapse to Verdict 1 except where noted): **FLIP** = legitimate default-then-flip
+success/validity flag. **ALPHA** = structural default for a color quantity with no alpha channel
+anywhere in the data it is drawn from. **CTOR** = plain value-type constructor echoing its own
+parameters, not a measurement API returned from OCCT.
+
+## Result: 25 Verdict 1, 1 Verdict 2, 0 Verdict 3, 0 Verdict 4
+
+| # | Site | Function | Field | Verdict | Evidence |
+|---|---|---|---|---|---|
+| 1 | `OCCTBridge_Document.mm:427` | `OCCTDocumentGetLabelColor` | `result.isSet = true` | 1 (FLIP) | Read the full function. Sits immediately after `result.r/g/b/a` are assigned from a real `Quantity_ColorRGBA` returned by `doc->colorTool->GetColor(label, xcafType, color)`. `isSet` is `true` only on the path where a color was actually found (two call sites in the function, both after a real fetch) — the "make absence representable" idiom, not a fabrication. |
+| 2 | `OCCTBridge_Document.mm:520` | `OCCTDocumentGetLabelMaterial` | `baseColor.isSet = true` | 1 (FLIP) | Same idiom, one level in: set immediately after `baseColor.r/g/b/a` are populated from `pbr.BaseColor` (a real `Quantity_ColorRGBA`), inside `if (visMat->HasPbrMaterial())`. |
+| 3 | `OCCTBridge_Document.mm:530` | `OCCTDocumentGetLabelMaterial` | `emissive.a = 1.0` | 1 (ALPHA) | Verified against the OCCT header, not assumed: `XCAFDoc_VisMaterialPBR::EmissiveFactor` is `NCollection_Vec3<float>` (`Libraries/OCCT.xcframework/macos-arm64/Headers/XCAFDoc_VisMaterialPBR.hxx:40`) — RGB only, matching the glTF spec's `emissiveFactor` (no alpha channel exists anywhere upstream for this quantity). `1.0` (fully opaque) is the only sensible constant for an `OCCTColor`-shaped slot with no real alpha to read. Genuinely different from candidate #7 below, which looked structurally identical but was not. |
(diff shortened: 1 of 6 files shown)


gsdali:
Finding accepted and fixed. Verified it first: `OCCTDocumentSetShapeColor` has one definition, one
declaration, one cross-reference index line, and **zero callers** in either the Swift sources or the
bridge. `Document.setShapeColor` uses the RGBA entry point only.

Delegated rather than deleted. Dead is not the risk here; **reachable and wrong** is. The symbol
stays exported, sitting directly above a near-identical function whose only difference is the alpha
argument, and the next contributor to reach for the shorter of two adjacent overloads silently
reintroduces the exact alpha loss this PR exists to fix. One body means nothing left to drift.

**No test, deliberately.** No test target imports `OCCTBridge` and `OCCTSwift` does not re-export
it, so covering this symbol means adding an import for the sole purpose of exercising a function
nothing calls. The delegation removes the failure mode structurally rather than detecting it, which
is the better trade at this size. Flagging it so the absence reads as a decision rather than an
oversight.

Issue #765, opened 2026-08-07
Spike: can the narrow unmeasured-value form become a gate rather than a census?


gsdali:
Part of the unmeasured-values phase (#726). **Depends on the production half**, because that
adjudication is what tells you which shape is worth detecting.

`Scripts/census-unmeasured-values.py` is deliberately not a gate: it exits 0 whether or not it finds
anything, because its output is a list for a human to adjudicate rather than a verdict on the tree.
CI runs only its `--self-test`, since a bare run could never fail and so could never signal.

#726 scoped the gate question as a follow-up spike, and the reason is sound: "this literal was never
derived" is not decidable in general. But a narrow form may be: **a struct field assigned only a
literal across every path in its own constructor.**

## What the spike has to answer

- What is the false-positive rate of the narrow form against the 62 adjudicated production
  candidates? A gate is only worth having if a clean run means something, and the census's own
  experience is that a large fraction of flags are correct code.
- Does it need an `ALLOWED` table? If so, note the lesson from `check-null-handle-guards.py`: its
  entries are keyed `(file, function)` with no argument index and no re-validation, so an exemption
  silently covers every argument of that function and every later change to it. Do not repeat that
  shape without deciding it is acceptable.
- Would it have caught the known instances? #605, #609, #522, #640, #725, and the
  `selfIntersectionCount` seed. A gate that catches none of the six defects that motivated it is not
  a gate worth adding.

## If the answer is no

Say so and close it. A census that a human reads is a legitimate end state, and pretending otherwise
by shipping a gate with a large exemption table is worse than not having one: an `ALLOWED` entry
nobody revalidates is indistinguishable from a detector that has gone blind, which is the failure
`okf/policies/prove-the-test-fails.md` exists to prevent.

gsdali:
## Spike result: no gate. Census stays a census.

Built and ran the narrow-form detector against the real tree (`Scripts/repro/765-narrow-form-gate/detect_narrow_unmeasured.py`, spike-quality, not wired into CI, not shipping). Full writeup and every measurement below is in `Scripts/repro/765-narrow-form-gate/README.md` (kept locally in this session's worktree, not committed — see the closing note on why no PR).

### The narrow form, precisely

[...]

PR #770, opened 2026-08-07
fix(#763): triage core bridge files' unmeasured-value candidates

### LINKED ISSUE #763: Unmeasured values, production half: adjudicate the census's 62 bridge candidates

gsdali:
## What & why

Works the non-Document/IO half of #726's unmeasured-values sweep: the 36 census candidates
(of 62 total) in `OCCTBridge_Topology.mm`, `OCCTBridge_Healing.mm`, `OCCTBridge_Properties.mm`,
`OCCTBridge_BRepGraph.mm`, `OCCTBridge_Surface.mm`, `OCCTBridge_Modeling.mm`,
`OCCTBridge_Spatial.mm`, `OCCTBridge_Geom2d.mm`, `OCCTBridge_Curve3D.mm`. The sibling 26
(`OCCTBridge_Document.mm`/`OCCTBridge_IO.mm`) are triaged separately on
`chore/763-triage-document`.

Full per-candidate table with evidence: `Scripts/repro/726-unmeasured-values/triage-core.md`.

Closes #763 (its non-Document/IO half).

**Verdict counts** (per census line, 36 total):

| Verdict | Lines |
|---|---|
| 1. Not an instance | 29 |
| 2. Compute it | 6 |
| 3. Make absence representable | 0 |
| 4. Remove the field | 1 |

29 lines are legitimate default-then-flip success flags (`isValid`/`success`, driven by real
control flow), branch-selected classification tags (`result.type`, `a.kind`), or local
config/options structs passed as arguments and never returned (`BRepGraph::ShapesView::Options`,
`MathInteg::KronrodConfig`) — outside #726's "returned through an API" scope entirely. One
(`OCCTBisectorPointOnBisCreate`'s `isInfinite`) is a plain constructor verified against
`Bisector_PointOnBis.hxx`'s own real 5-arg constructor, which has no `IsInfinite` parameter
either. Two candidates needed action.

### 1. `ShapeAnalysisResult.selfIntersectionCount` removed (verdict 4)

`OCCTBridge_Healing.mm:303`, `OCCTShapeAnalyze()`: `result.selfIntersectionCount = 0;  // Would
require more expensive computation`. This is #726's own named seed instance.

Verified before removing, not assumed: `Shape.isSelfIntersecting(timeout:)` /
`isSelfIntersecting(hardTimeout:)` (`Shape.swift:2107`/`2150`) already answer the question this
field claimed to, via `OCCTShapeSelfIntersectsBounded` → `BOPAlgo_ArgumentAnalyzer`'s real
self-interference check (the #319 kernel work). Grepped every reader: two Swift call sites and two
test files, none of which asserted a specific self-intersection count (there was never a real one
to assert) — both test files only re-derive `totalProblems` from the remaining fields and are
updated to drop the term (numerically a no-op, since it was always 0).

**Migration**: delete any read of `ShapeAnalysisResult.selfIntersectionCount`. Use
`shape.isSelfIntersecting(timeout:)` for a real check.

### 2. `ShapeAxis.extent` now sometimes non-nil for `revolutionAxes()`/`symmetryAxes()` (verdict 2, non-breaking in type, behavior-changing)

`OCCTBridge_Topology.mm:653`/`:694`, `OCCTShapeRevolutionAxes`/`OCCTShapeSymmetryAxes`:
`a.extentMin = 0; a.extentMax = 0; a.hasExtent = false;` on every reached path, unconditionally,
in both functions.

An earlier census-only pass (`Scripts/repro/726-unmeasured-values/README.md`, committed earlier on
this same base branch) labelled this "GOOD — already fixed", reasoning that `hasExtent` gating
`extent` into a Swift-side `nil` is the same fix `#583`/`#595`/`#609` applied elsewhere. Re-verified
rather than trusted (`okf/policies/measure-dont-assume.md`): `hasExtent` is `false` at all three
assignment sites in the tree and **`true` at none** — it never actually flips, unlike a real
success flag, so `extent` was `nil` unconditionally for every caller. Same defect class as
`selfIntersectionCount`, one layer removed. The struct's own header comment already documented the
intended contract (`extentMin`/`extentMax`, `+-DBL_MAX` sentinels for the unbounded case), so this
was an incomplete feature (v0.137, never mentioned as delivered in the CHANGELOG), not a
deliberately deferred one.

Computed: `occtComputeAxisExtent` (`OCCTBridge_Topology.mm`) projects the axis's own shape's
`BRepBndLib`/`Bnd_Box` geometric bounding box onto the axis direction from its origin — exact
where the box is tight along that direction (a bounded cylindrical/conical face along its own
axis, a box along a principal axis; both are what the new tests check), a safe enclosing interval
otherwise. `IsVoid()` → `hasExtent = false`; `IsOpen()` → `hasExtent = true` with `+-DBL_MAX`,
honoring the header's pre-existing sentinel contract rather than leaking `Bnd_Box`'s internal
`Precision::Infinite()` (1e100) as if it were a real bound.

**No signature or type change** — `ShapeAxis.extent: ClosedRange<Double>?` is unchanged — but this
is a real behavior change: code that (incorrectly) assumed `extent` is always `nil` for
`revolutionAxes()`/`symmetryAxes()` results will now sometimes see a real value. Flagging under
its own heading per the instructions even though it doesn't break a build, since it changes
observable output.

**Migration**: none required (additive information). If any caller special-cased "extent is always
nil" for these two entry points, that assumption no longer holds.

## Tests

- `Tests/OCCTShapeHealingTests/OCCTShapeHealingTests.swift` and
  `Issue702SolidDemotionTests.swift`: updated to drop `selfIntersectionCount` from their own
  independently-recomputed problem-count sums (both still pass; the term was always 0 so this is
  not a strong regression check on its own, but does prove the removal compiles and the arithmetic
  is unaffected).
- New: `Tests/OCCTSurfaceTests/OCCTSurfaceTests.swift`'s
  `cylinderRevolutionAxisHasExtent` — a radius-5/height-20 cylinder's revolution axis reports a
  non-nil `extent` whose span is 20 (matches the cylinder's own height; asserts span rather than
  absolute bounds because the axis direction's sign isn't fixed, same as the pre-existing
  `cylinderOneAxis` test already accepts).
- New: `Tests/OCCTTopologyTests/OCCTTopologyTests.swift`'s `cylinderSymmetryAxisHasExtent` — the
  same cylinder's symmetry axis (origin = centre of mass, at half height) reports `extent` of
  exactly `-10...10`.

**Prove the test fails** (`okf/policies/prove-the-test-fails.md`): both new tests were run once
with the fix reverted (the three `occtComputeAxisExtent(...)` call sites replaced back with the
original hardcoded literals) — both failed with the expected message. Restored (verified
byte-identical via `diff` against a pre-injection backup) and re-ran — both passed.

| State | `cylinderRevolutionAxisHasExtent` | `cylinderSymmetryAxisHasExtent` |
|---|---|---|
| Defect injected | FAILED — "expected a non-nil extent on the cylinder's revolution axis" | FAILED — "expected a non-nil extent on the cylinder's symmetry axis" |
| Fix restored | PASSED | PASSED |

## Gate scripts (all clean)

All 5 static gates + their `--self-test`s, plus the census's own `--self-test`:

```
check-bridge-index.py                       0 stale, 0 misfiled                       exit 0
check-bridge-index.py --self-test           18/18 cases correct                       exit 0
check-null-handle-guards.py                 all functions guard                       exit 0
check-null-handle-guards.py --self-test     24/24 cases correct                       exit 0
check-docs-defaults.py                      0 drifted, 0 unverified                   exit 0
check-docs-defaults.py --self-test          13/13 passed                              exit 0
derive-bridge-header-split.py --verify      0 ambiguous, 0 unmapped, 0 misfiled        exit 0
derive-bridge-header-split.py --self-test   8/8 cases correct                         exit 0
count-operations.py                         4306 == 4306 (README/API_REFERENCE match)  exit 0
census-unmeasured-values.py --self-test     10/10 cases correct                       exit 0
```

`census-unmeasured-values.py` (report mode) on this branch: **55** production candidates, down
from **62** before this PR. The 7-line drop is exactly `selfIntersectionCount` (1) plus
`extentMin`/`extentMax`/`hasExtent` at both call sites (6), matching this PR's fixes.
`OCCTShapeSymmetryAxes.kind` (the TAG candidate) is still reported, at a shifted line number —
expected, since `kind` was untouched and remains verdict 1.

`check-changelog-transcription.py`'s report (not a gate) is unaffected by this diff — it audits
the branch's past merge history, not this PR's content.

## Full `swift test`

**5483 tests, 1428 suites, 0 failures**, ~78s wall clock. Includes the two new extent tests and
the two edited healing tests, all passing.

### `ShapeAnalysisResult.selfIntersectionCount` removed; `ShapeAxis.extent` now computed for `revolutionAxes()`/`symmetryAxes()` (#763)

`ShapeAnalysisResult.selfIntersectionCount` is removed. It was always `0`, never computed (the
bridge's own comment admitted "would require more expensive computation") — use
`Shape.isSelfIntersecting(timeout:)` for a real self-intersection check.

`ShapeAxis.extent` (from `Shape.revolutionAxes(tolerance:)` and
`Shape.symmetryAxes(fractionalTolerance:)`) is now genuinely computed rather than always `nil`: it
reports the axis's own shape's bounding box projected onto the axis direction, in real 3D units.
`Face.primaryAxis` is unaffected and still never populates `extent`.

```swift
// Before: always nil
let axis = Shape.cylinder(radius: 5, height: 20)!.revolutionAxes().first
axis?.extent  // nil, always, regardless of the cylinder's actual height

// After: a real measured extent
axis?.extent  // Optional(0.0...20.0) (or -20.0...0.0 — axis direction sign varies)
```

## SemVer impact

MAJOR. `ShapeAnalysisResult.selfIntersectionCount` is a public stored property that no longer
exists; any consumer reading it will not compile. Suggested migration: the field was always `0`
and never computed, so there is nothing to migrate a *value* to — replace the read with
`shape.isSelfIntersecting(timeout:)` if a real self-intersection check is actually needed, or
delete the read if it was only ever checking a constant.

`ShapeAxis.extent` changing from unconditionally-`nil` to sometimes-non-nil for
`revolutionAxes()`/`symmetryAxes()` is source-compatible (no signature change) but is a real
behavior change, called out under its own heading above; on its own this would be MINOR at most,
folded into the MAJOR bump the field removal already requires this release.

## Checklist

- [x] New or changed behavior is covered by a unit test in the same PR — two new tests (extent),
      two updated tests (removal, compile + arithmetic).
- [x] Every new test was run once with its subject broken, and the failure is reported above
      (extent tests). The removal has no equivalent injection cycle (see triage doc's
      "Build/compile fallout" section for why — it's a field deletion enforced by the compiler,
      not a new positive check).

## Notes for the reviewer

- The `hasExtent` finding directly contradicts an earlier census-only pass's verdict on the same
  candidate (see triage doc, "Verdict 2" section, for the full reasoning on why "gates into an
  Optional" isn't sufficient evidence that a field is fixed — it has to actually flip on some real
  path, and this one never did).
- `occtComputeAxisExtent`'s bounding-box-projection method is exact for the cases this PR's tests
  check (a bounded cylindrical face/shape along its own axis) and a safe (non-tight) enclosing
  interval for curved geometry viewed along an arbitrary direction otherwise (e.g. a sphere's
  symmetry axes, or a cone's axis) — not claimed to be a tight geodesic extent in general, and the
  doc update (`docs/reference/Geometry2D.md`) says so.
- Did not touch `Scripts/repro/726-unmeasured-values/README.md` (the earlier, broader census-only
  pass covering all 62 candidates) — it's a point-in-time investigation record, not something this
  PR's scope calls for updating; `triage-core.md` is the new, current source for this PR's 36.

### DIFF
--- a/Scripts/census-unmeasured-values.py
+++ b/Scripts/census-unmeasured-values.py
@@ -31,9 +31,17 @@ A derivation that can be re-run does not go stale the same way a hand-count does
 WHAT THIS DOES NOT DO, DELIBERATELY: this is a CENSUS, not a gate. It is not wired into
 `ci.yml`'s `gate-scripts` job, and its exit code in report mode is always 0 (barring a crash) even
 when it finds candidates. Every one of the sites below still needs a human verdict. Some are
-real instances of the class; several measured here are the OPPOSITE, a value that already uses the
-"make absence representable" fix #583/#595/#609 established (e.g. `OCCTShapeAxis.hasExtent`
-correctly gating `extentMin`/`extentMax` into a `nil` on the Swift side). And #726 itself scopes
+real instances of the class; several measured are the OPPOSITE, a value that already uses the
+"make absence representable" fix #583/#595/#609 established.
+
+DO NOT TRUST AN EXAMPLE IN THIS COMMENT OVER A MEASUREMENT. Until #763 this paragraph named
+`OCCTShapeAxis.hasExtent` as its example of the correct pattern, "correctly gating
+`extentMin`/`extentMax` into a `nil` on the Swift side". It was not correct. All three assignments
+in `OCCTBridge_Topology.mm` set it `false` and nothing anywhere set it `true`, so `ShapeAxis.extent`
+was unconditionally `nil` and the gating never gated anything. It was the same defect as
+`selfIntersectionCount`, one layer removed, dressed as the remedy for it. The census had flagged
+that exact site and a reader dismissed it from the shape of the code rather than by checking whether
+the flag ever flips. And #726 itself scopes
 the gate-script question ("a struct field assigned only a literal across every path") as a
 follow-up spike, not something this pass commits to.
 
--- a/Scripts/repro/726-unmeasured-values/triage-core.md
+++ b/Scripts/repro/726-unmeasured-values/triage-core.md
@@ -2,8 +2,8 @@
 
 Scope: `OCCTBridge_Topology.mm` (9), `OCCTBridge_Healing.mm` (9), `OCCTBridge_Properties.mm` (7),
 `OCCTBridge_BRepGraph.mm` (4), `OCCTBridge_Surface.mm` (2), `OCCTBridge_Modeling.mm` (2),
-`OCCTBridge_Spatial.mm` (1), `OCCTBridge_Geom2d.mm` (1), `OCCTBridge_Curve3D.mm` (1). The sibling
-36 (`OCCTBridge_Document.mm`/`OCCTBridge_IO.mm`) are triaged separately on
+`OCCTBridge_Spatial.mm` (1), `OCCTBridge_Geom2d.mm` (1), `OCCTBridge_Curve3D.mm` (1) — 36 lines
+total. The sibling 26 (`OCCTBridge_Document.mm`/`OCCTBridge_IO.mm`) are triaged separately on
 `chore/763-triage-document`.
 
 File:line below is where `python3 Scripts/census-unmeasured-values.py` reported the candidate on
@@ -21,17 +21,24 @@ label turned out to be checking the wrong thing.
 
 ## Verdict counts
 
-| Verdict | Count |
-|---|---|
-| 1. Not an instance | 34 |
-| 2. Compute it | 1 (7 fields, 2 functions — see below) |
-| 3. Make absence representable | 0 |
-| 4. Remove the field | 1 |
+Counted per census *line* (36 total, matching `census-unmeasured-values.py`'s own unit — one line
+per flagged `(function, field)` pair, so `OCCTShapeSymmetryAxes`'s two `OCCTShapeAxis`
+constructions at lines 694/713 share one line each for `extentMin`/`extentMax`/`hasExtent`/`kind`,
+per the script's own (function, field-name) dedup):
 
-35 of 36 lines are verdict 1; two lines carry the two non-1 verdicts. The `hasExtent` finding
-spans 7 of the 36 census lines (both `OCCTShapeRevolutionAxes` and `OCCTShapeSymmetryAxes`'s
-`extentMin`/`extentMax`/`hasExtent`, plus `OCCTShapeSymmetryAxes`'s `kind` which stays verdict 1
-after the fix), counted here as one implementation item.
+| Verdict | Count | Lines |
+|---|---|---|
+| 1. Not an instance | 29 | all except the 7 below |
+| 2. Compute it | 6 | `Topology.mm:653` `extentMin`/`extentMax`/`hasExtent`, `Topology.mm:694` `extentMin`/`extentMax`/`hasExtent` — one implementation (`occtComputeAxisExtent`, see below) covering both call sites |
+| 3. Make absence representable | 0 | — |
+| 4. Remove the field | 1 | `Healing.mm:303` `selfIntersectionCount` |
+
+`Topology.mm:694`'s fourth field, `kind`, stays **verdict 1** even though it sits on the same line
+as three verdict-2 fields — it is a real branch-selected constant (always 7, "symmetry"), not a
+fabricated measurement, and untouched by the extent fix. So of `OCCTShapeSymmetryAxes`'s four
+flagged fields at that line, three are verdict 2 and one is verdict 1; `OCCTShapeRevolutionAxes`'s
+three flagged fields at `:653` are all verdict 2. 29 (verdict 1) + 6 (verdict 2) + 1 (verdict 4) =
+36.
 
 ## Verdict 4: remove `selfIntersectionCount` — IMPLEMENTED
 
--- a/Scripts/repro/726-unmeasured-values/triage-core.md
+++ b/Scripts/repro/726-unmeasured-values/triage-core.md
@@ -1,0 +1,290 @@
+# #763 triage: the non-Document/IO half (36 of 62 candidates)
+
+Scope: `OCCTBridge_Topology.mm` (9), `OCCTBridge_Healing.mm` (9), `OCCTBridge_Properties.mm` (7),
+`OCCTBridge_BRepGraph.mm` (4), `OCCTBridge_Surface.mm` (2), `OCCTBridge_Modeling.mm` (2),
+`OCCTBridge_Spatial.mm` (1), `OCCTBridge_Geom2d.mm` (1), `OCCTBridge_Curve3D.mm` (1). The sibling
+36 (`OCCTBridge_Document.mm`/`OCCTBridge_IO.mm`) are triaged separately on
+`chore/763-triage-document`.
+
+File:line below is where `python3 Scripts/census-unmeasured-values.py` reported the candidate on
+`refactor/381-pass1b` at `809f148` (branch point for this work), before any fix in this PR moved
+lines around. Each row's evidence is independent verification per
+`okf/policies/measure-dont-assume.md`: the function was read in full and, where an OCCT call is
+involved, its real behaviour was checked against the header (or `occt-refman`), not inferred from
+the identifier or the census's own pattern name.
+
+An earlier census-only pass (`Scripts/repro/726-unmeasured-values/README.md`, written against an
+earlier commit on this same branch) already adjudicated most of these sites. Its verdicts are
+noted where they agree. **One of them is corrected here** — see the `hasExtent` rows below — which
+is itself the point of re-verifying rather than transcribing: a previous "GOOD (already fixed)"
+label turned out to be checking the wrong thing.
+
+## Verdict counts
+
+| Verdict | Count |
+|---|---|
+| 1. Not an instance | 34 |
+| 2. Compute it | 1 (7 fields, 2 functions — see below) |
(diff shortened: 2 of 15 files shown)


gsdali:
Verified the two behaviour changes independently. Both hold, and the `hasExtent` one is the most
useful thing this triage has produced.

## `hasExtent` never flipped, confirmed

On the base branch, every assignment in the bridge:

```
OCCTBridge_Topology.mm:653   a.extentMin = 0; a.extentMax = 0; a.hasExtent = false;
OCCTBridge_Topology.mm:694   a.extentMin = 0; a.extentMax = 0; a.hasExtent = false; a.kind = 7;
OCCTBridge_Topology.mm:713   a.extentMin = 0; a.extentMax = 0; a.hasExtent = false; a.kind = 7;
```

Three sites, all `false`, nothing anywhere sets it `true`. `ShapeAxis.extent` was unconditionally
`nil` for every caller of `revolutionAxes()` and `symmetryAxes()`.

## Why this is worth more than the fix

**`Scripts/census-unmeasured-values.py`'s own header comment named this exact site as its example of
the correct pattern**, telling the reader that `OCCTShapeAxis.hasExtent` was "correctly gating
`extentMin`/`extentMax` into a `nil` on the Swift side". The detector flagged the site, and its
documentation instructed the adjudicator to dismiss it.

So this is not "the census produced a false positive that turned out to be true". It is a defect
wearing the costume of the remedy for that defect. `selfIntersectionCount` returns a fabricated `0`;
`hasExtent` returns a fabricated `false`, which renders a real value unreachable. The second is
harder to see precisely because the absence-representable pattern is what we tell people to do.

That claim was still live in the script on this branch, so I have pushed a commit correcting it. I
wrote it as a warning naming what happened rather than as a corrected example, because the failure
was not a badly chosen example: it was a flagged site dismissed on the shape of the code without
anyone checking whether the flag ever flips. **#765 would have inherited this directly**, since it
is told to calibrate a proposed gate's false-positive rate against these verdicts.

## The other change

`selfIntersectionCount` removal is the seed instance and the right call. `isSelfIntersecting`
already answers the question through `BOPAlgo_ArgumentAnalyzer`, so the field had no reason to
exist, and a documented fabricated zero (#702) was honest rather than correct.

## Numbers I checked

Census 62 to 55, a drop of exactly the 7 fixed lines, so nothing was suppressed rather than fixed.
Injection matrix fails both new tests with the defect restored. 5483 tests, 0 failures.

Ready for your review. Both breaking changes are called out per-item in the body.

Issue #877, opened 2026-08-12
Two orphaned doc comments found via SwiftLint (Surface.swift, Shape+Modeling.swift)


gsdali:
## What

Found while configuring `SwiftLint`'s `orphaned_doc_comment` rule for the code-style rollout
(#876), not fixed there since fixing either one turned out to obligate a full swift-format sweep
of the whole file under that PR's own "if you touch it, you fix it" manifest rule, and a
~1900-diagnostic reformat of two large files is out of scope for a CI-infrastructure PR.

Both are real, structurally identical bugs: a doc comment separated from the declaration it
documents by an inserted `// MARK:` line.

1. **`Sources/OCCTSwift/Surface.swift:1329`** — a doc comment ("Convert this surface to an array
   of Bezier surface patches...") sits orphaned before `// MARK: - Surface Singularity Analysis
   (v0.37.0)`, with nothing between it and the MARK. The function it describes,
   `toBezierPatches()` (line 1426), currently has **no doc comment of its own** — this one belongs
   there. Fix: move the comment block to sit directly above `toBezierPatches()`.

2. **`Sources/OCCTSwift/Shape+Modeling.swift:405`** — a doc comment ("Create a pipe shell with
   transition mode control...") sits orphaned before `// MARK: - Variable-Section Sweep
   (v0.21.0)`. Unlike case 1, this one is pure duplication: `pipeShell(spine:profile:mode:...)`
   (line ~391) already has its own complete doc comment covering the same `transition` parameter
   verbatim. Fix: delete the orphaned block, nothing to move.

## Why this matters beyond the two instances

Both are the same failure shape SwiftLint's `orphaned_doc_comment` rule is meant to catch, and
both slipped through undetected until this policy rollout added a check for it — exactly the
"no way of knowing the comments are up to date" problem the wider code-style proposal is about.

gsdali:
Fixed by PR #939 (6106b709), merged to `main`. The PR title used the `fix(#N)` form without a `Closes` keyword, so this did not auto-close; verified the work is present on `main` before closing.

A CHANGELOG entry for this landed in the release-docs sweep (`## Unreleased`), which the merge itself did not carry.

PR #912, opened 2026-08-15
fix(#910): guard ThruSectionsBuilder.generatedFace(from:) on IsDone()

### LINKED ISSUE #910: ThruSectionsBuilder.generatedFace(from:) doesn't check IsDone(), can return stale data after a failed reused build()
### OCCTSwift version

Current `main` (found during PR #909's review, unreleased).

### Platform

macOS (arm64), likely all platforms — bridge code, not platform-specific.

### Description

`ThruSectionsBuilder.generatedFace(from:)` (`OCCTThruSectionsGeneratedFace` in
`Sources/OCCTBridge/src/OCCTBridge_Modeling.mm`) reads the underlying
`BRepOffsetAPI_ThruSections::GeneratedFace()` without checking `IsDone()` first:

```objc
OCCTShapeRef OCCTThruSectionsGeneratedFace(OCCTThruSectionsRef ref, OCCTShapeRef edge) {
    auto ts = (OCCTThruSections*)ref;
    if (!ts || !edge) return nullptr;
    try {
        TopoDS_Shape face = ts->builder->GeneratedFace(edge->shape);
        if (face.IsNull()) return nullptr;
        return new OCCTShape{face};
    } catch (...) { return nullptr; }
}
```

The doc comment on the Swift side already says "Get the face generated from an edge **after
building**" (`Sources/OCCTSwift/ThruSectionsBuilder.swift`), but nothing enforces that — the
bridge function runs unconditionally, and OCCT's own `GeneratedFace()`/the wall-edge map it reads
(`myEdgeFace`, populated inside `CreateSmoothed()`'s per-edge loop) isn't cleared between `Build()`
calls.

A `ThruSectionsBuilder` reused across multiple `build()` calls — the same reuse pattern this
project's own test suites already exercise for other builders (`FilletBuilder`/`ChamferBuilder`'s
`doubleBuild()` tests in `Tests/OCCTStressTests/StressBuilderLifecycleTests.swift`) — can leave
`generatedFace(from:)` returning stale or partially-mixed data:

- If a `build()` call fails partway through `CreateSmoothed()` (including, since #905, the new
  `k >= 2` non-planar capping failure — but any other exception this file already throws has the
  same shape), the per-edge loop that populates `myEdgeFace` may have already run before the
  failure, so the map holds fresh bindings for this failed attempt's wall edges mixed with
  whatever was left from an earlier, successful `build()` call on the same builder for edges the
  failed attempt didn't touch.
- A caller that ignores the `build()` return value (or `isDone`/`checkResult`) and calls
  `generatedFace(from:)` anyway gets this mixed/stale data silently — no null, no error.

### Steps to reproduce

Not yet reduced to a minimal repro. The mechanism: build a `ThruSectionsBuilder` successfully once
(populating `myEdgeFace` for its wall edges), reuse the same builder for a second `build()` call
that fails partway through `CreateSmoothed()` (e.g. a `k >= 2` non-planar closed section per #905,
before that fix; or any other `Standard_Failure` this file already throws, unconditionally
reachable), then call `generatedFace(from:)` on an edge from either build and observe it returns
non-nil real geometry regardless of which `build()` (if either) actually succeeded.

### Crash log or error output

Not a crash — silent stale/incorrect data, same shape as the class of bug #905 itself was.

### Notes

Found during code review of PR #909 (the #905 fix). Not introduced or meaningfully worsened by
that PR: the same partial-mutation-before-failure shape already existed for every other exception
`CreateSmoothed()`/`CreateRuled()` can throw (`BRepFill_CompatibleWires` failures,
`Georges.GetStatus()` failures), so this is a pre-existing gap in how the bridge exposes a
stateful, reusable `ThruSectionsBuilder`, not a regression from #905's capping-guard fix. Filed
separately per this project's scope-boundary policy rather than folded into #909.

Candidate fix: guard `OCCTThruSectionsGeneratedFace` on `ts->builder->IsDone()` before reading
`GeneratedFace()`, matching the pattern most other post-build accessor bridge functions already
use — needs auditing whether other `ThruSectionsBuilder` accessors (and other reusable builders'
post-build accessors generally) have the same gap before deciding whether this is a one-function
fix or a class of fix.

gsdali:
## What & why

`ThruSectionsBuilder.generatedFace(from:)` (`OCCTThruSectionsGeneratedFace`) read OCCT's
`GeneratedFace()` without checking whether the last build actually succeeded. `GeneratedFace()` is
a bare lookup into `myEdgeFace`, which `Build()` never clears between calls, so a builder reused
after a failed rebuild kept answering from a prior successful build's (or the failed build's own
partial) bindings — silently, no null, no error.

**Revised after review round 1** (see the PR's own comment thread): the first commit guarded on
`ts->builder->IsDone()`, matching `OCCTThruSectionsShape`'s existing pattern — and that guard is
incomplete. `BRepOffsetAPI_ThruSections::Build()` has two punctual-section validation loops (a
wholly-degenerate MIDDLE section, reachable via the public `addVertex()` at an interior position)
that return `WrongUsage` **without ever calling OCCT's `NotDone()`**. On a builder that already
built successfully once, `IsDone()` stays stale-true through that failed rebuild. The shipped fix
is a bridge-side `built` field on the `OCCTThruSections` wrapper, set exactly once in
`OCCTThruSectionsBuild` via `IsDone() && GetStatus() == BRepFill_ThruSectionErrorStatus_Done`.

**Revised again after review round 2**: `built` alone is still not enough. OCCT's `myEdgeFace` map
is never *cleared*, only ever added to — so a THIRD build succeeding after an intervening failure
(build ok → add a mismatched section, build fails → `checkCompatibility(true)` reconciles it,
build ok again) can rebuild every section's edges, not just the newly-added one's, stranding an
*earlier* build's edge → face binding in the map without ever overwriting it. Measured empirically
before this round's fix: `generatedFace(from:)` answered non-nil with a face that was provably not
part of the successful third build's own `shape`. `OCCTThruSectionsGeneratedFace` now confirms the
face it finds is actually a member of the current `Shape()` (`TopExp_Explorer`) before returning
it, instead of trusting the map. Round 2 also found and fixed a related inconsistency: the six
builder-configuration setters (`setSmoothing`, `setMaxDegree`, `setContinuity`,
`checkCompatibility`, `setParType`, `setCriteriumWeight`) didn't invalidate `built` the way
`addWire`/`addVertex` already did, so changing a setting right after a successful build used to
leave `.shape` silently serving pre-change geometry. All eight mutators now invalidate
consistently, and the doc comments on `.shape`/`generatedFace(from:)` were rewritten to describe
that actual contract instead of the narrower "nil if the last build() call failed" one they no
longer matched.

Closes #910

### `ThruSectionsBuilder` no longer returns stale results after a failed rebuild, or after a build on a changed builder that hasn't been rebuilt (#910)

`generatedFace(from:)` and `shape` both read post-build OCCT state without reliably checking
whether the *last* `build()` call on that instance actually succeeded, and without accounting for
OCCT's own internal state never resetting between builds or setting changes on a reused builder.
Fixed in two rounds:

- **Outcome tracking.** A `ThruSectionsBuilder` reused across multiple `build()` calls could
  silently keep returning a prior successful build's geometry — through `generatedFace(from:)`,
  `shape`, and `build()`'s own return value — after a later `build()` call on the same instance
  genuinely failed, including failures that OCCT's own `IsDone()` does not reliably report on a
  reused builder (a punctual middle section added via `addVertex()`).
- **Stale binding after a later success.** Even with outcome tracking, a builder that succeeded,
  then failed, then succeeded again could still have `generatedFace(from:)` answer with a face left
  over from the *first* success rather than the current one — OCCT's internal edge→face map is
  additive-only. `generatedFace(from:)` now confirms its answer is actually part of the current
  `shape` before returning it.
- **Setting changes without a rebuild.** `setSmoothing`, `setMaxDegree`, `setContinuity`,
  `checkCompatibility`, `setParType`, and `setCriteriumWeight` now invalidate a prior successful
  build the same way `addWire`/`addVertex` already did — previously, changing a setting after a
  successful build left `.shape`/`generatedFace(from:)` silently serving geometry built under the
  old setting until the caller happened to add a new section too.

Fixed by tracking the real build outcome bridge-side rather than trusting `IsDone()` alone, and by
verifying `generatedFace(from:)`'s answer against the current build's own shape rather than
trusting OCCT's internal map.

## SemVer impact

PATCH. A caller of `ThruSectionsBuilder.build()`, `.shape`, or `.generatedFace(from:)` who reuses a
builder across `build()` calls now correctly gets `false`/`nil` after any failed rebuild, or after
changing a setting without rebuilding, instead of stale or wrong geometry from an earlier build. No
signature change; the only observable difference is a previously-buggy return value becoming
correct — including for callers who call `setSmoothing`/`setMaxDegree`/`setContinuity`/
`checkCompatibility`/`setParType`/`setCriteriumWeight` between a successful `build()` and reading
`.shape`/`generatedFace(from:)` without an intervening rebuild: that combination previously
returned the pre-change build's geometry and now correctly returns `nil` until `build()` is called
again. No migration.

## Checklist

      verification): five tests in `Tests/OCCTStressTests/StressBuilderLifecycleTests.swift` —
      `generatedFaceNilAfterFailedRebuild` (the original open/closed-mismatch scenario),
      `generatedFaceNilAfterWrongUsageOnReusedBuilder` (the `IsDone()`-defeating scenario round 1's
      review found), `generatedFaceNilAfterFailedRebuildRuledPath` (coverage for `CreateRuled()`'s
      different `myEdgeFace`-binding mechanism, now genuinely reached via `isRuled: true` + 3
      sections rather than the 2-section shortcut round 2's review caught it silently relying on),
      `generatedFaceIsMemberOfShapeAfterSuccessFailureSuccessOnReusedBuilder` (round 2's finding 1:
      success → failure → success no longer returns a stale binding), and
      `shapeNilAfterSettingChangedWithoutRebuild` (round 2's finding 2: all six setters invalidate
      a prior build, not just `addWire`/`addVertex`).
      failure is reported here:
      - Round 1, first commit: reverted the `IsDone()` guard and re-ran the original test — it
        failed with `Expectation failed: (loft.generatedFace(from: edge) → OCCTSwift.Shape) == nil`.
      - Round 1, second commit: reverted the bridge back to the `IsDone()`-only guard and re-ran
        `generatedFaceNilAfterWrongUsageOnReusedBuilder` — it failed with exactly the three
        symptoms the review predicted (`build() == true`, `shape != nil`, `generatedFace != nil`).
      - Round 2: removed the `TopExp_Explorer` membership check and re-ran
        `generatedFaceIsMemberOfShapeAfterSuccessFailureSuccessOnReusedBuilder` — failed with
        `Expectation failed: isMember`. Removed all six setters' `built = false` resets and re-ran
        `shapeNilAfterSettingChangedWithoutRebuild` — failed at both assertions (setContinuity and
        checkCompatibility). Also independently verified — by deleting the `addWire`/`addVertex`
        resets and running the full pre-round-2 `ThruSectionsBuilder`-touching test set (130 tests
        across 6 files) — that no *existing* test caught that reset's removal, confirming review
        round 2's finding 7 before adding coverage for it.
      - Restored the real fix after each injection; final state: all 5 required static gates clean,
        full `OCCTModelingTests` (645/645) and `OCCTStressTests` (365/365, +2 over round 1's 363)
        pass.
- [ ] Every static gate script is clean **except** `code-style`: `check-style-manifest.py` fails on
      this head because both touched `OCCTBridge_Modeling.*` files are still on
      `Scripts/style-manifest-bridge.txt`. `code-style` is not a required check (only
      `gate-scripts` is); bringing a 24k/3.4k-line pair of files into clang-format compliance is
      out of scope for this fix and is tracked as a deferred sub-issue of the format-check work,
      [#917](https://github.com/SecondMouseAU/OCCTSwift/issues/917) — round 2's review flagged that
      the previous version of this checklist implied full cleanliness inaccurately; this line now
      says what's actually true.

## Notes for the reviewer

Found during PR #909's review (the #905 fix). Filed separately per this project's scope-boundary
policy rather than folded into #909.

The deferred SIGSEGV originally noted here (`checkCompatibility(false)` + mismatched section edge
counts on a reused builder) has since been root-caused, patched, and shipped separately:
[#913](https://github.com/SecondMouseAU/OCCTSwift/issues/913),
[OCCT#1466](https://github.com/Open-Cascade-SAS/OCCT/pull/1466),
[#915](https://github.com/SecondMouseAU/OCCTSwift/pull/915) (verified against a full local kernel
rebuild, 5533/5533 tests).

**Round 2 review disposition** (12 findings, all independently re-verified against the pinned
`V8_0_1` OCCT source before acting on any of them, per this project's own practice — several
findings were confirmed by reading `BRepOffsetAPI_ThruSections.cxx` directly rather than trusting
the review's claim, and one candidate fix (finding 8's original framing) turned out to need its own
separate issue rather than a quick patch once the actual mechanism was traced):

- **1 (blocking, fixed):** `generatedFace(from:)` membership check against the current `shape`.
- **2 (fixed):** all six setters now invalidate `built`.
- **3 (fixed):** doc comments (Swift + bridge header) rewritten to match the real contract.
- **4 (fixed):** CHANGELOG entry and SemVer paragraph above now cover the setter-invalidation
  behavior change.
- **5 (fixed):** checklist's static-gate line above now states `code-style`'s real, deferred
  status instead of implying it's clean.
- **6 (fixed):** struct comment no longer cites `OCCTSectionBuilder` as a working precedent — it
  has the identical unfixed bug, tracked at #916.
- **7 (fixed):** added `shapeNilAfterSettingChangedWithoutRebuild`; independently confirmed via
  injection that no prior test caught either the setter gap or the original `addWire`/`addVertex`
  reset's removal.
- **8 (filed as a follow-up, out of scope here):**
  [#919](https://github.com/SecondMouseAU/OCCTSwift/issues/919) — traced to
  `BRepOffsetAPI_ThruSections::SetCriteriumWeight` itself silently no-oping on a negative weight
  (verified directly in the OCCT source, `myStatus = Failed` but `myCritWeights` left unchanged),
  compounded by `Build()` unconditionally resetting `myStatus = Done` on entry so the failure is
  unobservable even via `GetStatus()` by the time a caller checks. Unrelated to `generatedFace(from:)`
  staleness; needs its own bridge-signature change (`setCriteriumWeight` would need to return
  `Bool`) rather than fitting this PR's shape.
- **9 (fixed):** added an in-source `swift` snippet to `ThruSectionsBuilder`'s type doc, since none
  of its doc comments had one before.
- **10 (fixed):** `generatedFaceNilAfterFailedRebuildRuledPath` now uses `isRuled: true` with 3
  sections instead of silently relying on the 2-section `CreateRuled()` dispatch shortcut.
- **11 (not actionable here):** `ThruSectionsBuilder`'s unsynchronized `built` field on an
  `@unchecked Sendable` type matches every sibling builder class in this codebase (none guard their
  internal mutable state with a per-instance lock) — this project's documented thread-safety model
  is caller-side serialization via `OCCTSerial.withLock`, not per-field locking
  (`docs/thread-safety.md`), and the class-wide question is already tracked project-wide by #342.
  Adding a mutex to just this one field would be inconsistent with every other builder and
  wouldn't actually make concurrent same-instance use safe (the underlying OCCT builder pointer
  itself isn't reentrant either).
- **12 (fixed):** trimmed the struct's `built` comment and the `AddWire`/`GeneratedFace` comments
  that had grown to restate the same three facts multiple times across this PR's two rounds.


gsdali:
Pushed a response to the review's findings (commit 1ac2129):

- **Fixed, verified empirically before and after** (finding 1, the load-bearing one): `IsDone()`
  alone is the wrong gate. Reproduced the exact `AddVertex()`-at-an-interior-position scenario the
  review predicted — `Build()`'s "wholly-degenerate middle section" check returns `WrongUsage`
  without calling `NotDone()`, so a reused builder's `IsDone()` stays stale-true through a failed
  rebuild. Confirmed live: `build()` returned `true`, `shape`/`generatedFace(from:)` both still
  served the prior build's stale geometry.
- **Fixed** (finding 2): `OCCTThruSections` now carries a `built` field, set once in
  `OCCTThruSectionsBuild` via `IsDone() && GetStatus() == BRepFill_ThruSectionErrorStatus_Done`.
  `OCCTThruSectionsShape` and `OCCTThruSectionsGeneratedFace` both gate on that instead of
  re-deriving from OCCT state, closing the gap for all three ways `Build()`'s return value could
  diverge from what the accessors saw.
- **Fixed** (findings 3, 7, 9): the existing test is now `throws` + `try #require(...)` instead of
  silent-skip `guard let ... else { return }`; added the sibling `loft.shape == nil` assertion;
  dropped the no-op `checkCompatibility(true)` call (already OCCT's own default).
- **Fixed** (finding 4): added a ruled-path variant (`isRuled: true`) — `CreateRuled()` binds
  `myEdgeFace` via `BRepFill_Generator`, a different mechanism from `CreateSmoothed()`'s own loop.
- **Fixed** (finding 6): `///` doc comments on both `generatedFace(from:)` and `shape`, plus
  `docs/reference/Document-Completions.md`, now state the "nil if the last build did not succeed"
  contract.
- **Added**: a dedicated regression test for finding 1's own mechanism
  (`generatedFaceNilAfterWrongUsageOnReusedBuilder`) — proved it fails against the pre-fix
  `IsDone()`-only guard with exactly the three symptoms predicted (`build() == true`,
  `shape != nil`, `generatedFace != nil`), then passes after the `built`-flag fix. Full
  `StressThruSectionsBuilderLifecycleTests` (8/8), `OCCTModelingTests` (645/645), `OCCTStressTests`
  (363/363) all green.
- **Noted, not changed** (finding 5): the positive assertion's edge identity does depend on
  `BRepFill_CompatibleWires` leaving already-matching-topology input edges untouched — true today,
  not a documented guarantee. Left it in place but called out the dependency inline: the suggested
  alternative (deriving the edge from `loft.shape`) exercises different, equally-untested
  key-derivation semantics of `GeneratedFace()`, and the assertion this test actually exists to
  protect (the negative one, after the failed rebuild) doesn't depend on this identity holding
  across OCCT versions either way — it's checking for `nil`, not a specific face.
- **De-risked, not part of this PR** (finding 8): the deferred SIGSEGV this PR's own body flagged
  (`checkCompatibility(false)` + mismatched section edge counts on a reused builder) has since been
  root-caused, fixed, and filed separately: [#913](https://github.com/SecondMouseAU/OCCTSwift/issues/913),
  upstream [OCCT#1466](https://github.com/Open-Cascade-SAS/OCCT/pull/1466), OCCTSwift
  [#915](https://github.com/SecondMouseAU/OCCTSwift/pull/915) — verified against a full local
  kernel rebuild, 5533/5533 tests passing.

gsdali:
Ran my own review pass before the formal one, including dispatching `/code-review 912 high` as a
second, independent check. Combined findings, pushed a response (commit 7b0abee):

- **Fixed** (code-review finding, real CI blocker): `code-style` was failing — this PR's 3 touched
  files were still listed in `Scripts/style-manifest-{swift,bridge}.txt`. Brought
  `ThruSectionsBuilder.swift` into full compliance (one import-order fix) and removed it from the
  manifest. **`OCCTBridge_Modeling.{h,mm}` are not addressed**: measured a full `clang-format` pass
  at 3,392 and 24,133 diff lines respectively — disproportionate for this PR, matching
  `okf/policies/code-style.md`'s own `#877` precedent for exactly this situation. `code-style` isn't
  a required status check, so this doesn't block merging; flagging for a call on whether to defer
  via a tracked follow-up or absorb here.
- **Fixed** (code-review finding 4, real): all three new tests' successful build used exactly 2
  sections, and `Build()` dispatches any 2-section call to `CreateRuled()` regardless of `isRuled`
  — so `CreateSmoothed()`'s own `myEdgeFace` binding was never exercised, despite two tests'
  comments claiming otherwise. Redesigned the two "smoothed path" tests to use 3 sections; re-ran
  the revert-and-confirm-it-fails cycle against the redesigned fixture.
- **Fixed** (code-review finding 5): corrected a test comment that overclaimed coverage of finding
  1's specific mechanism.
- **Fixed** (code-review finding 6, real, pre-existing): `AddWire`/`AddVertex` never reset `built`,
  so a section added after a successful build without a following `build()` call left
  `shape`/`generatedFace(from:)` silently stale. Fixed; no existing test relied on the old
  behavior.
- **Fixed** (code-review finding 7): the new doc comment on `generatedFace(from:)` was
  self-contradictory on a literal read. Reworded.
- **Documented, not changed** (code-review finding 3): added the missing rationale for fixing this
  bridge-side rather than via a kernel patch, directly on the `built` field.
- **Filed separately**, not folded in: `OCCTSectionBuilder` (same file) has the identical staleness
  bug this PR fixes for `OCCTThruSections` — `built` is only ever set `true`, never reset on a
  failed rebuild. [#916](https://github.com/SecondMouseAU/OCCTSwift/issues/916).

Verified: full `StressThruSectionsBuilderLifecycleTests` (8/8), `OCCTModelingTests` (645/645),
`OCCTStressTests` (363/363), and every static gate script clean.

gsdali:
Pushed round 2 fixes (`115f17a`). Independently re-verified every finding against the pinned V8_0_1 OCCT source (not just applying the review at face value) before acting on any of them — full disposition, including where I diverged (finding 8 became its own issue rather than a quick patch once traced, finding 11 rejected as inconsistent with the rest of the codebase) is in the updated PR body. Summary:

- **1 (blocking):** confirmed live via a throwaway probe before fixing — `generatedFace(from:)` after a success→failure→success sequence returned a face that was provably not part of the third build's own shape. Fixed with a membership check against the current `Shape()`.
- **2, 3, 6, 7, 10, 12:** fixed as described.
- **4, 5:** addressed in the PR body itself (CHANGELOG/SemVer now cover the setter-invalidation change; checklist now states `code-style`'s real status instead of implying cleanliness).
- **8:** real, but a separate defect (`SetCriteriumWeight` itself no-ops on a negative weight, and `Build()` overwrites the only status that would report it) — filed as #919, sub-issue of #910.
- **9:** added an in-source snippet.
- **11:** not actionable in this PR — matches the existing unsynchronized-state pattern on every sibling builder class; project-wide question tracked by #342.

All 5 required static gates clean, full `OCCTModelingTests` (645/645) and `OCCTStressTests` (365/365, +2) pass. `code-style` still fails as documented (deferred to #917, not a required check).

gsdali:
**swift build + test (macOS) investigation, resolved.** This job failed 3/3 reruns on `115f17a` (round 2's commit) with a double-SIGSEGV crash. Investigated thoroughly before concluding anything:

- Round 1's commit (`7b0abee`) passed this same job cleanly.
- Reran `main`'s CI twice (same day, same runner pool, unrelated code) as a control — **2/2 clean**.
- Ran the full local suite 7 times on this exact commit — **7/7 clean**, 5537/5537 tests each.

Net: a real, CI-runner-specific signal (not general infra flakiness — the control rules that out; not a deterministic bug — 7 clean local runs rule that out), not reproducible locally, root cause not established. Filed as [#920](https://github.com/SecondMouseAU/OCCTSwift/issues/920) with the full evidence rather than dismissed or left silent. `swift build + test (macOS)` is not a required check (`gate-scripts` is, and it's green), so not treating this as blocking, but flagging it plainly rather than quietly rerunning until green.

gsdali:
Went back through this round's **12 inline comments** individually rather than relying on the summary body I'd worked from — the inline comments carried more precise detail, and cross-checking each one against the actual diff (`f3efeb8`) turned up six real gaps between what the review asked for and what I'd actually done:

- **Finding 3**: only fixed 2 of the 3 doc locations it named, and missed a 4th (`docs/reference/Shape-Completions.md`'s `shape` entry) with the same stale wording. Fixed both.
- **Finding 6**: fixed the checklist text but not what was actually asked — an inline `# deferred, see #917` annotation in `style-manifest-bridge.txt` itself, and (separately) verified its prediction that `swift-format lint --strict` would fail once `ThruSectionsBuilder.swift` left the Swift manifest. It was right — confirmed directly and fixed.
- **Finding 9**: added a class-level snippet, but the ask was specifically `- Parameter`/`- Returns` + a fenced snippet on `generatedFace(from:)`. Added.
- **Finding 10**: fixed the test's `isRuled` gap but missed its second half — the fragility caveat on the positive `!= nil` assertion, present on the original test but not the two I touched. Added to both.
- **Finding 11**: responded in the PR body but didn't do the "minimum" ask — an in-source `///` note that instances aren't safe for concurrent use. Added.
- **Finding 12**: asked to keep the "bridge-side, not a kernel patch" paragraph in full while trimming the rest — my trim dropped it entirely. Restored.

Also caught, while re-reading finding 8 closely: a second silent-discard I'd missed when filing #919 (`setContinuity` unconditionally resets criterium weights to `1,1,1`, discarding a caller's prior `setCriteriumWeight` with no diagnostic) — added as a follow-up comment on #919 rather than a new issue, since it's the same investigation.

Finding 5's shared-mixin suggestion is still consciously not done (real refactor, touches #916's in-flight fix, out of proportion here) — flagging that explicitly rather than letting it look overlooked.

All 5 required gates clean (re-verified `check-style-manifest.py` against the actual merge-base, not the since-moved `origin/main` tip — local runs against a live branch ref can show spurious noise from unrelated merges). Full `OCCTModelingTests`/`OCCTStressTests` pass.

PR #915, opened 2026-08-15
fix(#913): guard CreateSmoothed's fixed-stride shapes array against a mismatched section edge count

### LINKED ISSUE #913: ThruSectionsBuilder checkCompatibility(false) + mismatched section edge counts SIGSEGVs on a reused builder
### OCCTSwift version

Current `main` (found while writing PR #912's #910 fix, unreleased).

### Platform

macOS (arm64), likely all platforms — bridge/kernel code, not platform-specific.

### Description

While probing for a deterministic way to make a *reused* `ThruSectionsBuilder`'s second `build()`
call genuinely fail (for #910), I hit a SIGSEGV with this shape:

1. `ThruSectionsBuilder(isSolid: true, isRuled: false)`, `checkCompatibility(false)`.
2. Add two closed circle wire sections, `build()` — succeeds.
3. Reuse the same builder: add a third section with a very different edge count from the first two
   (a triangle, 3 edges, vs. each circle's 1), `build()` again.

The second `build()` call (or a `generatedFace(from:)` call shortly after it — I couldn't fully
isolate which, see below) crashes the process with SIGSEGV, no exception, no catchable failure.

### Suspected mechanism

`BRepOffsetAPI_ThruSections::Build()` only reconciles differing section edge counts through
`BRepFill_CompatibleWires` when `myWCheck` (`checkCompatibility`) is `true`. With it `false`,
`CreateSmoothed()`'s per-edge walk assumes every consecutive pair of sections has a compatible
edge topology and doesn't defend against a mismatch — the "no check" path
(`BRepOffsetAPI_ThruSections.cxx`, the `else // no check` branch feeding `myEdgeNewIndices`)
takes each section's edges at face value. A circle (1 edge) followed by a triangle (3 edges),
reused on a builder whose *previous* successful build already populated internal per-section state,
looks like a plausible trigger for the lockstep edge-explorer walk in `CreateSmoothed()`
(`BRepOffsetAPI_ThruSections.cxx`, around the `anExp1`/`anExp2` loop shared with `CreateRuled()`)
running out of sync and reading/writing out of bounds.

This has NOT been root-caused — it's a "here's a crash, here's a plausible mechanism" report, not
a diagnosed defect. I could not cleanly isolate whether the crash is inside `Build()` itself or a
subsequent `generatedFace(from:)` call, because stdout was fully buffered and lost on the SIGSEGV;
isolating that (and whether it needs the "reused builder" precondition at all, or reproduces on a
single fresh builder with mismatched sections + `checkCompatibility(false)`) is the first step for
whoever picks this up.

### Steps to reproduce

Not yet reduced to a minimal, isolated repro (see above — the reused-builder precondition may not
even be necessary). The reproducing shape, in Swift:

```swift
let loft = ThruSectionsBuilder(isSolid: true, isRuled: false)
loft.checkCompatibility(false)
loft.addWire(circleWireShape1)   // 1 edge
loft.addWire(circleWireShape2)   // 1 edge
_ = loft.build()                 // succeeds

loft.addWire(triangleWireShape)  // 3 edges — very different topology, no compatibility layer
_ = loft.build()                 // SIGSEGV
```

### Crash log or error output

`Segmentation fault: 11`, uncaught by the bridge's own `try`/`catch (...)` in
`OCCTThruSectionsBuild` — consistent with an OS-level signal (out-of-bounds read/write) rather than
a `Standard_Failure` OCCT itself would raise and this bridge already catches.

### Notes

Found incidentally while investigating #910, not the same defect: #910 is a stale-data bug in
`generatedFace(from:)`'s bridge guard; this is a hard crash, and (as far as investigated so far)
looks kernel-side rather than bridge-side. Filed separately per this project's scope-boundary
policy rather than folded into #912.

gsdali:
## What & why

`ThruSectionsBuilder`, given `checkCompatibility(false)` and 3+ sections where a later section has
a different edge count than the first, either SIGSEGVs/SIGBUSes (more edges than section 1) or
silently succeeds with wrong geometry (fewer edges) instead of failing cleanly. Root cause is in
`BRepOffsetAPI_ThruSections::CreateSmoothed()`: it derives the edge count it assumes every section
has from section 1 alone, allocates a fixed-size array on that assumption, then fills it by walking
each section's *actual* edges with no bounds check. Found incidentally while hunting for a #910
failure trigger; filed separately as #913 per this project's scope-boundary policy.

**Revised after review** (see the PR's own comment thread — 12 findings, all addressed or
explicitly documented): the guard is symmetric by construction (an inequality test, not a "too
many" test), so it always rejected the fewer-edges direction too — the first version of this PR's
own SemVer note and patch description just undersold that half. The punctual-section exemption now
also verifies a section actually has an edge before exempting it (hardening, not a proven-live
fix — see notes below), and its predicate is hoisted into a lambda shared with the pre-existing
fill loop instead of duplicated.

Closes #913

### `ThruSectionsBuilder` no longer returns wrong or crashing results for a mismatched section edge count under `checkCompatibility(false)` (#913)

`BRepOffsetAPI_ThruSections::CreateSmoothed()` derived the edge count it assumes every section has
from section 1 alone, and filled a fixed-size array on that assumption with no bounds check. With
`checkCompatibility(false)`, nothing reconciles differing section edge counts first (the default,
`checkCompatibility(true)`, does this via `BRepFill_CompatibleWires`). A later section with **more**
edges than section 1 overran the array — heap corruption, observed as a SIGSEGV/SIGBUS once at
least 3 sections are involved (2 sections always take a different code path that doesn't share this
allocation shape). A section with **fewer** edges didn't crash, but silently misaligned per-section
strides to the wrong geometry, reporting `build() == true` for an invalid result. Both directions
are the same contract violation and are now rejected the same way. Fixed upstream:
[OCCT#1466](https://github.com/Open-Cascade-SAS/OCCT/pull/1466), carried as
`Scripts/patches/0027-*` and verified against a full local kernel rebuild (all 17 carried patches,
full `swift test` passing).

## SemVer impact

PATCH. A caller using `checkCompatibility(false)` with 3+ sections of differing edge counts now
gets a clean `build() == false` instead of either a process crash (more edges) or a silently wrong
successful result (fewer edges). No signature change, no migration — this corrects both directions
of the same contract violation into the "graceful failure" contract every other
`ThruSectionsBuilder` error path already has.

## Checklist

      verification): `mismatchedSectionEdgeCountWithoutCheckFailsCleanly()` (the original
      more-edges crash, gated — see below) and `punctualApexWithMatchingSectionsStillSucceedsUnderCreateSmoothed()`
      (the punctual-exemption regression guard, PR review finding 5 — a legitimate cone-apex loft
      under `CreateSmoothed()` must keep succeeding) in
      `Tests/OCCTStressTests/StressBuilderLifecycleTests.swift`. The fewer-edges direction is
      proven upstream instead, in the same GTest file this patch already touches (see below) —
      not duplicated at the Swift level.
      failure is reported here: ran `mismatchedSectionEdgeCountWithoutCheckFailsCleanly` filtered
      against the *currently pinned* (unfixed) remote kernel first — it crashed the whole
      `swift test` process with `exited with unexpected signal code 11` (SIGSEGV), matching the
      standalone C++ repro and PR #915's own original CI run exactly. Then rebuilt the kernel
      locally with the patch (`Scripts/build-occt.sh`, all 17 carried patches applying cleanly) and
      re-ran: passes, along with the full suite.

## Notes for the reviewer

**Root-cause method**: a standalone C++ reproducer (no Swift/bridge involved) with a custom
SIGSEGV/SIGBUS handler (`backtrace_symbols_fd`, since lldb/core dumps are blocked in this sandboxed
dev environment) isolated the crash to `CreateSmoothed()`. Instrumenting the fill loop directly
confirmed the write index reaches one past the array's upper bound before the corrupting
assignment. The original "reused builder" framing in #913's own report turned out to be a symptom,
not the cause: a single `Build()` call with all mismatched sections present from the start doesn't
reliably crash either, even though the identical out-of-bounds write still occurs — what varies is
process/allocator state at the time of the overrun. Full mechanism, the override-link validation
(with and without `No_Exception`/`NDEBUG`, matching a Release build), and the fewer-edges/zero-edge
findings from review live in `Scripts/patches/README.md`'s `0027` entry, and in a committed
reproducer at `Scripts/repro/913-thrusections-createsmoothed-section-edge-count-guard/`
(README + standalone `.mm` + `stock.txt`/`patched.txt` transcripts, matching #905's own convention
— missing from this PR's first version, per review finding 7).

**Test gating (review finding 1, the load-bearing one — confirmed live via this PR's own first CI
run, which crashed `swift build + test (macOS)` with exactly the predicted SIGSEGV)**:
`mismatchedSectionEdgeCountWithoutCheckFailsCleanly` is now `@Test(.enabled(if:
ProcessInfo.processInfo.environment["OCCTSWIFT_LOCAL"] == "1"))` — it needs patch 0027, which isn't
in `Package.swift`'s pinned kernel asset yet, and SwiftPM runs every test target in one process, so
an unguarded crash there would have aborted the whole suite for every future PR until the pin
moves, matching this project's own `#585` failure shape and `#905`/PR #909's precedent (which
avoided adding a Swift test for the identical reason). `kernel-integration.yml` sets
`OCCTSWIFT_LOCAL=1` when it builds `Scripts/patches/` from source, so the test runs (and is
verified) there, not against the pinned kernel.

**This PR does a full local kernel rebuild for verification** (`Scripts/build-occt.sh`, ~1hr,
`Libraries/OCCT.xcframework` is gitignored so it isn't in this diff) rather than deferring to
override-link evidence alone, unlike #905/#909's `0026`. `Package.swift`'s `url:`/`checksum:` pin is
**not** touched here — per the release process, that only moves at a release commit, and patch 0027
joins 0026 as carried-but-not-yet-in-the-pinned-asset until then.

Upstream: [OCCT#1466](https://github.com/Open-Cascade-SAS/OCCT/pull/1466) (open, mergeable, 4 GTests
total — 2 original plus a fewer-edges regression test added after review, both crash/wrong-result-
and-fix proven the same way as this PR's own Swift tests; a zero-edge GTest was written, found not
to distinguish pre/post-fix behavior on the pristine kernel, and removed rather than kept as
unproven coverage — see the PR #915 review thread, finding 4, for the full attempt).


gsdali:
Pushed a response to the review's findings (commit 702ff2c):

- **Fixed, and confirmed live** (finding 1, the load-bearing one): `mismatchedSectionEdgeCountWithoutCheckFailsCleanly`
  did SIGSEGV `swift build + test (macOS)` — checked this PR's own first CI run, which had
  genuinely failed with `SIGSEGV 'segmentation violation' detected`, exactly as predicted. Gated
  with `@Test(.enabled(if: ProcessInfo.processInfo.environment["OCCTSWIFT_LOCAL"] == "1"))`;
  verified it correctly skips without the env var and runs (and passes, against the rebuilt
  kernel) with it.
- **No action needed** (finding 2): already resolved by PR #912's own fix, cross-referenced only.
- **Fixed, verified empirically before documenting** (finding 3): built a probe confirming the
  exact claim — against the pristine kernel, a 2/1/3-edge input under `checkCompatibility(false)`
  reports `Build() == true` and `Shape().IsValid() == false`. The guard's inequality test already
  rejected this (no code change needed there), but the patch header, README, CLAUDE.md, and this
  PR's own SemVer note all previously undersold it as "crash → failure" only. Corrected everywhere.
- **Fixed** (finding 4): the punctual-section exemption now also verifies a section has at least
  one edge before exempting it. Tried to reproduce a live crash for the zero-edge case, including
  via a reused builder — couldn't. The hardening is kept (correct, cheap, matches what the
  exemption is supposed to mean), but documented honestly as unconfirmed rather than claimed as a
  proven fix; the GTest I first wrote for it turned out to pass even against pristine, unpatched
  code, so I removed it rather than ship coverage that doesn't prove anything.
- **Fixed** (finding 5): added `punctualApexWithMatchingSectionsStillSucceedsUnderCreateSmoothed` —
  nothing previously pinned a legitimate cone-apex loft succeeding once this patch's guard reaches
  `CreateSmoothed()` (3+ sections). Verified passing against both the pinned and the patched
  kernel, since the exemption itself is unmodified pre-existing behavior.
- **Fixed** (finding 6): added a `CLAUDE.md` Known OCCT Bugs entry, matching every other carried
  patch.
- **Fixed** (finding 7): committed a reproducer at
  `Scripts/repro/913-thrusections-createsmoothed-section-edge-count-guard/`, matching #905's own
  convention.
- **Fixed** (finding 8): `guard let ... else { return }` → `try #require(...)`.
- **Fixed** (finding 9): hoisted the punctual-section predicate into a lambda shared with the
  pre-existing fill loop, in both the local patch and the already-submitted
  [OCCT#1466](https://github.com/Open-Cascade-SAS/OCCT/pull/1466) (new commit pushed there).
- **Documented, not restructured** (finding 10): left the check where it is; the patch's own
  writeup now explains why (re-deriving `myNbEdgesInSection` from `Build()`'s existing "no check"
  loop isn't equivalent — it's a max, not section 1's own count — and moving the check would touch
  a more central, higher-risk function for a duplication concern, not a correctness one).
- **Fixed** (finding 11): `AddVertex()` creates a wire with exactly **one** degenerate edge, not
  zero — corrected the comment (and my own earlier README text) everywhere it appeared.
- **Fixed** (finding 12): both signals were genuinely observed, in different binaries — a
  custom-handler standalone reproducer showed SIGSEGV, a GTest binary with OS default signal
  handling showed SIGBUS. Not a contradiction to resolve to one; documented both, with which
  binary showed which.

Re-verified end to end after all of the above: full local kernel rebuild (all 17 patches, patch
0027 v2 correctly recognized as "already applied" against my in-place source edits, confirming the
regenerated patch file matches byte-for-byte), full `swift test` 5534/5534 passing, all 6 static
gates clean.

Issue #917, opened 2026-08-15
Bring OCCTBridge_Modeling.h/.mm into clang-format compliance (deferred from PR #912)


gsdali:
### What

`Sources/OCCTBridge/include/OCCTBridge_Modeling.h` and `Sources/OCCTBridge/src/OCCTBridge_Modeling.mm`
need a full `clang-format` pass and removal from `Scripts/style-manifest-bridge.txt`, per the
"if you touch it, you fix it" rule #876 established.

### Where this came from

PR #912 (#910's fix) touched both files while they were still on the exemption manifest, which
`Scripts/check-style-manifest.py` correctly flagged: `code-style` is red on that PR. Measured the
actual size of bringing each into compliance before deciding what to do about it:

```
clang-format -style=file Sources/OCCTBridge/src/OCCTBridge_Modeling.mm     -> 24,133 diff lines
clang-format -style=file Sources/OCCTBridge/include/OCCTBridge_Modeling.h  ->  3,392 diff lines
```

(`.mm` is 10,998 lines total, `.h` is 3,863 — the diff touches most of the file, not a narrow
region, consistent with #876's own rollout-day measurement of "multi-thousand-line-per-file
clang-format diffs" for `Sources/OCCTBridge`.)

Confirmed there's no partial/scoped shortcut: `Scripts/check-style-manifest.py` only checks
manifest *membership*, not formatting — the actual enforcement is
`.github/workflows/code-style.yml`'s `clang-format --dry-run --Werror` step, which runs against
every `OCCTBridge` file **not** on the manifest. Removing a file from the manifest without actually
reformatting it just moves the failure there instead.

### Why deferred rather than fixed in PR #912

A ~24,000-line reformat is an order of magnitude beyond a bug-fix PR's actual change, and would
make that PR's real diff unreviewable. This is the same "disproportionate for the PR at hand" shape
`okf/policies/code-style.md`'s own `#877` precedent describes (there: two orphaned doc comments,
deferred rather than obligating a full-file `swift-format` sweep under this same manifest rule).

### Scope

Both `OCCTBridge_Modeling.h` and `OCCTBridge_Modeling.mm` are large, shared files covering dozens
of unrelated OCCT operation families (not just `ThruSections`), so this is its own dedicated PR:
run `clang-format -style=file -i` on each, confirm the result still builds and the full test suite
still passes (a pure whitespace/formatting pass shouldn't change any behavior, but these are two of
[...]

PR #918, opened 2026-08-15
fix(#916): reset OCCTSectionBuilder's built flag on failed rebuild

### LINKED ISSUE #916: OCCTSectionBuilder's built flag never resets to false on a failed rebuild (same class as #910)
### OCCTSwift version

Current `main` (found during PR #912's review, unreleased).

### Platform

macOS (arm64), likely all platforms — bridge code, not platform-specific.

### Description

`OCCTSectionBuilder` (`Sources/OCCTBridge/src/OCCTBridge_Modeling.mm`, backs `SectionBuilder`) has
the identical staleness defect PR #912 just fixed for `OCCTThruSections`, in a sibling reused
builder in the same file:

```objc
struct OCCTSectionBuilder {
    BRepAlgoAPI_Section section;
    bool built;
    OCCTSectionBuilder() : section(), built(false) {}
    OCCTSectionBuilder(const TopoDS_Shape& s1, const TopoDS_Shape& s2) : section(s1, s2, false), built(false) {}
};

OCCTShapeRef OCCTSectionBuilderBuild(OCCTSectionBuilderRef builder) {
    if (!builder) return nullptr;
    try {
        builder->section.Build();
        if (!builder->section.IsDone()) return nullptr;
        builder->built = true;
        return new OCCTShape{builder->section.Shape()};
    } catch (...) { return nullptr; }
}

OCCTShapeRef OCCTSectionBuilderAncestorFaceOn1(OCCTSectionBuilderRef builder, OCCTShapeRef edge) {
    if (!builder || !edge || !builder->built) return nullptr;
    ...
}
```

`built` is only ever set to `true`, on a successful build. On a **failure** — `!IsDone()` — the
function returns `nullptr` without ever resetting `built` back to `false`. So on a `SectionBuilder`
reused via `init1(...)`/`init2(...)` + `build()` (matching the reuse pattern PR #912's own
regression tests exercise for `ThruSectionsBuilder`), a builder that already built successfully
once keeps `built == true` through a later failed rebuild — `ancestorFaceOn1(edge:)` and
`ancestorFaceOn2(edge:)` would then skip their guard and read state from the failed/incomplete
rebuild's `BRepAlgoAPI_Section` instance instead of correctly returning `nil`.

### Steps to reproduce

Not yet reduced to a minimal repro that reliably makes `BRepAlgoAPI_Section::Build()` itself fail
(`IsDone() == false`) — a quick attempt using non-intersecting/far-apart shapes did not trigger a
failure (the section is just empty, still `IsDone() == true`). Whoever picks this up needs to find
an actual failure trigger for `BRepAlgoAPI_Section::Build()` first (an invalid plane/surface
argument via `init1(plane:)`/`init2(surface:)` is the likely candidate, unexplored) before this can
be proven live the way PR #912's own regression tests were.

### Notes

Found during PR #912's review (the #910 fix) while confirming that `built` is an established
naming convention in this bridge (it is — this struct predates #912's own use of the same name for
`OCCTThruSections`). Filed separately per this project's scope-boundary policy rather than folded
into #912, and because it needs its own live repro before a fix can be validated the same way #910
was.

Candidate fix, mirroring #912's own pattern: `builder->built = false;` on the `!IsDone()` path (and
on the `catch (...)` path) in `OCCTSectionBuilderBuild`, before returning `nullptr`.

gsdali:
## What & why

`OCCTSectionBuilder`'s `built` field (gating `ancestorFaceOn1(edge:)`/`ancestorFaceOn2(edge:)`) was
only ever set `true` on a successful `Build()` and never reset on a failed rebuild — the same
staleness class PR #912 fixed for the sibling `OCCTThruSections` struct (#910). A `SectionBuilder`
reused via `init1`/`init2` + `build()` that had already built successfully once kept `built ==
true` through a later failed rebuild, so `ancestorFaceOn1`/`ancestorFaceOn2` skipped their guard
and read state from the failed/incomplete rebuild's `BRepAlgoAPI_Section` instance.

**No live repro existed when #916 was filed** — a quick attempt (two far-apart, non-intersecting
boxes) just produces an empty, still-`IsDone()` section, not a failure. This PR finds one. Read
`BRepAlgoAPI_BooleanOperation::Build()`/`BOPAlgo_PaveFiller::Init()` directly
(`Libraries/occt-src/...`) and swept 13 candidate triggers against the real pinned kernel:
self-intersecting/bowtie faces, coincident/duplicate solids (including two exactly-coincident
spheres), an empty compound, a degenerate collinear-point face, NaN and zero-coefficient plane
coefficients, an invalid #905-style uncapped loft solid, and the too-few-arguments path. None make
`Build()` cleanly fail on a builder that already built successfully once. The **only** clean
(non-throwing) trigger found is a literal null `TopoDS_Shape` argument
(`BOPAlgo_PaveFiller::Init()`'s own `AddError(new BOPAlgo_AlertNullInputShapes)` check) — and
`Shape` never wraps one, so it is **not reachable through OCCTSwift's public Swift API**.

Proved the defect live anyway, at the bridge boundary rather than through Swift: compiled the real,
**unmodified** `OCCTBridge_Modeling.mm` and drove its actual exported `OCCTSectionBuilder*` C
functions with a hand-constructed null-wrapping `OCCTShape` as the one input Swift's type system
can't produce. This found something worse than the issue anticipated — **an uncatchable SIGSEGV**,
not just a stale answer: the failed rebuild's `BOPAlgo_PaveFiller::Init()` bails via
`AddError(BOPAlgo_AlertNullInputShapes)` *before* setting its own `myDS`, and
`HasAncestorFaceOn1`/`HasAncestorFaceOn2` — reached only because `built` was wrongly still true —
dereferences it. Full writeup, driver, and before/after transcripts:
[`Scripts/repro/916-sectionbuilder-built-flag-stale/`](Scripts/repro/916-sectionbuilder-built-flag-stale/).

The fix mirrors #912's `OCCTThruSections` pattern: `built = false` on both the `!IsDone()` and
`catch (...)` paths in `OCCTSectionBuilderBuild`. It also mirrors #912's own second-review finding
(the `AddWire`/`AddVertex` staleness gap): all six
`Init1Shape`/`Init1Plane`/`Init1Surface`/`Init2Shape`/`Init2Plane`/`Init2Surface` functions now
reset `built = false` on success too, closing the adjacent "re-init after a successful build,
without rebuilding" gap — which **is** reachable through pure Swift (unlike the null-shape SIGSEGV
above), and is what the new Swift regression test exercises.

Closes #916

### `SectionBuilder` no longer returns stale results after a failed rebuild (#916)

`ancestorFaceOn1(edge:)` and `ancestorFaceOn2(edge:)` read post-build OCCT state gated on a
bridge-side `built` flag that was only ever set `true`, never reset on a failed rebuild. A
`SectionBuilder` reused across multiple `build()` calls could silently keep answering from a prior
successful build's geometry after a later `build()` call on the same instance genuinely failed —
and, in the worst case (an argument that fails `BRepAlgoAPI_Section::Build()`'s own internal
null-shape check), could crash instead of returning `nil`. Calling `init1`/`init2` again after a
successful build without a following `build()` call had the identical staleness gap. Fixed by
resetting the tracked build outcome on every path that invalidates the previous result.

## SemVer impact

PATCH. A caller of `SectionBuilder.ancestorFaceOn1(edge:)`/`.ancestorFaceOn2(edge:)` who reuses a
builder across `build()` calls, or calls `init1`/`init2` again without a following `build()`, now
correctly gets `nil` instead of stale geometry from an earlier build. No signature change; the only
observable difference is a previously-buggy return value becoming correct (or, in the crash case,
no longer crashing). No migration.

## Checklist

      verification): `ancestorFaceNilAfterReinitWithoutRebuild` in
      `Tests/OCCTStressTests/StressBuilderLifecycleTests.swift`, covering the half of #916 reachable
      through the public Swift API (the `Init*`-without-rebuild staleness gap). The other half (a
      genuinely failed `build()` itself, via the null-shape mechanism) is proven at the bridge
      boundary instead — see `Scripts/repro/916-sectionbuilder-built-flag-stale/` — since it isn't
      reachable through Swift's type-safe `Shape` API. Also fixed a fixture bug while writing this
      test: `Shape.box(width:height:depth:)` (no origin) centers at origin, while
      `Shape.box(origin:width:height:depth:)` treats `origin` as a *corner* — mixing the two (as an
      early draft of this test did, copying `SectionBuilderTests.sectionTwoShapes`'s own fixture
      verbatim) produces boxes that only touch tangentially along a shared face plane, not a genuine
      3D overlap, so no section edge ever resolved an ancestor face. Verified directly in C++ against
      the real kernel before rewriting the fixture.
      failure is reported here: reverted the `built = false` reset in `OCCTSectionBuilderInit1Shape`
      only, re-ran `ancestorFaceNilAfterReinitWithoutRebuild` — it failed with both
      `ancestorFaceOn1(edge:)` and `ancestorFaceOn2(edge:)` wrongly returning non-nil (`Expectation
      failed: ... == nil` on both). Restored the fix; the test and the rest of
      `StressSectionBuilderLifecycleTests` (7/7) pass again. Separately, at the bridge level: ran
      `occt_916_section_builder_stale` against the unmodified bridge file first (SIGSEGV, see
      `before.txt`), then against the fixed one (clean exit, correct `nil`, see `after.txt`).

## Notes for the reviewer

**`code-style` is expected red here, for the same reason as #912**: `OCCTBridge_Modeling.mm` is
still on `Scripts/style-manifest-bridge.txt` (a full `clang-format` pass is ~24,000 diff lines for
this file alone — disproportionate for a 2-function-signature change, and tracked by #917, a
sub-issue of #876). `code-style` is not a required status check (only `gate-scripts` is, confirmed
via the repo's rulesets), so this doesn't block merge. Unlike #912, this PR *did* bring
`SectionBuilder.swift` into full `swift-format`/`swiftlint` compliance and remove it from
`Scripts/style-manifest-swift.txt` — matching #912's own precedent for the analogous small Swift
file (import order, one blank-comment-line-after-summary fix per doc comment, one line wrap).

**Verified**: `StressSectionBuilderLifecycleTests` (7/7), full `OCCTModelingTests` (645/645), full
`OCCTStressTests` (361/361), all six static gate scripts (`check-bridge-index`,
`check-null-handle-guards`, `check-docs-defaults`, `check-docs-existence`,
`derive-bridge-header-split --verify`, `count-operations`), and `check-style-manifest.py --base
origin/main` (clean except the expected, precedented `OCCTBridge_Modeling.mm` exemption above).


gsdali:
**Self-review pass, per request, before your own review.** Fixed the `code-style` failure and independently verified (not just re-read) every substantive claim in the PR body.

**`code-style` fix**: merged current `origin/main` (brings in #912's already-merged `# deferred, see #917` annotation on `OCCTBridge_Modeling.mm` in `style-manifest-bridge.txt`, plus resolved a real, small conflict in `style-manifest-swift.txt` — this branch and an unrelated already-merged PR each removed a different file from the same list). Updated the annotation to note this is now confirmed recurring (#912 and #918 both hit it) rather than a one-off. `code-style` still fails, for the same tracked/deferred reason as #912/#915 — not a new problem.

**Self-review findings**:
- Re-ran the SIGSEGV repro myself rather than trusting `before.txt`/`after.txt`: compiled the unmodified bridge, reverted the fix via `git apply -R`, reproduced exit 139 directly. Caught my own mistake along the way — a first attempt reverted only `OCCTSectionBuilderBuild` and got a false negative, because the driver's `Init1Shape` call still carried the *other* half of the fix and independently reset `built` before the trigger could fire. Redid it as a full reverse-patch.
- Re-ran the "prove the test fails" claim for `ancestorFaceNilAfterReinitWithoutRebuild" by injecting the same defect myself — matches the PR body exactly.
- Verified the `Shape.box(origin:...)` corner-vs-center fixture claim against the actual doc comments.
- **The one question this PR's own history should have raised but didn't address**: does `OCCTSectionBuilder` have #912's *finding-1* class of gap — a `built` flag insufficient on its own because internal state accumulates across builds (needing a membership check too, the way `OCCTThruSections` ended up needing)? Checked the actual OCCT source (`BRepAlgoAPI_BuilderAlgo::IntersectShapes`): `myDSFiller` is `delete`+`new`'d fresh on *every* `Build()` call — architecturally the opposite of `ThruSections`' additive-only `myEdgeFace` map. A simple `built` flag is genuinely sufficient here; confirmed empirically with a throwaway probe that `setApproximation`/`computePCurveOn1/2` not resetting `built` is correct, not a gap (`SectionBuilder` has no cached `.shape` accessor to go stale — `build()` always re-executes fresh, unlike `ThruSectionsBuilder`).
- Found and fixed one real gap: no fenced ```swift snippet on the changed public API (CLAUDE.md's Documentation Standards — the same class of finding #912's own round 2 caught). Added to `ancestorFaceOn1(edge:)`/`ancestorFaceOn2(edge:)`, plus `- Parameter`/`- Returns` structure.

All 5 required static gates clean, full `swift test` (5611/5611), `SectionBuilder`-specific suites (11/11) re-verified after every change. Pushed as `fac32a1`.

Issue #388, opened 2026-07-26
Pass 4d — Mesh + presentation/misc lane duplication audit


gsdali:
Part of the segmented duplication audit in #377. **Passes 2a (#382), 2b (#383) and 3 (#384) are the prerequisites** — check they are closed before starting.

This lane: meshing plus the Metal-visualization presentation and misc utilities.

## Step 1 — Re-derive the scope. The list below is already stale.

The previous version said ~2,173 LOC. Measured 2026-08-19 it is **2,345**; `MeasurementHelpers.swift` alone went 225 → 375.

```bash
wc -l Sources/OCCTSwift/{Mesh,MeasurementHelpers,KDTree,PresentationMesh,ZLayerSettings,ClipPlane,Camera,DisplayDrawer,PixMap,FontManager,Date,OCCTSerialQueue}.swift
```

**This lane has the only confirmed detector hits of the four Pass 4 lanes.** Measured 2026-08-19, `detect-duplicate-logic.py --swift` reports six candidate pairs repo-wide and **two are yours**:

| pair | score | shared shingles |
|---|---|---|
| `PresentationMesh.swift:45` / `:120` — both `shadedMesh` | 0.93 | 172 |
| `PresentationMesh.swift:85` / `:158` — both `edgeMesh` | 0.90 | 111 |

Two same-named overloads in one 185-line file, each near-duplicating its sibling. Start there. It is the highest-value known lead in any Pass 4 lane, and the other three lanes return nothing from this detector at all.

## Files

**This exact format is machine-read.** `.claude/workflows/duplication-audit.js` parses this section and expects one backtick-wrapped repo-relative path per line with a LOC count in parentheses. Keep the shape when you update the numbers, or the workflow aborts with "Could not resolve a file scope".

- `Sources/OCCTSwift/Mesh.swift` (723)
- `Sources/OCCTSwift/MeasurementHelpers.swift` (375)
- `Sources/OCCTSwift/PresentationMesh.swift` (185)
- `Sources/OCCTSwift/ZLayerSettings.swift` (184)
- `Sources/OCCTSwift/ClipPlane.swift` (166)
- `Sources/OCCTSwift/Camera.swift` (159)
- `Sources/OCCTSwift/KDTree.swift` (110)
- `Sources/OCCTSwift/PixMap.swift` (108)
- `Sources/OCCTSwift/Date.swift` (106)
- `Sources/OCCTSwift/DisplayDrawer.swift` (102)
[...]

Issue #385, opened 2026-07-26
Pass 4a — Features lane duplication audit


gsdali:
Part of the segmented duplication audit in #377. **Passes 2a (#382), 2b (#383) and 3 (#384) are the prerequisites** — check they are closed before starting.

---

# If you are picking this up cold, start here

Written for someone with no prior context on this repo. Work through it in order.

**What a "pass" is.** #377 splits one repo-wide duplication sweep into thirteen per-layer passes. Each audits a fixed set of files for duplicated logic, files one sub-issue per confirmed finding, fixes them, and closes. This is Pass 4a, the feature-modelling lane: threads, sheet metal, feature recognition and reconstruction, the GD&T write path, sketching.

**What "duplication" means here.** Not just identical text. Four angles: a helper reimplemented under a second name, copy-pasted maths that has drifted, doc comments that no longer describe the code below them, and two parallel types that should be one. In this repo those are frequently **live bugs**. Pass 2a found `Shape.fixed(...)` silently discarding three parameters and two tolerance defaults that had drifted apart for the same question; Pass 3 found six call sites that had lost a null guard their siblings kept.

**Behaviour fixes are in scope.** If a duplication has a real defect on one side, fix it here. See #377's "Scope: behavior fixes are in scope".

---

# Step 0 — Read the policies before you touch anything

Working rules live in **[`okf/policies/`](../tree/main/okf/policies)**. Read these five now:

| policy | why it binds you here |
|---|---|
| [`measure-dont-assume.md`](../blob/main/okf/policies/measure-dont-assume.md) | This repo's most-repeated failure is a census built by grep that was wrong. Do not trust any count below, including the file list. |
| [`search-before-building.md`](../blob/main/okf/policies/search-before-building.md) | The point of the pass. Before adding a helper, find the one that exists. |
| [`issue-tracking.md`](../blob/main/okf/policies/issue-tracking.md) | Every issue needs `type:*` and `priority:*`. CI comments if you forget. Pass 3 filed three issues with **zero** labels because the author wrote `**Type:** chore` in the body prose instead of applying the label. |
| [`changelog-on-merge.md`](../blob/main/okf/policies/changelog-on-merge.md) | Your PR carries its CHANGELOG entry **in the body, never in the diff**. |
| [`prove-the-test-fails.md`](../blob/main/okf/policies/prove-the-test-fails.md) | Every new test is run once with its subject broken, and you report that it failed. |

[...]

PR #961, opened 2026-08-19
Pass 3: XDE/Assembly duplication audit (closes #384)

### LINKED ISSUE #384: Pass 3 — Document/XDE assembly duplication audit
Part of the segmented duplication audit in #377. **Pass 2a (#382) and Pass 2b (#383) are both closed**, so this pass is unblocked.

---

# If you are picking this up cold, start here

This issue is written for someone with no prior context on this repo. Work through it in order. Nothing below assumes you have seen the earlier passes.

**What a "pass" is.** #377 splits one repo-wide duplication sweep into thirteen per-layer passes. Each pass audits a fixed set of files for duplicated logic, files one sub-issue per confirmed finding, fixes them, and closes. This is Pass 3, the Document/XDE assembly layer.

**What "duplication" means here.** Not just identical text. The four angles the tooling looks for are: a helper reimplemented under a second name, copy-pasted maths that has since drifted, doc comments that no longer describe the code below them, and two parallel types that should be one. In this repo those are frequently **live bugs**, not tidiness: Pass 2a found `Shape.fixed(...)` silently discarding three of its parameters, `FaceFixer.Status` returning wrong values since the type was added, and two tolerance defaults that had drifted apart (`1e-6` against `1e-7`) for the same question.

**Behaviour fixes are in scope.** If a duplication turns out to have a real defect on one side, fix it here rather than filing it away. #377's "Scope: behavior fixes are in scope" section covers why.

---

# Step 0 — Read the policies before you touch anything

The repo's working rules live in **[`okf/policies/`](../tree/main/okf/policies)**. They are short. Read these five before starting, and the rest when the step that binds them arrives:

| policy | why it binds you here |
|---|---|
| [`measure-dont-assume.md`](../blob/main/okf/policies/measure-dont-assume.md) | This repo's single most-repeated failure is a census built by grep that was wrong. Do not trust any count in this issue, including the file list below. |
| [`search-before-building.md`](../blob/main/okf/policies/search-before-building.md) | The whole point of the pass. Before adding a helper, find the one that already exists. |
| [`issue-tracking.md`](../blob/main/okf/policies/issue-tracking.md) | Every issue you create needs a `type:*` and a `priority:*` label. CI comments if you forget. |
| [`changelog-on-merge.md`](../blob/main/okf/policies/changelog-on-merge.md) | Your PR carries its CHANGELOG entry **in the PR body, never in the diff**. Whoever merges transcribes it. |
| [`prove-the-test-fails.md`](../blob/main/okf/policies/prove-the-test-fails.md) | Every new test is run once with its subject broken, and you report that it failed. Adding a test is not the rule; watching it fail is. |

Also read, when you reach the step:

- [`semver-at-release.md`](../blob/main/okf/policies/semver-at-release.md) — **never edit `docs/SEMVER.md` in a PR.** Breaks go in the PR body under `## SemVer impact` and are assembled at release. This has been missed twice (#829, #943) and both times shipped a break table missing an entry.
- [`code-style.md`](../blob/main/okf/policies/code-style.md) — the "if you touch it, you fix it" manifest rule. See the trap section below; it has red-lit `main` before.
- [`docs-current.md`](../blob/main/okf/policies/docs-current.md) — docs ship in the same PR as the change, not in a later sweep.
- [`writing-style.md`](../blob/main/okf/policies/writing-style.md) — applies to issue bodies, PR bodies and comments.
- [`code-structure.md`](../blob/main/okf/policies/code-structure.md) — if a fix makes you want to split a file, this decides whether you should.

`okf/index.md` lists the rest. `CLAUDE.md` at the repo root is the project overview and has a long "Known OCCT Bugs" section worth skimming for the subsystem you land in.

---

# Step 1 — Re-derive the scope. Do not trust the list below.

**The file list in the previous version of this issue was already wrong**, because `Document.swift`'s own breakout (#393, executed as #661/#662/#663) shipped in v2.0.0 and evicted 140 standalone types out of it. The same thing happened to Pass 2a. Measure first:

```bash
wc -l Sources/OCCTSwift/Document*.swift \
      Sources/OCCTSwift/{BillOfMaterials,Material,Color}.swift \
      Sources/OCCTSwift/{AssemblyGraph,AssemblyItemId,AssemblyNode}.swift \
      Sources/OCCTSwift/{GDTInfo,GDTWrite,Annotation}.swift
```

**Measured 2026-08-19** (re-run it; this will drift too).

## Files

**This exact format is machine-read.** `.claude/workflows/duplication-audit.js` parses this section
for its scope and expects one backtick-wrapped repo-relative path per line followed by a LOC count
in parentheses. Keep the shape when you update the numbers, or the workflow aborts with
"Could not resolve a file scope".

- `Sources/OCCTSwift/Document.swift` (2694)
- `Sources/OCCTSwift/Color.swift` (211)
- `Sources/OCCTSwift/Material.swift` (183)
- `Sources/OCCTSwift/BillOfMaterials.swift` (172)
- `Sources/OCCTSwift/DocumentError.swift` (18)

`Document.swift` was 16,167 lines when this issue was written. It is now 2,694. **The old "does this need splitting into two agent runs?" question is dead** — the whole lane is now smaller than a single Pass 2a file group.

**Decide explicitly which of these belong to this lane**, and say so in the pass's own PR: `AssemblyGraph.swift`, `AssemblyItemId.swift`, `AssemblyNode.swift`, `GDTInfo.swift`, `GDTWrite.swift`, `Annotation.swift`, `BRepGraph+Attributes.swift`. They are XDE/assembly-shaped and were evicted from `Document.swift`, but no pass has claimed them. **Pass 2a left three Wire files unclaimed in exactly this way and nobody noticed for weeks** — see #382's closing comment. Claim them or explicitly hand them to a Pass 4 lane; do not leave the question open.

Also sweep the bridge side for the same lane: `Sources/OCCTBridge/src/OCCTBridge_Document.mm` and `OCCTBridge_XCAF.mm` if present.

---

# Step 2 — Branch

Cut a fresh branch **from `origin/main`**, not from local `main` (which may be stale) and not from `refactor/377-segmented-audit` (dead since 2026-07-29):

```bash
git fetch origin
git checkout -b refactor/384-pass3 origin/main
```

---

# Step 3 — Run the duplication detector

A committed clone-detector already exists. **Do not write a new one and do not grep by hand.**

```bash
python3 Scripts/repro/784-duplication-rescan/detect-duplicate-logic.py --self-test   # prove it still detects
python3 Scripts/repro/784-duplication-rescan/detect-duplicate-logic.py --swift       # Swift API scope
python3 Scripts/repro/784-duplication-rescan/detect-duplicate-logic.py --bridge      # C++ bridge scope
```

`--self-test` first, every time. Three gate scripts in this repo were confidently wrong because their detector had gone blind while still reporting "all clear" (#618, #624/#630, #626). A detector that reports nothing because it is broken looks identical to one reporting nothing because the code is clean.

Its `README.md` in the same directory explains the shingling and scoring. Findings are ranked; a score of 1.00 means the two bodies are effectively identical.

**The script is a starting point, not the answer.** It finds near-duplicate *bodies*. It cannot find a doc comment that lies about its implementation, or two types that should be one. That is what Step 4 is for.

---

# Step 4 — Run the audit workflow

There is a purpose-built multi-agent workflow at **`.claude/workflows/duplication-audit.js`**. It reads this issue's own `## Files` section for its scope, bin-packs the files into finder-sized groups, and runs four duplication-specific angles across them with an adversarial verify stage.

Invoke it with this issue number as its argument (`384`). It is *not* the generic code-review workflow — that one reads a diff, and this pass has no diff.

**It will not find everything, and it produces false positives.** Nobody has measured this workflow's false-positive rate, so do not assume one; what *is* measured is the sibling over-coverage detector built in #928, which ran at **41% false over a 40-row hand-adjudicated sample**. Expect the same order. Treat every candidate as something to adjudicate against the real code, never as a finding list to file wholesale.

---

# Step 5 — Triage into sub-issues

**One GitHub sub-issue per confirmed finding.** Pass 2a produced 26, Pass 2b produced 13. That is the expected order of magnitude; a pass that files three has probably under-audited, and one that files eighty has not adjudicated.

`.claude/workflows/duplication-triage.js` drafts the bodies. Pass it `{parentIssue: 384, findings: [...]}`. **It deliberately does not create the issues** — you file each one yourself so every creation is auditable. Each draft carries the duplication site with file:line, the observed divergence, the test coverage found at each location, and the specific OCCT C++ class each side calls.

### Creating and labelling

Every issue needs a `type:*` **and** a `priority:*` label, plus `refactor` and `phase:3` so the project board picks it up:

```bash
gh issue create \
  --title "<specific, standalone title — not the auditor's summary verbatim>" \
  --body-file draft.md \
  --label "type:chore,priority:P3,refactor,phase:3"
```

Choosing the labels:

- **`type:bug`** if the duplication has already diverged into wrong behaviour. **`type:chore`** if the copies still agree and this is drift prevention. Pass 2b's #879 was `type:bug`, its #886 `type:chore` — read both for the distinction.
- **`priority:P2`** for a divergence a caller can observe today. **`priority:P3`** for consistent copies. `P1` only if it is a crash or data loss.

Then link each as a sub-issue of this one. Note **`-F`, not `-f`** — the field is numeric and `-f` fails with a confusing 422:

```bash
ID=$(gh api repos/SecondMouseAU/OCCTSwift/issues/<new-number> --jq '.id')
gh api -X POST repos/SecondMouseAU/OCCTSwift/issues/384/sub_issues -F sub_issue_id=$ID
```

Verify: `gh api repos/SecondMouseAU/OCCTSwift/issues/384/sub_issues --jq '.[] | "\(.number) \(.state)"'`

### Project board

This initiative has its own board, [OCCTSwift Refactor (#377)](https://github.com/orgs/SecondMouseAU/projects/2), with a real workflow (`Backlog`, `In Progress`, `PR Review`, `Code-Review`, `Ready`, `Done`). Move cards **at the moment the workflow already visits them** — opening a PR moves it to `PR Review` — not in a later sweep. `issue-tracking.md` has the full rule.

---

# Step 6 — Fix

One PR per sub-issue, or one per tight cluster of related sub-issues. Pass 2a clustered several (`fix/833-839-840-841-844-796-topology-cluster`); that is fine when the fixes touch the same file.

Each PR needs, per policy:

- `Closes #<n>` — **repeat the keyword per issue**; `Closes #1, #2` only closes the first. Several issues in this repo stayed open after their fix merged because the PR title said `fix(#N)` with no keyword at all.
- A `## CHANGELOG entry` section **in the body**, not in `docs/CHANGELOG.md`.
- A `## SemVer impact` section. If your fix breaks a caller, say so here and **do not touch `docs/SEMVER.md`**.
- Tests, each proven to fail against the unfixed code, with that reported.

Before pushing: `swift build`, `swift test`, and every gate script. `gate-scripts` is the **one required check on `main`** — a red one blocks the merge.

```bash
python3 Scripts/check-bridge-index.py
python3 Scripts/check-null-handle-guards.py
python3 Scripts/check-docs-defaults.py
python3 Scripts/check-docs-existence.py
python3 Scripts/derive-bridge-header-split.py --verify
python3 Scripts/count-operations.py
python3 Scripts/census-unmeasured-values.py
python3 Scripts/census-doc-occt-attribution.py
python3 Scripts/check-changelog-transcription.py
python3 Scripts/check-style-manifest.py --base origin/main
```

CLAUDE.md's "Static Gate Scripts" section documents each. There is an opt-in pre-commit hook at `Scripts/git-hooks/pre-commit`.

---

# Step 7 — Close the pass

Close this issue when every sub-issue is closed **and** you have answered the Step 1 scope question in writing. Post a closing comment recording:

- which files you claimed and which you handed on,
- anything you found and deliberately did **not** fix, with the reason,
- any finding that turned out not to be real, so the next pass does not re-investigate it.

Pass 2a's closing comment is the model.

---

# Traps this repo has actually hit

Every one of these has cost real time here. They are not hypothetical.

**The style manifest is a ratchet.** Files listed in `Scripts/style-manifest-{swift,bridge}.txt` are exempt from formatting checks — but the moment your PR *touches* one, you must bring it fully clean and remove it from the manifest in that same PR. You cannot add a file back: `check-style-manifest.py` refuses. Removing a file from the manifest without actually making it lint-clean just moves the failure to the linter, which is how `main` sat red for five merges (#942).

**Where you put the extracted helper is a correctness decision, not a style one.** When you factor
duplication into a shared helper, decide its placement by **reach**: count every site with that logic,
across every file, before choosing. A `static` helper in a `.mm` is confined to that one translation
unit, so any copy living in another `.mm` cannot converge on it and drifts independently from then on.

This is not theoretical. `occtComputeBoundingBox` was file-static in `OCCTBridge_Topology.mm`, and that
is exactly **why** `OCCTShapeGetBounds` over in `OCCTBridge_Properties.mm` was the one bounds entry
point with no `IsVoid()` guard — a defect that shipped, and was fixed on 2026-08-17 (#943) by moving
the helper to `Sources/OCCTBridge/src/OCCTBridge_Internal.h` as `inline`. It then happened again inside
this very pass: #949's `occtDocumentInit` is file-static in `OCCTBridge_Document.mm` while twelve copies
of the same boilerplate sit in `OCCTBridge_IO.mm`, six of which have already lost the null guard their
siblings kept. That is #957.

**So before extracting anything:** `grep -rn '<the distinctive call>' Sources/OCCTBridge/src/*.mm`. If it
appears in more than one file, the helper goes in `OCCTBridge_Internal.h` as `inline`. If the sites are
not identical — different wrapper types, different subsets of the work — that difference is itself worth
recording, because a divergence between copies is usually where the bug is.

**`OCCTBridge_Modeling.mm` is huge and format-sensitive.** Check whether it is still on the bridge manifest before touching it.

**A green test suite is not coverage.** Some tests are gated on `OCCTSWIFT_LOCAL=1` and silently skip in CI. The suite reports the same total whether a gated test ran or was skipped — only the per-test line distinguishes them. If you rely on a test to prove something, confirm from the log that it *started*.

**`swift-format` diagnostics strip inline code spans.** A doc comment reading ``Split `self` by `tool`.`` is quoted back as `"Split  by  ."`. The doubled spaces are an artefact of the message, not damage to the file. Someone lost time filing that as a defect (#942).

**Building against the released kernel gives false results.** If you change anything in `Sources/OCCTBridge/`, build from source. In a git worktree, `Libraries/` is its own tree — symlink it from the main checkout and set `OCCTSWIFT_LOCAL=1`, or you are testing the shipped binary rather than your change.

**Branch off `origin/main`, and refetch before transcribing anything.** Two branches adding a CHANGELOG entry both insert at the top of the same section; if your base is stale you get a conflict whose resolution silently drops the other side's work.

---

## Ordering

**Depends on:** Pass 2a (#382) and Pass 2b (#383). Both closed, so this is unblocked.

**Feeds:** all four Pass 4 lanes (#385–#388) — features, drawing, export and mesh all touch Document. Fix or triage findings here before those start.

**Companion:** #810 is the refman-coverage audit for this same lane and runs *after* this pass, because the duplication work changes what there is to compare.

SMKiloBOT:
## CHANGELOG entry

**Pass 3: XDE/Assembly duplication audit**

- **Refactor:** Extracted shared document initialization boilerplate into `occtDocumentInit()` helper (used by `OCCTDocumentCreate`, `OCCTDocumentLoadSTEP`, and 12 sites in `OCCTBridge_IO.mm`). Eliminates ~30 lines of duplicated code per pair, adds missing null-check guard to 6 sites in `IO.mm`.

- **Refactor:** Templated naming trace pair (`OCCTDocumentNamingTraceForward` / `OCCTDocumentNamingTraceBackward`) on iterator type (`TNaming_NewShapeIterator` / `TNaming_OldShapeIterator`). Eliminates ~30 lines of duplicated boilerplate.

- **Refactor:** Templated format enumeration pair (`OCCTDocumentReadingFormats` / `OCCTDocumentWritingFormats`) on member function pointer. Eliminates ~25 lines of duplicated boilerplate.

- **Refactor:** Unified Swift-side format enumeration (`readingFormats` / `writingFormats`) and naming trace (`tracedForward` / `tracedBackward`) accessors in `Document.swift` behind private helpers parameterized by the bridge call. Each pair was byte-identical apart from the bridge symbol.

- **Fix:** Moved `occtDocumentInit` to `OCCTBridge_Internal.h` as `inline` so it can be used across translation units. All 14 document-creation sites now route through the shared helper. Added missing null-check guard to 6 sites in `IO.mm` that previously dereferenced `doc->Main()` without checking `IsNull()`.

- **Internal:** Probe confirmed `TDocStd_Application::NewDocument("MDTV-XCAF", doc)` does not return null on the pinned OCCT 8.0.1 kernel — the six previously unguarded `IO.mm` sites were not on a latent crash branch.

- **Style:** Fixed all 91 swift-format doc comment violations in `Document.swift` (periods, blank lines, parameter syntax, missing Returns sections). Removed `Document.swift` from style manifest (now compliant).

## SemVer impact

No breaking changes. All changes are internal refactors — public C API and Swift API surfaces unchanged. The behavioral change in #957 (clean `nullptr` return instead of potential crash if `NewDocument` ever failed) is a safety improvement with no API change.

**Closes #384** (when merged to main, sub-issues #949, #950, #951, #952, #957 will close automatically)


gsdali (changes requested):
The pass itself is good work: five findings, five fixes, a real bug shape investigated properly in #957, the helper-placement rule surfaced and now written into #385 to #388, and a closing audit that answers the scope question. The revised audit on #384 in particular is a proper job — angles 3 and 4 actually run, §4 corrected to "merged to integration, closes on the `main` merge", #957's labels fixed, and the `NewDocument` non-finding recorded so nobody re-derives it.

Three things to settle before this goes to `main`, one of them a defect that ships if it merges as-is.

## 1. The 16 malformed parameter docs are still here

PR #959 carried two `CHANGES_REQUESTED` reviews asking for these, and merged into the integration branch without them. They are now in this PR, on their way to `main`:

```
  33  /// - Parameters: url: URL to the STEP file
 178  /// - Parameters: url: Output file URL
 256  /// - Parameters: parent: Parent node (nil for document root)
 392  /// - Parameters: node: The label containing the selection
      … 12 more
```

Counted on this branch's `Document.swift`: **16 of this hybrid, 26 correct nested forms, 0 correct singular forms.**

`- Parameters:` opens a list; entries belong on following indented lines. Written inline it renders as a plain bullet, so those sixteen symbols lose their parameter documentation in Quick Help and on the docs site. `swift-format` accepts it, which is why `code-style` is green and will stay green.

That is the sharp edge here: this PR's own CHANGELOG says it "Fixed all 91 swift-format doc comment violations in `Document.swift` … Removed `Document.swift` from style manifest (now compliant)". Compliant by the linter, yes. But the file leaves the manifest in a state where sixteen of its parameter docs do not work, and once it is off the manifest nothing will flag them again.

Either form is a one-line change per site:

```swift
/// - Parameter url: URL to the STEP file.
```

`load(from url: URL, progress: ImportProgress? = nil)` at line 33 also documents only `url`; `progress` is undocumented.

## 2. The CHANGELOG entry contradicts itself, and I had already fixed this wording

Three bullets, in order:

> - **Refactor:** … adds missing null-check guard to 6 sites in `IO.mm`.
> - **Fix:** … Added missing null-check guard to 6 sites in `IO.mm` that previously dereferenced `doc->Main()` without checking `IsNull()`.
> - **Internal:** Probe confirmed … the six previously unguarded `IO.mm` sites were **not** on a latent crash branch.

The first two describe a defect being fixed. The third says there was no defect. The third is the measured one — `Scripts/repro/957-newdocument-null/probe-output.txt`, which I ran and committed on the #958 branch precisely so this claim would rest on evidence.

This exact wording was corrected in #958's body before it merged, to:

> Six sites in `IO.mm` proceeded without the `IsNull()` check their eight siblings performed; all fourteen now share one guarded path. **This is consistency, not a crash fix** — the probe below shows the unguarded branch was not reachable on this kernel.

The aggregated entry has reverted to the pre-correction framing. Since this is the entry that lands in the release record, it is the version that matters.

Two smaller things in the same entry: the first and fifth bullets are near-duplicates of each other, and `## SemVer impact` repeats the disproved claim ("potential crash if `NewDocument` ever failed"). Nothing here is consumer-visible now, so the entry can simply say so — Pass 2a's precedent groups by consumer impact, and this pass's honest grouping is "internal only, zero behaviour change" throughout.

## 3. ~230 deferred doc violations need an issue, not a sentence

§7 of the closing audit defers roughly 230 missing-period summary violations across the eleven declared files to "a dedicated doc cleanup pass or PR". That is a reasonable call — it is genuinely a different job from a duplication audit.

But there is no issue for it. Searching open issues for it returns only #384 itself, and #384 is about to close. **Deferred work with no tracker is how the three Wire files from Pass 2a went unnoticed for weeks**, and it is the same shape as the unclaimed-files problem this pass just fixed properly in §2.

File it before this merges, so the sentence in the closing comment points somewhere.

## What I am not asking for

The refactors are right. `occtDocumentInit` in `OCCTBridge_Internal.h` as `inline` with two overloads, the templated iterator pair, the member-function-pointer format pair, and the Swift-side `documentFormats(via:)` are all good, and #955/#956 keeping their helpers `static` was the correct call rather than an oversight — that logic exists in one TU only. The 16,323-line churn in `OCCTBridge_Document.mm` is the clang-format pass discharging the manifest obligation properly, not logic.

Fix the sixteen, restore the corrected CHANGELOG wording, file the doc-cleanup issue, and this closes the pass cleanly.

gsdali (changes requested):
Confirming the disposition: **all three items in my previous review need fixing in this PR. None of them is optional, and the doc violations are not to be deferred.**

## The doc violations get fixed here, not filed

The closing audit's §7 proposed deferring them to "a dedicated doc cleanup pass or PR". That is off the table — fix them in this PR.

**And the number in §7 is wrong.** It says ~230. Measured just now with `swift-format lint --strict --configuration .swift-format` on this branch, file by file:

| file | violations | on manifest |
|---|---|---|
| `Annotation.swift` | 83 | yes |
| `AssemblyNode.swift` | 83 | yes |
| `Color.swift` | 66 | yes |
| `BillOfMaterials.swift` | 58 | yes |
| `Material.swift` | 44 | yes |
| `GDTWrite.swift` | 22 | yes |
| `GDTInfo.swift` | 11 | yes |
| `AssemblyItemId.swift` | 2 | yes |
| `DocumentError.swift` | 2 | yes |
| `AssemblyGraph.swift` | 1 | yes |
| **total** | **372** | |

`Document.swift` is already at 0 and off the manifest, so it is done.

372, not ~230. Worth noting how the estimate went wrong: §7 characterised them as "one-line summaries missing terminating periods", which is one rule. The real set spans that plus blank-line-after-summary, malformed parameter lists and missing `Returns:` sections. Do not work from the estimate — work from what the linter prints per file.

## Both halves, for each file

Every one of these ten is currently **on** `Scripts/style-manifest-swift.txt`, so there is no "if you touch it, you fix it" obligation today; they are legitimately grandfathered. Fixing them is a deliberate scope decision, and it means doing **both** halves per file:

1. Bring the file lint-clean.
2. Remove its line from `Scripts/style-manifest-swift.txt`.

Doing only (2) is #942, which kept `main` red for five merges. Doing only (1) leaves the file exempt, so nothing holds it clean afterwards and the work rots. And the manifest is a one-way ratchet: `check-style-manifest.py` refuses re-additions, so once a line is removed there is no going back.

## The risk to watch, because it already happened once at smaller scale

#959 fixed 91 violations in `Document.swift` and introduced **16 malformed parameter docs** doing it — `- Parameters: url: …` on one line, which `swift-format` accepts and DocC does not render. That is a 17% defect rate on the fix itself, and it is still unfixed in this PR (item 1 of my previous review).

You are now about to do **four times that volume**. Please:

- Use `- Parameter x:` for a single parameter, or `- Parameters:` followed by indented `- x:` lines for several. Never `- Parameters: x:` inline.
- Document every parameter in the signature, not just the first. `load(from url: URL, progress: ImportProgress? = nil)` at `Document.swift:33` documents only `url`.
- Where `ValidateDocumentationComments` reports a drifted parameter list or a missing `Returns:`, read the signature and describe what it actually does. Do not invent text to satisfy the rule. #945 took `wedge`'s parameter docs from OCCT's own `BRepPrimAPI_MakeWedge.hxx` rather than guessing, which is the standard here.
- After the pass, grep for the bad shape as a self-check: `grep -rnE '/// - Parameters: [a-zA-Z_]+:' Sources/OCCTSwift/` should return nothing.

Also expect the diagnostic quirk: `swift-format` strips inline code spans, so a good summary reading ``Split `self` by `tool`.`` is quoted back as `"Split  by  ."`. The doubled spaces are the message, not the file. Someone already lost time filing that as damage during #942.

## Recap of what this PR needs

1. **The 16 malformed parameter docs in `Document.swift`** corrected (lines 33, 178, 256, 392, plus 12 more).
2. **The CHANGELOG entry** restored to the corrected wording — "consistency, not a crash fix" — with the duplicate bullet dropped and `## SemVer impact` stopping repeating the disproved "potential crash" claim. Everything in this pass is internal-only; the entry can say so plainly.
3. **All 372 doc violations fixed**, and those ten files removed from the manifest.

No issue to file for item 3 any more, since it is being done here.

The refactors themselves are unchanged and still good. This is all documentation and record-keeping, and it is the last thing standing between Pass 3 and `main`.

Issue #943, opened 2026-08-17
Breaking: Shape.bounds becomes Optional, so a void shape stops fabricating (0,0,0)-(0,0,0) (follow-up to #834)


gsdali:
Follow-up to #834, which was **closed on the partial fix**. This is the part that was deferred.

## What is still wrong

`Shape.bounds` fabricates `(0,0,0)-(0,0,0)` for a void shape, indistinguishable from a genuine zero-size shape at the origin:

```swift
// Sources/OCCTSwift/Shape.swift:2274
public var bounds: (min: SIMD3<Double>, max: SIMD3<Double>)

// Sources/OCCTSwift/Shape+Analysis.swift:2252  — same Bnd_Box, same BRepBndLib::Add
public var boundingBox: (min: SIMD3<Double>, max: SIMD3<Double>)?
```

The two compute the identical `Bnd_Box` and disagree only on this case, because `bounds` is a non-optional tuple with no way to express "no geometry".

## What #834 actually delivered, and what it did not

#834 closed `COMPLETED` on 2026-08-12 via #870. It shipped:

- a bridge-side `IsVoid()` guard (`OCCTBridge_Properties.mm:2067`) which, on a void box, deliberately writes `0` to all six outputs — that makes the fabrication *intentional*, not absent;
- a `- Warning:` doc comment on `bounds`, `size` and `center` describing the defect in full.

It explicitly did **not** change the signature. The CHANGELOG entry says so: *"the full fix (`bounds` becoming Optional) is a breaking change to 20 call sites and is documented as a candidate follow-up, not executed here."*

So today the defect is documented, deliberate, and untracked by any open issue. This issue is the tracker.

## Blast radius is three properties, not one

`size` and `center` both derive from `bounds` and inherit the fabrication. Their own doc comments already admit it: `.zero` "is a fabricated answer, not a measurement".

```swift
Sources/OCCTSwift/Shape.swift:2274  public var bounds:  (min: SIMD3<Double>, max: SIMD3<Double>)
Sources/OCCTSwift/Shape.swift:2285  public var size:    SIMD3<Double>
Sources/OCCTSwift/Shape.swift:2294  public var center:  SIMD3<Double>
```

Whether `size`/`center` also become optional, or keep returning a non-optional with a documented sentinel, is the design decision this issue has to make. Making only `bounds` optional and leaving the other two is defensible, but it should be a choice rather than a side effect.

## Call-site counts, and why they are upper bounds

[...]

Issue #761, opened 2026-08-07
AAG's hand-rolled face/edge adjacency duplicates BRepGraph's outerWire(of:)/sharedEdges(between:and:)


gsdali:
## What

`AAG` (`Sources/OCCTSwift/FeatureRecognition.swift`) maintains its own hand-rolled
face/edge-adjacency machinery, built pairwise over `OCCTFacesAreAdjacent`/
`OCCTFaceGetSharedEdges`/`OCCTEdgeGetConvexity`, entirely independent of
`Sources/OCCTSwift/BRepGraph.swift`, which already exposes the same two questions as
first-class operations over a `BRepGraph_Tool`-backed graph: `outerWire(of:)` and
`sharedEdges(between:and:)`.

`PocketFeature.isOpen`'s enclosure test (#735, then reworked in PR #753's review round) is
a second, independent example of the same pattern: it answers "which faces border this
edge" via `Edge.adjacentFaces(in:)` plus a hand-rolled `facesAreSame` (a hand-rolled
`TopoDS_Shape::IsSame` bridge through `OCCTShapeFromFace`/`OCCTShapeIsSame`), rather than
through `BRepGraph`'s own adjacency model.

## Why this is a real risk, not just style

`AAG` and `BRepGraph` now have two independent implementations of "which edges bound this
face" and "which edges are shared between these two faces." That has already drifted twice
in ways that mattered:

- #642: `AAG`'s node identity (`faces()` vs `orientedFaces()`) needed its own fix once a
  compound shared a face between two solids; nothing in `BRepGraph`'s model needed touching
  for the equivalent case, because it does not share the assumption.
- #699: `AAG`'s adjacency needed retrofitting with solid-scoping (`solidGroups`) so two
  occurrences in different solids stop being treated as adjacent; again, a fix local to
  `AAG`'s own pairwise probing, invisible from `BRepGraph`'s side.

A correctness fix to one is not guaranteed to reach the other, because there is no shared
implementation between them, only a shared *question*. PR #753's own review flagged this
directly against the enclosure-test code (see "Altitude" finding in the review), and the
adjudication on that PR confirmed it is real but out of scope for a bug-fix PR:

> The `BRepGraph` duplication half of that finding stands and is worth an issue, but not
> this PR.

## Scope for this issue

Investigate whether `AAG` can be built on top of `BRepGraph`'s adjacency primitives instead
[...]

PR #980, opened 2026-08-20
fix(#964): report the assembly-count bound instead of returning it as the count

### LINKED ISSUE #964: Document.assemblyItemCount silently caps at 100,001 and returns the cap as the count
Found while writing the `- Returns:` section that `swift-format`'s `ValidateDocumentationComments` demanded for `Document.assemblyItemCount(maxDepth:)`, during the Pass 3 follow-up work.

## What

`OCCTDocumentAssemblyItemCount` (`Sources/OCCTBridge/src/OCCTBridge_Document.mm`) stops counting at 100,001 and returns that number as if it were the real total:

```cpp
int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef doc, int32_t maxDepth)
{
  if (!doc || doc->doc.IsNull())
    return 0;
  try
  {
    int level = (maxDepth <= 0) ? INT_MAX : maxDepth;
    XCAFDoc_AssemblyIterator iter(doc->doc, level);
    int count = 0;
    while (iter.More())
    {
      count++;
      iter.Next();
      if (count > 100000)
        break;
    }
    return count;          // <- 100001 for any document with more items
  }
  catch (...) { return 0; }
}
```

A document with 100,000 assembly items and a document with ten million both report **100001**, and the caller has no way to tell those apart, or to tell either from a document that genuinely has 100,001.

## Why this is a defect rather than a sensible guard

This is the shape #726's census exists to find: **a value returned as a measurement that was never measured.** It is also the shape that produced #761 (a silent buffer cap) and PR #768 (a dropped alpha channel) — the two cases #377's whole duplication programme cites as its motivating examples.

A bound on an unbounded walk is reasonable. Reporting the bound as the answer is not. The caller asked "how many items are there" and receives a number that may be a count or may be a ceiling, with nothing distinguishing them.

`Document.assemblyItemCount(maxDepth:)` forwards it directly, so the Swift API has the same property.

## Options

1. **Return an optional or a result type** — `nil` / `.exceededLimit` when the cap is hit, so "too many to count" is expressible. Breaking, and this is a fresh major cycle, so the timing is good.
2. **Raise or remove the cap.** `XCAFDoc_AssemblyIterator` is finite over a finite document; the cap appears to be defensive against a cyclic assembly rather than a real size limit. If cycles are the concern, detect the cycle rather than bounding the count.
3. **Keep the cap and report it** via a second out-parameter or a companion `assemblyItemCountWasCapped` property.

Whichever is chosen, the number needs to stop being ambiguous. Option 2 is worth measuring first: if the cap exists for cycles, and `XCAFDoc_AssemblyIterator` cannot cycle, it can simply go.

## Interim

`Document.swift`'s doc comment now carries a `- Warning:` naming this issue, so a caller reading the API sees it. That is documentation of a defect, not a fix.

## Not in Pass 3's scope

Pass 3 is a duplication audit; this is a single-site correctness defect that surfaced while satisfying a doc-comment linter rule. Filing separately rather than widening that pass. It belongs to the same lane, so #810 (the refman-coverage audit for Document/XDE) may also want to know about it.

gsdali:
## What & why

`OCCTDocumentAssemblyItemCount` stopped at 100,001 and returned that number as though it were the total:

```cpp
if (count > 100000) break;
return count;
```

A document with 100,000 items, one with ten million, and one that genuinely has 100,001 all reported the same number, with nothing to tell them apart. That is the shape #726's census exists to find, and the same class as #761's buffer cap and PR #768's dropped alpha channel.

## The bound stays, and that is deliberate

`XCAFDoc_AssemblyIterator` has `myMaxLevel` and a `myFringe` work list but **no visited set**, so it is a depth-bounded tree walk rather than a cycle-safe graph walk. Our bridge passes `INT_MAX` when `maxDepth <= 0`, which is the default. A malformed self-referencing assembly would therefore iterate essentially forever. **Removing the bound trades a wrong number for a hang.**

Cycle detection was considered and rejected on the structure rather than on effort: `XCAFDoc_AssemblyItemId` is a **path**, so a cycle generates infinitely many distinct ids and path-dedup never terminates; label-dedup would terminate but undercount, because a label legitimately appearing many times is exactly what an assembly instance *is*.

## What changes

Hitting the bound is now reported instead of returned.

```cpp
int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef doc, int32_t maxDepth, bool* outTruncated);
```

```swift
public func assemblyItemCount(maxDepth: Int = 0) -> Int?   // nil == "a floor, not a count"
```

No sentinel value, because a sentinel is the defect being fixed rather than relocated.

## Tests, proven to fail first

Both were run against an injected always-truncated bridge before the fix:

```
✘ iterateAssembly()              Expectation failed: (count → nil) != nil
✘ smallAssemblyCountIsComplete() Expectation failed: (count → nil) != nil
```

`smallAssemblyCountIsComplete` **could not have been written before this change** — "counted 100,001" and "gave up at 100,001" were the same value, so there was nothing to assert.

## Relationship to #968

#968 also targeted #964 and is closed as stale: it was branched before #958, #966 and #969 landed and carried a commit reverting the Pass 3 reapply, which would have taken `occtDocumentNamingTraceImpl`, `occtDocumentFormatsImpl` and `occtDocumentInit` from present on `main` to absent. Its chosen fix — remove the cap, add cycle detection — is the one rejected above.

## Also updated

- `docs/reference/Document-OCAF-Attributes.md`: signature, a `Returns:` line, and an example that unwraps.
- `docs/CHANGELOG.md`: #966's line added a `- Warning:` describing this defect. It is superseded within the same unreleased section, so it now says so rather than describing something fixed two entries below it.

## Verification

**5635 tests / 1464 suites / 0 failures** (`OCCTSWIFT_LOCAL=1`, local kernel). All ten gate scripts exit 0, including `check-borrowed-handles` and `check-style-manifest --base origin/main`.

### `Document.assemblyItemCount(maxDepth:)` reports when it could not finish counting (#964)

The bridge stops walking an assembly at 100,000 items, because `XCAFDoc_AssemblyIterator` keeps no visited set and a malformed self-referencing assembly would otherwise iterate to `INT_MAX` depth. It used to return the bound as though it were the total, so a document with more items was indistinguishable from one that genuinely had that many.

The bound remains — removing it trades a wrong answer for a hang — but it is now reported. `assemblyItemCount(maxDepth:)` returns `Int?`, and `nil` means the walk was truncated and the number would be a floor rather than a count. Migration: unwrap, `if let n = doc.assemblyItemCount() { … }`.

## SemVer impact

**MAJOR.** `Document.assemblyItemCount(maxDepth:)` changes from `Int` to `Int?`, which is a compile error for every caller. Migration is to unwrap or coalesce. This is the same shape as #943's six bounding-box accessors, and for the same reason: a value that could not be distinguished from a real measurement.

Row for `docs/SEMVER.md`'s next major table:

| Break | Kind | Detail |
|---|---|---|
| `Document.assemblyItemCount(maxDepth:)` returns `Int?` | compile error | #964 |

Closes #964

### DIFF
--- a/Sources/OCCTBridge/include/OCCTBridge_Document.h
+++ b/Sources/OCCTBridge/include/OCCTBridge_Document.h
@@ -2310,8 +2310,8 @@ int32_t OCCTDocumentPresentationGetMode(OCCTDocumentRef _Nonnull doc, int64_t la
 /// `outTruncated` is set true when that bound is reached, in which case the return value is a
 /// floor rather than a count (#964). Pass NULL if you do not care.
 int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef _Nonnull doc,
-                                      int32_t                  maxDepth,
-                                      bool* _Nullable          outTruncated);
+                                      int32_t maxDepth,
+                                      bool* _Nullable outTruncated);
 
 // MARK: - XCAFDoc_DimTol (v0.90.0)
 
--- a/Sources/OCCTBridge/include/OCCTBridge_Document.h
+++ b/Sources/OCCTBridge/include/OCCTBridge_Document.h
@@ -2040,7 +2040,15 @@ int32_t OCCTDocumentPresentationGetMode(OCCTDocumentRef _Nonnull doc, int64_t la
 
 /// Count the number of assembly items in a document.
 /// @param maxDepth Maximum depth to traverse (0 = unlimited)
-int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef _Nonnull doc, int32_t maxDepth);
+/// Count assembly items, depth-first, to `maxDepth` levels (0 = unlimited).
+///
+/// The walk is bounded at 100,000 items because `XCAFDoc_AssemblyIterator` keeps no visited
+/// set, so a malformed self-referencing assembly would otherwise iterate to `INT_MAX` depth.
+/// `outTruncated` is set true when that bound is reached, in which case the return value is a
+/// floor rather than a count (#964). Pass NULL if you do not care.
+int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef _Nonnull doc,
+                                      int32_t                  maxDepth,
+                                      bool* _Nullable          outTruncated);
 
 // MARK: - XCAFDoc_DimTol (v0.90.0)
 
--- a/Sources/OCCTBridge/src/OCCTBridge_Document.mm
+++ b/Sources/OCCTBridge/src/OCCTBridge_Document.mm
@@ -10316,8 +10316,14 @@ int32_t OCCTDocumentPresentationGetMode(OCCTDocumentRef doc, int64_t labelId)
 
 #include <XCAFDoc_AssemblyIterator.hxx>
 
-int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef doc, int32_t maxDepth)
+// #964: the walk's upper bound. Reaching it means the count is a floor, not a measurement,
+// which `outTruncated` reports so a caller can tell the two apart.
+static const int kAssemblyItemCountLimit = 100000;
+
+int32_t OCCTDocumentAssemblyItemCount(OCCTDocumentRef doc, int32_t maxDepth, bool* outTruncated)
 {
+  if (outTruncated)
+    *outTruncated = false;
   if (!doc || doc->doc.IsNull())
     return 0;
   try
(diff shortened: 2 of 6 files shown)


gsdali:
Your `code-style` check is red for the manifest rule, not for anything wrong with this diff: it
edits `Sources/OCCTBridge/include/OCCTBridge_Document.h` (the `OCCTDocumentAssemblyItemCount`
signature, around line 2040) while that file is still listed on `Scripts/style-manifest-bridge.txt`,
and `check-style-manifest.py` requires the entry to be removed in the same PR. Removing it obliges
a full `clang-format` sweep of the file, measured at 1,714 `diff -u` lines.

#971's PR is doing exactly that sweep, for its own two-line comment correction, and takes the file
off the manifest. Once it lands, this PR only has to keep its own hunk `clang-format` clean, which
it already is, rather than carry the sweep itself.

The rebase will conflict at that hunk, since the reformat moves every line number in the file. The
resolution is to re-apply the three-line signature onto the formatted context; nothing semantic is
in the way.

One thing worth fixing while you are there, unrelated to the above: the new comment block leaves two
summary lines stacked on the same declaration.

```c
/// Count the number of assembly items in a document.
/// @param maxDepth Maximum depth to traverse (0 = unlimited)
/// Count assembly items, depth-first, to `maxDepth` levels (0 = unlimited).
```

The first two lines are the old comment and the third is the new summary, so the `@param` now sits
between two summaries and the first sentence is stated twice.

PR #997, opened 2026-08-20
refactor(#972): deduplicate 2D→3D lifting logic in Sketch, SheetMetal, FeatureReconstructor

### LINKED ISSUE #972: 2D→3D lifting logic reimplemented in Sketch, SheetMetal, FeatureReconstructor
## Duplication site

**Primary (worse-drifted) location:** `Sources/OCCTSwift/Sketch.swift:133-135`
```swift
private func lift(_ p2: SIMD2<Double>, with placement: Placement) -> SIMD3<Double> {
    placement.origin + p2.x * placement.xAxis + p2.y * placement.yAxis
}
```

**Duplicate counterparts:**
- `Sources/OCCTSwift/SheetMetal.swift:68-70` — `Flange.worldPoint(_:)`
```swift
fileprivate func worldPoint(_ p: SIMD2<Double>) -> SIMD3<Double> {
    origin + p.x * uAxis + p.y * vAxis
}
```
- `Sources/OCCTSwift/FeatureReconstructor.swift:335` — inline in `applyExtrude`
```swift
let pts3D = e.profilePoints2D.map {
    placement.origin + $0.x * placement.xAxis + $0.y * placement.yAxis
}
```

## Divergence

All three implementations are functionally identical: they map a 2D point `(x, y)` to 3D using an origin and two orthogonal basis vectors (`xAxis`/`yAxis` or `uAxis`/`vAxis`). The only differences are:
- Naming: `lift` vs `worldPoint` vs inline
- Receiver: `Placement` struct vs `Flange` struct vs inline
- Visibility: `private` method vs `fileprivate` method vs inline code

No behavioral divergence — all produce the same 3D point for equivalent inputs.

## Tests

- **Sketch.swift** — No direct unit tests for `lift()`. Coverage via `Sketch.buildProfile(in:graph:)` → `Issue72SketchTests` / `ConstructionContextTests`.
- **SheetMetal.swift** — Covered by `SheetMetalBuilderTests` (flange extrusion path exercises `worldPoint`).
- **FeatureReconstructor.swift** — Covered by `FeatureReconstructorTests` (extrude path exercises inline lifting).

**No dedicated test** exercises the lifting logic in isolation for any of the three.

## OCCT functions

None directly — this is pure Swift SIMD math. The `Placement` struct's xAxis/yAxis are derived from `perpendicularBasis(to:)` in `ConstructionEntity.swift` which uses `OCCTPerpendicularBasis` bridge function.

---

**Concrete maintenance cost:** A bug fix or enhancement to the lifting logic (e.g., adding a tolerance check, handling degenerate bases, or supporting a third basis vector) must be applied in three places. The `Placement` struct in `ConstructionEntity.swift` is the natural home for this helper since both `Sketch` and `FeatureReconstructor` already use `Placement`, and `SheetMetal.Flange` stores equivalent `origin`/`uAxis`/`vAxis` fields.

SMKiloBOT:
## Summary

Addresses issue #972 by consolidating the duplicated 2D→3D lifting logic into a single shared method on the `Placement` struct.

## Changes

1. **`Sources/OCCTSwift/ConstructionEntity.swift`**: Added `lift(_:)` method to `Placement` struct that maps a 2D point to 3D using the placement's origin and basis vectors.

2. **`Sources/OCCTSwift/Sketch.swift`**: Removed private `lift(_:with:)` method; now uses `placement.lift(_:)`.

3. **`Sources/OCCTSwift/SheetMetal.swift`**: 
   - Added `placement: Placement` property to `Flange`
   - Removed `fileprivate func worldPoint(_:)`
   - Updated all call sites to use `flange.placement.lift(_:)`

4. **`Sources/OCCTSwift/FeatureReconstructor.swift`**: Updated `applyExtrude` to use `placement.lift(_:)` instead of inline lifting code.

## Verification

- All Sketch tests pass (6 tests)
- All SheetMetal tests pass (16 tests)
- All FeatureReconstructor tests pass (20 tests)
- Gate scripts pass: `check-bridge-index`, `check-null-handle-guards`, `check-docs-defaults`

## SemVer Impact

No breaking changes. The new `Placement.lift` method is additive API. The removed methods (`Sketch.lift`, `Flange.worldPoint`) were private/fileprivate.

Closes #972 (sub-issue of #385 Pass 4a duplication audit)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 A comprehensive Swift wrapper for [OpenCASCADE Technology (OCCT)](https://www.opencascade.com/) 8.0.1, providing B-Rep solid modeling for macOS and iOS. **v3.0.0 — SemVer-stable; see [SEMVER.md](docs/SEMVER.md#v300) before upgrading from v2.x.**
 
-**4,333 wrapped operations** | macOS 12+ / iOS 15+ (arm64), visionOS and tvOS untested | OCCT 8.0.1
+**4,334 wrapped operations** | macOS 12+ / iOS 15+ (arm64), visionOS and tvOS untested | OCCT 8.0.1
 
 ## Quick Start
 
--- a/Scripts/style-manifest-swift.txt
+++ b/Scripts/style-manifest-swift.txt
@@ -55,7 +55,6 @@ Sources/OCCTSwift/EvolvedSectionInfo.swift
 Sources/OCCTSwift/EvolvingFilletEdge.swift
 Sources/OCCTSwift/Exporter.swift
 Sources/OCCTSwift/ExtremaTypes.swift
-Sources/OCCTSwift/FeatureReconstructor.swift
 Sources/OCCTSwift/FillConstraint.swift
 Sources/OCCTSwift/FilletBuilder.swift
 Sources/OCCTSwift/FillingPoleGrid.swift
@@ -157,8 +156,6 @@ Sources/OCCTSwift/ShapeRayIntersection.swift
 Sources/OCCTSwift/ShapeType.swift
 Sources/OCCTSwift/SharedLibrary.swift
 Sources/OCCTSwift/SheetLayout.swift
-Sources/OCCTSwift/SheetMetal.swift
-Sources/OCCTSwift/Sketch.swift
 Sources/OCCTSwift/StepHeader.swift
 Sources/OCCTSwift/STEPOptions.swift
 Sources/OCCTSwift/SurfaceIntersectionResult.swift
--- a/Sources/OCCTSwift/ConstructionEntity.swift
+++ b/Sources/OCCTSwift/ConstructionEntity.swift
@@ -38,6 +38,14 @@ public struct Placement: Sendable, Hashable {
         let (x, y) = perpendicularBasis(to: z)
         self.init(origin: origin, xAxis: x, yAxis: y, zAxis: z)
     }
+
+    /// Map a 2D point in the placement's local (x, y) plane to 3D world space.
+    ///
+    /// - Parameter p: 2D coordinates where `x` scales `xAxis` and `y` scales `yAxis`.
+    /// - Returns: The corresponding 3D point: `origin + x * xAxis + y * yAxis`.
+    public func lift(_ p: SIMD2<Double>) -> SIMD3<Double> {
+        origin + p.x * xAxis + p.y * yAxis
+    }
 }
 
 /// Fusion 360-style recipe for a construction plane.
--- a/Sources/OCCTSwift/FeatureReconstructor.swift
+++ b/Sources/OCCTSwift/FeatureReconstructor.swift
@@ -331,9 +331,7 @@ public struct FeatureReconstructor: Sendable {
             return
         }
         let placement = Placement(origin: e.planeOrigin, normal: e.planeNormal)
-        let pts3D = e.profilePoints2D.map {
-            placement.origin + $0.x * placement.xAxis + $0.y * placement.yAxis
-        }
+        let pts3D = e.profilePoints2D.map { placement.lift($0) }
         guard let wire = Wire.polygon3D(pts3D, closed: true) else {
             recordSkip(ctx: &ctx, id: e.id,
                        reason: .occtFailure("wire construction failed"),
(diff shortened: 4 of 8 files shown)


gsdali:
Reviewed as pass owner. Two things are better than the earlier discarded attempt at this issue, and
one is a defect this PR introduces.

**Better than the prior attempt, both deliberately checked:**
- `Flange.worldPoint` is gone, not left as a dead forwarder with no callers.
- `placement` is a **stored** property computed once in `init`, not a computed property rebuilt on
  every access inside the per-point loops at `:592`, `:593` and `:659`.

**The operation count is derived, not hand-edited.** `count-operations.py` on this branch reports
`DERIVED 4340` with both README and API_REFERENCE matching, ✓ on both.

---

## Blocking: `Flange.placement` publishes a basis that violates `Placement`'s contract

`SheetMetal.swift`, `Flange.init`:

```swift
let n = Vector3DMath.normalize(normal) ?? normal
self.normal = n                                    // normalized
self.vAxis = vAxis ?? Vector3DMath.cross(n, uAxis)
self.placement = Placement(
    origin: origin, xAxis: uAxis, yAxis: self.vAxis, zAxis: normal)   // raw `normal`, not `n`
```

`Placement` documents itself as "origin plus an **orthonormal** basis", with `// unit` on each of
`xAxis`, `yAxis`, `zAxis` (`ConstructionEntity.swift:17-22`).

1. **`zAxis` gets the raw parameter while `self.normal` gets the normalized one.** For any caller
   passing a non-unit normal, `flange.normal != flange.placement.zAxis`: two public properties of one
   struct disagreeing about the same direction. That is the exact divergence class this pass exists
   to remove, introduced by the fix for it. One-character cause, `normal` where `n` was meant.
2. **`xAxis` is never normalized at all**, and `yAxis` is `cross(n, uAxis)`, unit only if `uAxis`
   already is.

**`lift()` itself is not affected**, and I checked rather than assumed: it reads only
`origin`/`xAxis`/`yAxis`, and the old `worldPoint` was `origin + p.x * uAxis + p.y * vAxis`, so a
non-unit `uAxis` scaled the result before this PR too. No behavioural regression on the lift path.
What is new is the **public `placement` property**, which hands callers a `Placement` they are
entitled to treat as orthonormal.

### Why `zAxis: n` alone is the wrong fix

It removes the `normal` disagreement, and it should be made regardless. But it leaves `xAxis`
non-unit, and normalizing `xAxis` is **not** available as a follow-up: `lift()` must keep scaling by
`uAxis` to match `worldPoint`, so normalizing it would change every flange's geometry.

That is the real finding. `Flange`'s own doc comment says `uAxis` and `vAxis` "need not be derivable
from `normal` — explicit control of all three axes", i.e. `Flange` deliberately carries a possibly
non-orthonormal frame, and `Placement` deliberately does not. Storing one in the other is a type
mismatch, not a spelling mistake.

Three ways out, and I would take the second:

1. `zAxis: n` and document on `Flange.placement` that its basis is not necessarily unit. Cheapest,
   but it weakens `Placement`'s contract for every other consumer of the type.
2. **Keep `placement` `private`/`fileprivate`** and expose only the lifting behaviour, so no caller
   can observe the non-orthonormal basis. `lift` is the only thing the three call sites need, and
   this issue is about deduplicating the lift, not about publishing a frame.
3. Normalize the whole basis and accept the geometry change, with tests proving what moves. Out of
   scope for a dedup issue and source-breaking.

Whichever you pick, `zAxis: n` goes in too.

## Non-blocking

`SheetMetal.swift:483` still reads "whose **worldPoints** are on the seam line", referring to the
method this PR deletes. A doc comment that no longer describes the code below it is Angle 3 of this
very audit, so it should not ship in a PR from it.

## What I verified

`worldPoint` has no remaining definition or call site; `placement` is stored, not computed;
`count-operations.py` derives 4340 on this branch and both docs match; no `CHANGELOG` or
`docs/SEMVER.md` in the diff; `gate-scripts`, `code-style` and `code-structure` all green.

I have **not** independently re-run the full suite against this branch, and the manifest removals
(`Sketch`, `SheetMetal`, `FeatureReconstructor`) mean most of the 359/251 line delta is the required
`swift-format` reformat rather than logic. Worth a reviewer's eye on the FeatureReconstructor hunks
specifically, where reformat and real change are interleaved.

Issue #1004, opened 2026-08-20
GD&T: 71 of 85 accessors on the three XCAFDimTolObjects classes are still unwrapped


gsdali:
#996 landed the GD&T **model**: one read family, typed enums, and OCCT's dimension kinds as
distinct `Bounds` cases instead of one struct with zero-filled fields. It deliberately left the
rest of the accessor surface alone, because the model change was source-breaking and worth doing
once, and widening the structs afterwards is additive.

This is what it left. Counts are measured against the pinned kernel's headers
(`Libraries/OCCT.xcframework/macos-arm64/Headers`), counting public `Get*`/`Is*`/`Has*`/`Nb*`
methods, and the "exposed" column is what `Document.Dimension`/`GeomTolerance`/`Datum` carry after
#996.

| OCCT class | public accessors | exposed | exposed before #996 |
|---|---|---|---|
| `XCAFDimTolObjects_DimensionObject` | 42 | 11 | 4 |
| `XCAFDimTolObjects_GeomToleranceObject` | 22 | 2 | 2 |
| `XCAFDimTolObjects_DatumObject` | 21 | 1 | 1 |

## `XCAFDimTolObjects_DimensionObject`, 31 unexposed

Exposed after #996: `GetType`, `GetValue`, `GetValues` (as the `Bounds` discriminator),
`IsDimWithRange`, `GetLowerBound`, `GetUpperBound`, `IsDimWithPlusMinusTolerance`,
`GetLowerTolValue`, `GetUpperTolValue`, `IsDimWithClassOfTolerance`, `GetClassOfTolerance`.

Unexposed, grouped by what a caller would use them for:

- **Qualifiers.** `GetQualifier` / `HasQualifier` (`XCAFDimTolObjects_DimensionQualifier`, 4
  members: min, max, average), `GetAngularQualifier` / `HasAngularQualifier`
  (`XCAFDimTolObjects_AngularQualifier`, 4 members: small, large, equal). Both are the "is this
  20mm nominal, or 20mm **maximum**" distinction, which changes what the number means.
- **Modifiers.** `GetModifiers` (`NCollection_Sequence<XCAFDimTolObjects_DimensionModif>`, 24
  members). A sequence, so this needs a bridge count-plus-index pair rather than a scalar.
- **Presentation.** `GetNbOfDecimalPlaces`, `GetPlane` / `HasPlane`, `GetPointTextAttach` /
[...]

PR #1011, opened 2026-08-20
One GD&T read family, and OCCT's dimension kinds instead of zero-filled fields (#996)

### LINKED ISSUE #996: GD&T read surface exists twice: GDTInfo/Document untyped family and GDTWrite typed family
The GD&T read surface exists twice: an untyped family in `GDTInfo.swift` + `Document.swift`, and a
typed family in `GDTWrite.swift`. Both read the same three bridge calls and rebuild the same fields
into different structs. Six public accessors where three would do.

Neither the body-similarity detector nor a reference scan can find this, which is why the Pass 4a
sweep did not: the two families share no call, and the duplication *is* the absent reference. It was
found by reading `GDTInfo.swift`, a file no pass had claimed until this one adopted it.

## Duplication site

**Type families**

| concept | untyped | typed |
|---|---|---|
| dimension | `GDTInfo.swift:6` `DimensionInfo` | `GDTWrite.swift:75` `Document.Dimension` |
| tolerance | `GDTInfo.swift:18` `GeomToleranceInfo` | `GDTWrite.swift:83` `Document.GeomTolerance` |
| datum | `GDTInfo.swift:26` `DatumInfo` | `GDTWrite.swift:89` `Document.Datum` |

**Accessors, both on `Document`**

| | untyped | typed | shared bridge call |
|---|---|---|---|
| one | `Document.swift:209` `dimension(at:)` | `GDTWrite.swift:97` `typedDimension(at:)` | `OCCTDocumentGetDimensionInfo` |
| one | `Document.swift:219` `geomTolerance(at:)` | `GDTWrite.swift:109` `typedGeomTolerance(at:)` | `OCCTDocumentGetGeomToleranceInfo` |
| one | `Document.swift:226` `datum(at:)` | `GDTWrite.swift:117` `typedDatum(at:)` | `OCCTDocumentGetDatumInfo` |
| all | `Document.swift:238/243/248` `dimensions` / `geomTolerances` / `datums` | `GDTWrite.swift:122/126/130` `typedDimensions` / `typedGeomTolerances` / `typedDatums` | via the singular accessor |

`dimension(at:)` and `typedDimension(at:)` each call `OCCTDocumentGetDimensionInfo(handle, Int32(index))`,
guard on the same `info.isValid`, and read the same four fields. Same shape for the tolerance pair.

The three pairs sit at three different stages of one unfinished migration:

- **dimension** and **tolerance**: two independent bridge calls, fully duplicated.
- **datum**: `typedDatum(at:)` already delegates to `datum(at:)` and only re-wraps the result. Half converged.

## Divergence

**The copies agree today.** Checked rather than assumed, because the obvious-looking bug is not one:

`typedDimension(at:)` returns `nil` when `DimensionType(rawValue: info.type)` fails, where
`dimension(at:)` returns the value with a raw `Int32`. That reads as a live behavioural split, and it
is not reachable at the pinned kernel. Measured against `Libraries/OCCT.xcframework/macos-arm64/Headers`:

- `XCAFDimTolObjects_DimensionType` declares **32** members; `Document.DimensionType` declares **32**
  cases, ordinals contiguous `0...31`, names aligned in declaration order.
- `XCAFDimTolObjects_GeomToleranceType` declares **16**; `Document.GeomToleranceType` declares **16**,
  contiguous `0...15`, aligned.
- `OCCTBridge_Document.mm:989` and `:1029` set `info.type = (int32_t)dimObj->GetType()`, a direct cast
  of the OCCT enum with no sentinel and no remap.

So every value the bridge can produce has a case, and `rawValue:` cannot fail. This is drift-in-waiting,
not drift: **the two enums are hand-transcribed from OCCT headers with no gate checking them.** The
moment OCCT adds a member (both enums have grown across versions), `dimension(at:)` keeps returning it
and `typedDimension(at:)` starts returning `nil` for it, silently, with no test that would notice.

Two structural differences that are live now but are not defects:

- `DimensionInfo` / `GeomToleranceInfo` / `DatumInfo` carry no `index`; `Dimension` / `GeomTolerance` /
  `Datum` each carry one.
- The untyped structs are `Sendable` only; the typed ones are `Sendable, Hashable`.

## Tests

Both paths are covered, in the same file, by two suites that **never meet on the same document**, so
no existing test would catch the two families diverging.

- **Untyped**, `Tests/OCCTXCAFTests/OCCTXCAFTests.swift:34` `@Suite("GD&T Document Tests")` /
  `GDTDocumentTests`, lines 43-94. Covers **only the empty and out-of-range cases**:
  `dimensions.isEmpty`, `geomTolerances.isEmpty`, `datums.isEmpty`, and `dimension(at:)` /
  `geomTolerance(at:)` / `datum(at:)` returning `nil` at `0`, `-1`, `999` on an empty document.
  **No test reads a populated dimension, tolerance or datum through the untyped path.**
- **Typed**, `Tests/OCCTXCAFTests/OCCTXCAFTests.swift:4460`
  `@Suite("v0.140 Document GD&T write + typed enums")` / `DocumentGDTTests`, lines 4479-4539. Covers
  **only the populated cases**: authors dimensions/tolerances/datums and reads them back via
  `typedDimension(at:)`, `typedGeomTolerance(at:)`, `typedDatum(at:)`, `typedDimensions`,
  `typedGeomTolerances`, `typedDatums`.

The coverage is exactly complementary: the untyped path is tested only where it returns nothing, the
typed path only where it returns something. Any fix here needs a test that reads the same populated
document through both.

## OCCT functions

Both families reach identical OCCT C++ through the same three bridge functions in
`Sources/OCCTBridge/src/OCCTBridge_Document.mm`:

| bridge fn | OCCT classes and methods |
|---|---|
| `OCCTDocumentGetDimensionInfo` (`:966`) | `XCAFDoc_DimTolTool::Set`, `XCAFDoc_DimTolTool::GetDimensionLabels`, `TDF_Label::FindAttribute`, `XCAFDoc_Dimension::GetID`, `XCAFDoc_Dimension::GetObject`, `XCAFDimTolObjects_DimensionObject::GetType` / `GetValues` / `GetLowerTolValue` / `GetUpperTolValue`, `TColStd_HArray1OfReal::Value` |
| `OCCTDocumentGetGeomToleranceInfo` (`:1006`) | `XCAFDoc_DimTolTool::Set`, `XCAFDoc_DimTolTool::GetGeomToleranceLabels`, `XCAFDoc_GeomTolerance::GetID` / `GetObject`, `XCAFDimTolObjects_GeomToleranceObject::GetType` / `GetValue` |
| `OCCTDocumentGetDatumInfo` (`:1040`) | `XCAFDoc_DimTolTool::Set`, `XCAFDoc_DimTolTool::GetDatumLabels`, `XCAFDoc_Datum::GetID` / `GetObject`, `XCAFDimTolObjects_DatumObject::GetName`, `TCollection_HAsciiString::String` / `Length` |

Enum sources: `XCAFDimTolObjects_DimensionType.hxx`, `XCAFDimTolObjects_GeomToleranceType.hxx`.

## Suggested direction, not prescribed

Keep one family. The typed one is the better surface, and the untyped one is the one with the wider
public reach, so this is a source-breaking change either way and wants a `## SemVer impact` statement.

Whatever lands, the hand-transcribed enums want a derivation or a gate against the pinned headers, in
the shape of `Scripts/derive-bridge-header-split.py`. That is the part that actually stops the drift;
merging the structs alone just makes one copy instead of two of the same untested transcription.

`GDTInfo.swift`, `GDTWrite.swift` and the six `Document.swift` accessors are all clean of
`Scripts/style-manifest-swift.txt`, so no reformat ratchet applies.

gsdali:
## What & why

The GD&T read surface existed twice: an untyped family on `Document` returning raw `Int32` type
codes, and a typed family behind `typedDimension(at:)` and friends, both reading the same three
bridge calls and rebuilding the same fields into different structs. There is now one family, the
typed structs under the untyped family's method names, plus the enum-drift gate that was the half of
#996 merging the structs alone would not have fixed.

Both families also flattened OCCT's dimension kinds into one struct with zero-filled fields, and
that is a live wrong answer rather than only duplication. It is fixed here, along with a second
silent-success bug found while measuring the mutators.

Closes #996

## The wrong answer, reproduced before fixing

`XCAFDimTolObjects_DimensionObject` stores a dimension's magnitude in one values array whose
**length** is the discriminator, and its own predicates are the length test:

| kind | length | predicate | slots |
|---|---|---|---|
| unset | 0 (null array) | neither | nothing |
| simple | 1 | neither | the nominal value |
| range | 2 | `IsDimWithRange()` | lower bound, upper bound |
| plus/minus | 3 | `IsDimWithPlusMinusTolerance()` | value, lower tol, upper tol |

`OCCTDocumentGetDimensionInfo` read `vals->Value(vals->Lower())`, the first slot, and called it
`value`. For a range that is the lower bound, and the two tolerance accessors answer a flat `0`
because a range has no tolerances. **Reproduced against the pinned kernel before touching the
bridge**, both directly on the object and after a real `TDF_Label` round trip
(`Scripts/repro/996-gdt-read-surface/`, `transcript-before.txt`):

```
range 10..12   values=[10,12]  IsRange=1  GetValue=11  GetLowerBound=10 GetUpperBound=12
               bridge before #996 -> value=10 lowerTol=0 upperTol=0
```

`value=10, lowerTol=0, upperTol=0` is exactly what a plain 10mm dimension with zero tolerance reads
back as. A caller could not tell the two apart. `GetValue()` is the accessor that answers correctly:
the single value for a simple dimension, the **midpoint** for a range, `0` for an empty array.

This is the #609/#726 class, so `Scripts/census-unmeasured-values.py` was read first. It does not
flag this site, and correctly: every field was assigned from a real call, so no literal is present.
The fabricated zero is OCCT's, one layer down, returned from an accessor that does not apply. The
remedy is the one #583/#595/#609 established, make absence representable, applied at the boundary
where the inapplicability is known. The census reports **72 / 87 / 3 candidates before this change
and 72 / 87 / 3 after**, measured by stashing.

## A second bug, found measuring the mutators

`SetLowerTolValue` / `SetUpperTolValue` return `false`, and change nothing, for a dimension that is
already a range. `OCCTDocumentSetDimensionTolerance` discarded both returns and reported success, so
`setDimensionTolerance(at:)` on a range dimension answered `true` having done nothing. It now returns
their conjunction. Measured in part 3 of the repro.

## The model

One family, `Document.Dimension` / `GeomTolerance` / `Datum`, reached through `dimension(at:)`,
`geomTolerance(at:)`, `datum(at:)` and the three plurals. `DimensionInfo` / `GeomToleranceInfo` /
`DatumInfo` and the six `typed*` spellings are gone.

```swift
public struct Dimension: Sendable, Hashable {
    public enum Bounds: Sendable, Hashable {
        case unset
        case simple
        case range(lower: Double, upper: Double)
        case plusMinus(lowerTolerance: Double, upperTolerance: Double)
    }
    public struct ClassOfTolerance: Sendable, Hashable {
        public let isHole: Bool
        public let formVariance: DimensionFormVariance
        public let grade: DimensionGrade
    }
    public let type: DimensionType
    public let value: Double?              // GetValue(); nil when bounds is .unset
    public let bounds: Bounds
    public let classOfTolerance: ClassOfTolerance?
    public let index: Int
    // lowerBound / upperBound / lowerTolerance / upperTolerance project out of `bounds`, nil
    // where the kind does not carry that quantity.
}
```

Three decisions worth stating, each taken from a measurement rather than from the shape of the API:

- **`classOfTolerance` is a separate optional, not a fourth `Bounds` case.** OCCT stores the ISO 286
  class outside the values array, so a range dimension can carry one. Measured (part 3), and
  covered by `rangeDimensionCanCarryAClassOfTolerance`.
- **`value` is `Double?`.** `GetValue()` answers `0` for a dimension with no values array, which is
  the same fabricated zero one layer up. The optional is the only shape with no invented number
  anywhere.
- **`.unset` is modelled but is not Swift-reachable.** `createDimension` always writes a one-element
  values array, so nothing in this package can produce one; it comes from imported data. See
  "Notes for the reviewer".

Vocabulary mirrors OCCT so the mapping back to the refman is direct: `Bounds`'s cases are named for
`IsDimWithRange` / `IsDimWithPlusMinusTolerance`, and `DimensionFormVariance` / `DimensionGrade` are
the ISO 286 letter and number `GetClassOfTolerance` returns.

## The upstream doc defect

The pinned header has the doc comments on `GetUpperTolValue` and `GetLowerTolValue` **swapped**, and
the 8.0.1 refman renders the same swap, so it is in OCCT's source and not in our copy:

```cpp
//! Returns the lower value of the toleranced dimension, otherwise - zero.
Standard_EXPORT double GetUpperTolValue() const;

//! Returns the upper value of the toleranced dimension, otherwise - zero.
Standard_EXPORT double GetLowerTolValue() const;
```

**The functions are correct; only the comments are wrong.** Determined empirically, two independent
constructions: writing `SetLowerTolValue(-0.3)` / `SetUpperTolValue(0.7)` yields `[20, -0.3, 0.7]`
with `GetLowerTolValue()` reading back `-0.3`; and binding a raw `[100, 200, 300]` array gives
`GetLowerTolValue()` 200 (slot 2) and `GetUpperTolValue()` 300 (slot 3), the same slots the setters
wrote. The `-0.3 / +0.7` pair is asymmetric on purpose: a symmetric `-0.1 / +0.1` cannot tell a
correct accessor from a swapped one. `OCCTBridge_Document.mm` carries a comment recording the
measurement and saying not to "fix" the bridge to match the comments; `plusMinusDimensionKeeps
ToleranceOrder` pins it, and injection III below is that exact mistake.

Not filed upstream from this PR: it is a comment-only defect in a file no carried patch touches, and
worth batching with the next `XCAFDimTolObjects` submission rather than sending alone.

## The enum-drift gate

`Document.DimensionType` (32) and `GeomToleranceType` (16) were hand-transcribed from the headers
with nothing checking them, and this PR adds two more, `DimensionFormVariance` (29) and
`DimensionGrade` (20). The bridge casts OCCT's enum straight to `int32` with no sentinel and no
remap, and `dimension(at:)` returns `nil` when `rawValue:` fails, so a member OCCT adds becomes a
silently dropped dimension.

`Scripts/derive-gdt-enums.py` derives all four from the pinned headers into
`Scripts/occt-gdt-enums.txt` (97 members) and compares Swift against it:

- `--verify` needs no kernel, so it is what gates in CI's `gate-scripts`, which runs on
  `ubuntu-latest` with no `Libraries/OCCT.xcframework`. A gate that read the headers directly would
  report SKIPPED on every CI run. Same split, same reason, as `Scripts/occt-packages.txt`.
- `--reverify-headers` re-derives from the pinned headers, for after an OCCT bump. It also lists the
  unbound enums so the coverage boundary is visible rather than implied.
- `--write-manifest` rewrites the manifest.
- `--self-test` is 12 cases.

Wired into `ci.yml`'s `gate-scripts` (self-test plus `--verify`) and into the optional pre-commit
hook, matching flag for flag as that hook requires.

**Which enums are gated, and which are not.** The four that are bound, and no more.
`XCAFDimTolObjects` ships eleven others (`AngularQualifier` 4, `DatumModifWithValue` 5,
`DatumSingleModif` 22, `DatumTargetType` 5, `DimensionModif` 24, `DimensionQualifier` 4,
`GeomToleranceMatReqModif` 3, `GeomToleranceModif` 17, `GeomToleranceTypeValue` 3,
`GeomToleranceZoneModif` 4, `ToleranceZoneAffectedPlane` 3), and not one is bound in Swift, because
the accessors that would return them are unwrapped. Gating an enum nothing returns would assert a
correspondence that does not exist. `--reverify-headers` names all eleven with member counts, and
#1004 says any that get bound are added to the gate in the same PR.

## Proving the tests fail

**Nine tests added.** Every one run once with its subject broken. Seven injections into the bridge
and one into Swift, one at a time, restoring between; the counts are of the 22 tests in the three
GD&T suites.

| # | injection | tests failed |
|---|---|---|
| I | `value` = first array slot (the pre-#996 formula) | 1: range keeps its bounds |
| II | no kind discrimination, everything `.simple` | 5: range, plus/minus, range+class, refused tolerance, and the pre-existing create-and-read |
| III | `lowerTol`/`upperTol` swapped (the header's comments taken literally) | 2: plus/minus order, create-and-read |
| IV | class of tolerance never reported | 2: H7 hole, range+class |
| V | the two OCCT setter results discarded (pre-#996 behaviour) | 1: tolerance on a range is refused |
| VI | plural accessor rebuilt independently of the singular (the two-family shape) | 1: populated document through one family |
| VII | `setDimensionBounds` reports success on a missing index | 1: mutators reject an out-of-range index |
| VIII | a one-element values array reported `.unset` | 2: simple dimension, H7 hole |

V, VI and VII each fail exactly one test, and a different one each, which is what makes them
isolating rather than coincidental. I and II both hit the range test, and II's larger set is the
point: it is the whole discriminator, so it takes out every kind-dependent assertion at once.

A ninth injection was tried first and rejected: making the shared lookup return `true` past the end
of the sequence leaves a null handle for the caller to dereference, so the test process crashes
instead of failing an assertion. VII is the same defect expressed where a test can observe it.

**The gate's `--self-test` is 12 cases, with a full removal matrix.** Every guard drops the count:

| guard removed | cases correct |
|---|---|
| (baseline) | 12/12 |
| A: all-caps token rule in `swift_case_name` | 6/12 |
| B: leading anchor on `SWIFT_CASE` (the doc-comment trap) | 5/12 |
| C: explicit-ordinal branch in `parse_occt_enum` | 11/12 |
| D: "an unrelated Swift enum is not a defect" tolerance | 11/12 |
| E: the COUNT check in `compare` | 11/12 |
| F: the MISSING-ENUM check in `compare` | 11/12 |
| G: enum-close detection in `parse_swift_enums` | 11/12 |
| H: the ordinal column in `read_manifest` | 11/12 |

A and B drop by more because both break a fixture every comparison case shares; C through H each
isolate one case. **Row G came back green on the first matrix**, meaning the guard did nothing on
this tree, so rather than keep or delete it on an argument, a case was written for the shape it
actually protects against (a trailing `: Int` raw enum, which `SWIFT_ENUM_OPEN` deliberately ignores,
leaking its cases into the previous enum) and the row went red.

The gate was also run against the real tree with two injections: dropping `sizeWithPath` from
`DimensionType` (COUNT plus two MEMBER rows, exit 1), and transposing `js` and `j` in
`DimensionFormVariance` with the count unchanged (two MEMBER rows, exit 1).

## Verification

- Full `swift test`: **5666 tests in 1467 suites, all passing**.
- `swift build` clean, no warnings from the touched files.
- Every gate and every `--self-test`, including `derive-gdt-enums.py --verify` and
  `--reverify-headers`: all pass.
- `check-style-manifest.py --base origin/main`: clean. All three Swift files and both bridge files
  are off the manifest and stay off it; `swift-format lint --strict`, `swiftlint --strict` (0
  violations in 225 files) and `clang-format --dry-run --Werror` all clean.
- `count-operations.py`: 4339, matching README and `docs/API_REFERENCE.md`.
- `census-unmeasured-values.py`: 72 / 87 / 3 before and after.

## Docs

`docs/reference/Annotation.md` is now the single reference for the whole GD&T surface: the four
enums, `Bounds`, `ClassOfTolerance`, the merged read path, and the write path including the two new
setters and `setDimensionTolerance`'s changed return. `docs/reference/Document.md`'s GD&T section is
replaced by a pointer to it, which is what removes the duplicate documentation the duplicate API had.
`docs/occtswift-wrapping-gaps.md`'s unbound-enum bullet drops from twenty-one to nineteen and from
thirteen GD&T enums to eleven.

### One GD&T read family, and OCCT's dimension kinds instead of zero-filled fields (#996)

`Document`'s GD&T read surface existed twice: an untyped family (`DimensionInfo`,
`GeomToleranceInfo`, `DatumInfo`, reached through `dimension(at:)` and friends) and a typed family
(`Document.Dimension`, `GeomTolerance`, `Datum`, reached through `typedDimension(at:)` and friends),
both reading the same three bridge calls. There is now one: the typed value types under the untyped
family's method names. `DimensionInfo`, `GeomToleranceInfo`, `DatumInfo`, `typedDimension(at:)`,
`typedGeomTolerance(at:)`, `typedDatum(at:)`, `typedDimensions`, `typedGeomTolerances` and
`typedDatums` are removed.

**A range dimension used to read back as a different dimension.**
`XCAFDimTolObjects_DimensionObject` encodes a dimension's magnitude in a values array whose length
is the discriminator, and the bridge read the first slot and called it `value`. For a range that is
the lower bound, and both tolerance accessors answer 0, so a 10..12 range came back as
`value = 10, lowerTolerance = 0, upperTolerance = 0`, which is exactly what a plain 10mm dimension
with zero tolerance comes back as. `Document.Dimension.bounds` mirrors OCCT's own predicates
instead:

```swift
if let dim = doc.dimension(at: 0) {
    switch dim.bounds {
    case .range(let lower, let upper): print("ranges \(lower) to \(upper)")
    case .plusMinus(let lo, let hi): print("\(dim.value ?? 0) \(lo)/\(hi)")
    case .simple: print("\(dim.value ?? 0)")
    case .unset: print("no value")
    }
}
```

`value` is now `Double?` and `classOfTolerance` is optional, so nothing OCCT reports as an
inapplicable 0 is surfaced as a measurement. The four flat accessors survive as computed properties
returning `Double?`: `lowerBound`, `upperBound`, `lowerTolerance`, `upperTolerance`.

**The ISO 286 tolerance class is now readable.** `Document.DimensionFormVariance` (29 cases) and
`Document.DimensionGrade` (20 cases) join `DimensionType` and `GeomToleranceType`, and
`Dimension.classOfTolerance` reports the `H7`-style class OCCT stores outside the values array.
Three new write methods reach the kinds: `setDimensionBounds(at:lower:upper:)`,
`setDimensionClassOfTolerance(at:isHole:formVariance:grade:)`, and
`setDimensionTolerance(at:lower:upper:)` which now returns `false` instead of a silent `true` when
OCCT refuses the call, as it does for a dimension that is already a range.

**The four enums are now gated against the pinned kernel.** `Scripts/derive-gdt-enums.py` derives
all 97 members from OCCT's own headers into `Scripts/occt-gdt-enums.txt` and fails CI on any
divergence, which nothing checked before: the bridge casts OCCT's enum across unremapped, so a
member OCCT adds used to become a silently dropped dimension.

Reproducer and measurements: [`Scripts/repro/996-gdt-read-surface/`](Scripts/repro/996-gdt-read-surface/).
The remaining accessor surface, 71 of 85 across the three `XCAFDimTolObjects` classes, is #1004.

## SemVer impact

MAJOR. Nine public symbols are removed and one public struct changes shape.

Removed: `DimensionInfo`, `GeomToleranceInfo`, `DatumInfo`, `Document.typedDimension(at:)`,
`typedGeomTolerance(at:)`, `typedDatum(at:)`, `typedDimensions`, `typedGeomTolerances`,
`typedDatums`.

Changed: `Document.Dimension` loses its stored `value: Double`, `lowerTolerance: Double` and
`upperTolerance: Double` and gains `value: Double?`, `bounds: Bounds` and
`classOfTolerance: ClassOfTolerance?`; `lowerTolerance` and `upperTolerance` come back as computed
`Double?`. `Document.setDimensionTolerance(at:lower:upper:)` keeps its signature but now returns
`false` where it previously returned `true` for a call OCCT refused.

Migration:

- `doc.typedDimension(at: i)` → `doc.dimension(at: i)`, and the same for the other five `typed*`
  spellings. The value type is unchanged in name.
- `dim.value` is now optional: `dim.value ?? 0` restores the old reading, though a caller that
  cares about the difference should switch on `dim.bounds` instead.
- `dim.lowerTolerance` / `dim.upperTolerance` are now `Double?` and are `nil` for anything but a
  plus/minus dimension. `?? 0` restores the old reading exactly, including the case where the 0 was
  wrong.
- A caller holding a `DimensionInfo` moves to `Document.Dimension`; `type` changes from `Int32` to
  the `DimensionType` enum, so `info.type == 15` becomes `dim.type == .sizeDiameter`.

## Checklist

      verification), see [SecondMouseAU/OCCTReconstruct#397](https://github.com/SecondMouseAU/OCCTReconstruct/issues/397)
      for the ecosystem-wide test-coverage standard this is piloting.
      failure is reported here, see [okf/policies/prove-the-test-fails.md](../okf/policies/prove-the-test-fails.md).
      It is assessed at release on `main`, not per PR.

## Notes for the reviewer

**The one case with no Swift test, stated rather than hidden.** `Bounds.unset` and the `nil` `value`
that goes with it are not reachable from Swift: `createDimension` always writes a one-element values
array, and nothing else on `Document` produces a dimension. It arrives only from imported data,
where OCCT's `myVal` can be null. It is modelled anyway, because the bridge has to decide what to
report for that shape and reporting `value = 0` is the same defect this PR fixes one case over.
OCCT's side of it is measured in the repro's part 3 (`GetValues().IsNull() = 1`, `GetValue() = 0`),
which is the strongest evidence available without a hand-built STEP fixture. Injection VIII exercises
the bridge's `.unset` branch from the other direction, by forcing a reachable dimension down it.

**Why one family rather than keeping both and deprecating.** The issue's own framing, and the reason
this is worth a source break, is that the two families' test coverage was exactly complementary:
`GDTDocumentTests` tested the untyped path only where it returned nothing, `DocumentGDTTests` the
typed path only where it returned something. No test read a populated document through both, so
nothing would have caught them diverging. `populatedDocumentReadsThroughOneFamily` is the test that
closes that, and injection VI is what proves it would catch a re-divergence.

**`GeomTolerance` and `Datum` are unchanged apart from their names' reachability.**
`XCAFDimTolObjects_GeomToleranceObject::GetValue()` is a single stored number with no predicate to
discriminate it, so there is no `Bounds` equivalent to add; the same is true of the datum name.
Their unwrapped accessors, 20 each, are #1004.

**What was kept from the earlier attempt on this issue and what was discarded.** The ground-truth
`.mm` and the overall bridge/Swift shape were sound and are kept, extended with a third part
measuring the mutators. Discarded: a `datums` accessor written as
`(0..<max(0, datumCount - 1))`, which is off by one and would have failed the suite's own new
assertion; a bridge comment claiming `SetLowerBound` must be called first, which measurement shows
is not true in either direction; a non-optional `value: Double` that kept the fabricated zero; the
discarded `SetLowerTolValue`/`SetUpperTolValue` returns; and the gate script itself, which that
attempt referenced from a source comment but never wrote.

**Reproducer.** `Scripts/repro/996-gdt-read-surface/` carries the probe, its transcript against the
pinned kernel, and a README walking the encoding, the doc-comment swap and the mutator contracts.
The transcript is labelled "bridge before #996" and stays valid after the fix, because that line is
arithmetic the probe does itself rather than a call into the bridge.


gsdali:
Reviewed as pass owner. I re-ran the parts that decide whether this is trustworthy rather than
taking the summary on trust.

**The gate is real, verified by my own injections rather than its self-test.** `--self-test` is 12/12
and `--verify` is clean on this branch, but a passing self-test is not evidence the gate catches a
live defect. I injected two into `GDTRead.swift` directly:

- deleted `case dimensionPresentation = 31` → `problems: 1`
- silently renumbered `case locationLinearDistance = 2` to `= 12` → `problems: 1`

and restoring the file returns it to `problems: 0`. The second is the one that matters: an ordinal
drift is exactly the failure that would otherwise be invisible, since the case still exists and the
code still compiles.

**Wiring confirmed**: `ci.yml` runs both `--self-test` and `--verify` in `gate-scripts`, the required
check, and `--verify` needs no kernel so it will actually run in CI rather than reporting SKIPPED the
way `census-doc-occt-attribution.py` does.

**The doc-defect handling is right.** The comment left in `OCCTBridge_Document.mm` records the
measurement, states the conclusion the opposite way round from the header ("the functions are not
swapped"), gives the raw `[100,200,300]` evidence, says *do not "fix" these two to match the
comments*, and points at the repro. That is the correct disposition for an upstream doc bug we
cannot fix from here, and it is the kind of note that stops someone "correcting" the bridge in six
months.

**Not gating the eleven unbound enums is the right call**, and worth restating because it will look
like an omission later: gating an enum whose accessor is not wrapped would assert a correspondence
that does not exist. `--reverify-headers` naming all eleven with member counts is what makes that
recoverable when #1004 binds them.

---

## One cross-PR hazard, not a defect in this PR

`count-operations.py` derives **4339** here, and README/API_REFERENCE both match. But PR #997 derives
**4340** on its own branch, because it adds `Placement.lift`. Both are correct against `main`'s 4339
and both are internally consistent; they are not consistent with each other.

Whichever merges second inherits a stale total in two files, and `count-operations.py` is a gate, so
it will go red rather than silently pass. Worth knowing it is coming so nobody hand-edits the number
to clear it. Rebase and re-derive.

## Two things I did not verify

I have not independently re-run the full suite (the 5666-test figure is the agent's), and I have not
re-measured the range-dimension reproduction. Both are claimed with a committed transcript at
`Scripts/repro/996-gdt-read-surface/`, which is the right place for them, and CI's macOS
`swift build + test` job passing on this branch is independent corroboration of the suite figure.

The `Bounds.unset` case being unreachable from Swift is stated in the PR rather than hidden, which is
the correct handling for a branch that only imported data can reach.

## SemVer

MAJOR is right: nine public symbols removed and one return value's meaning corrected. Per
`semver-at-release.md` that belongs in the PR body and nowhere else, and `docs/SEMVER.md` is
correctly absent from the diff.

Issue #1022, opened 2026-08-20
XCAFDoc_Datum::GetObject reads the datum point's X from the annotation plane, and SIGSEGVs when there is no plane


gsdali:
`XCAFDoc_Datum::GetObject` builds the datum's point from the **annotation plane's** location array
instead of the point's own. One line, `XCAFDoc_Datum.cxx`'s `ChildLab_Pnt` block:

```cpp
occ::handle<TDataStd_RealArray> aPnt;
if (Label().FindChild(ChildLab_Pnt).FindAttribute(TDataStd_RealArray::GetID(), aPnt)
    && aPnt->Length() == 3)
{
  gp_Pnt aP(aLoc->Value(aPnt->Lower()),        // aLoc, not aPnt
            aPnt->Value(aPnt->Lower() + 1),
            aPnt->Value(aPnt->Lower() + 2));
  anObj->SetPoint(aP);
}
```

`aLoc` is the handle the `ChildLab_PlaneLoc` block a few lines above fills. The sibling
`XCAFDoc_GeomTolerance::GetObject` has the identical block and reads `aPnt->Value(aPnt->Lower())`,
so this is a one-character divergence rather than a shared idiom.

Two consequences, and the second is the one that matters.

**A wrong answer when the datum has both.** A datum written with plane location `(6,6,6)` and point
`(7,7,7)` reads back with point `(6,7,7)`. The stored point is not recoverable.

**An uncatchable SIGSEGV when the datum has a point and no plane.** `aLoc` is declared before the
plane block and only assigned by a successful `FindAttribute`, so with no annotation plane it is
still a null handle when the point block dereferences it. That is an OS signal, not a
`Standard_Failure`, so the bridge's `catch (...)` cannot absorb it.

## Reachability from OCCTSwift today

`OCCTDocumentGetDatumInfo` calls `datumAttr->GetObject()` for every datum, so `Document.datums`,
`Document.datum(at:)` and `Document.datumCount`'s siblings all take this path. Any imported document
whose datum carries a point without an annotation plane crashes the process on read. Nothing in this
repo can author that shape today (there is no datum point/plane write path), so the exposure is
import-side only, which is also why no existing test catches it.

## Reproducer

`Scripts/repro/1004-gdt-accessors/datum_point_without_plane.mm`, added by #1004's first PR.

```
== part 1: a datum with BOTH a plane and a point ==
  wrote plane loc (6,6,6) and point (7,7,7)
[...]

Issue #1021, opened 2026-08-20
OCCT class coverage: eight classes reached at 12-50%, with the reasons unrecorded


gsdali:
`Scripts/repro/385-coverage-sample/occt-class-coverage.py` measures, per OCCT class, how many of its
public methods this bridge actually invokes. Sampled during Pass 4a. **#1004 already covers the three
`XCAFDimTolObjects` classes**; this is everything else the sample turned up, so the question is not
re-asked from scratch by a later pass.

## Measured on the pinned 8.0.1 headers

| OCCT class | public | we call | cover |
|---|---:|---:|---:|
| `GeomPlate_BuildPlateSurface` | 34 | 4 | 12% |
| `Plate_Plate` | 23 | 4 | 17% |
| `BRepFeat_Gluer` | 9 | 2 | 22% |
| `BRepFeat_MakeDPrism` | 13 | 3 | 23% |
| `XCAFDoc_DimTolTool` | 39 | 9 | 23% |
| `GeomPlate_MakeApprox` | 3 | 1 | 33% |
| `BRepMAT2d_BisectingLocus` | 11 | 5 | 45% |
| `NLPlate_NLPlate` | 12 | 6 | 50% |

For calibration, the same probe reports `ShapeFix_Face` at 82% and `BRepFeat_MakeCylindricalHole` at
88%, so the metric has signal and the low rows are not an artefact of it.

## Read the caveats before acting on any row

**The denominator counts `Set*` and `DumpJson`.** A class this bridge deliberately reads and never
writes will score low without anything being wrong. This screens, it does not adjudicate.

**The probe was wrong the first time it ran, and the failure is instructive.** Scanning only
`Sources/OCCTBridge/src/*.mm` reported `BRepFeat_MakeCylindricalHole` at **0%**, contradicting
CLAUDE.md's own Known OCCT Bugs entry for #532. The calls live in a shared helper in
`OCCTBridge_Internal.h`, which is where this repo deliberately puts them. A `.mm`-only scan is blind
to precisely the most-shared call sites. Fixed, 0% became 88%, and it is recorded in the sample's
README so it is not re-learned.

**No measured false-positive rate**, on the same terms as its siblings: `census-doc-occt-attribution.py`
got 41% over a 40-row hand-adjudicated sample (#928) and that is why it reports rather than gates.

## Why the plate rows are worth looking at first

[...]

PR #1025, opened 2026-08-20
feat(#1004): the GD&T dimension accessors that change what a dimension's number means

### LINKED ISSUE #1004: GD&T: 71 of 85 accessors on the three XCAFDimTolObjects classes are still unwrapped

gsdali:
Closes #1004 in part. This is the dimension half; `XCAFDimTolObjects_GeomToleranceObject` and
`XCAFDimTolObjects_DatumObject`, plus #1021's `XCAFDoc_DimTolTool` row, follow in a second PR
against this same base.

## What lands

`Document.Dimension` gains four fields and `Document.DimensionType` gains two computed properties,
each with the mutator that authors it. Every one was chosen because it changes what a dimension's
number means, which is the test #1004 set:

| Swift | OCCT |
|---|---|
| `Dimension.qualifier`, `setDimensionQualifier(at:_:)` | `GetQualifier` / `HasQualifier` / `SetQualifier` |
| `Dimension.angularQualifier`, `setDimensionAngularQualifier(at:_:)` | `GetAngularQualifier` / `HasAngularQualifier` / `SetAngularQualifier` |
| `Dimension.decimalPlaces`, `setDimensionDecimalPlaces(at:left:right:)` | `GetNbOfDecimalPlaces` / `SetNbOfDecimalPlaces` |
| `Dimension.modifiers`, `setDimensionModifiers(at:_:)` | `GetModifiers` / `SetModifiers` |
| `DimensionType.isDimensionalLocation` / `.isDimensionalSize` | the two static classifiers |

Bridge side: six new C functions, six new `OCCTDimensionInfo` fields, and the count-plus-index pair
#1004 asked for on the modifier sequence.

**Each read accessor ships with its mutator on purpose.** A read this package cannot author has no
way to be tested against a document of our own, which is the bar the task set; that rule is written
into `GDTWrite.swift`'s header comment so the next pass does not have to re-derive it.

## The subset, and the omissions

`XCAFDimTolObjects_DimensionObject` has 42 public accessors. This wraps seven of them and leaves 26
unwrapped, each recorded with its reason in a new
`docs/occtswift-wrapping-gaps.md` section, `GD&T dimension accessors left unwrapped (#1004)`.
Three of those reasons are measurements rather than judgement calls:

**`GetSemanticName` was implemented, measured, and removed again before this PR opened.** The
semantic name is stored in the dimension label's own `TDataStd_Name`, which
`XCAFDoc_DimTolTool::AddDimension()` initialises to the literal `"DGT:Dimension"`. `GetObject` reads
that attribute back, so a dimension nobody named reports the GD&T table's own marker string.
`SetObject`'s `setString` helper returns early on a null handle, so the name cannot be cleared once
set either. The first draft of this PR wrapped it and its test failed with
`(doc.dimension(at: index)?.semanticName -> "DGT:Dimension") == nil`, which is how it was caught.
`XCAFDoc_GeomTolerance` and `XCAFDoc_Datum` carry the same shape with `"DGT:Tolerance"` and
`"DGT:Datum"`.

**`GetDirection` has no presence predicate.** `GetDirection(gp_Dir&)` returns `true`
unconditionally (`XCAFDimTolObjects_DimensionObject.cxx:419-423`) and writes `myDir`, which no
constructor assigns; measured, an object that never had a direction reports `(1,0,0)`, the
default-constructed `gp_Dir`. Surfacing that is the fabricated-magnitude shape #609 exists to catch.

**`GetPath` is deferred, not declined.** It is the one omission with real interpretive value, since
`.locationWithPath` and `.sizeWithPath` cannot be read without it. It returns a `TopoDS_Edge`, so it
is a handle-lifetime change to the read struct rather than a field, and it wants its own PR.

## Two defects found while measuring

**The create path handed OCCT indeterminate members.** `XCAFDimTolObjects_DimensionObject`'s
constructor assigns four booleans and nothing else, and `XCAFDoc_Dimension::SetObject` then reads
the qualifier, the angular qualifier, the decimal places and the tolerance class off the object
unconditionally to decide what to store. `OCCTDocumentCreateDimension` set only the type and the
values array, so those four reads were of storage nobody wrote. Fixed by assigning the neutral
values explicitly.

Stated plainly, because it bears on how much this fix is worth: the probe poisons the allocator's
free list with same-size blocks first, and **the poison does not come through**. Every such member
still reads 0 against this kernel and this allocator, which is why nothing ever caught it. The fix
is correct rather than load-bearing today, and injection G below is what holds it: swapping the
neutral qualifier for a non-neutral one fails two tests, so the guard is covered even though its
current effect is unobservable.

**`XCAFDoc_Datum::GetObject` reads the datum point's X from the annotation plane, and SIGSEGVs when
there is no plane.** Filed as #1022, with a reproducer in this PR's own repro directory. Not fixed
here: it needs a kernel patch and an xcframework rebuild, which is different work from a Swift read
surface. It is already reachable from `Document.datums` today, before any of this lands.

## Enums joining the gate

Three, added to `derive-gdt-enums.py`'s `BOUND` and to `Scripts/occt-gdt-enums.txt` in this PR, per
#1004's own rule. The gate now covers **7 enums, 129 members** (was 4 / 97):

| Swift | OCCT | members |
|---|---|---|
| `Document.DimensionQualifier` | `XCAFDimTolObjects_DimensionQualifier` | 4 |
| `Document.AngularQualifier` | `XCAFDimTolObjects_AngularQualifier` | 4 |
| `Document.DimensionModifier` | `XCAFDimTolObjects_DimensionModif` | 24 |

**Eight are left unbound** and `--reverify-headers` still names them: `DatumModifWithValue`,
`DatumSingleModif`, `DatumTargetType`, `GeomToleranceMatReqModif`, `GeomToleranceModif`,
`GeomToleranceTypeValue`, `GeomToleranceZoneModif`, `ToleranceZoneAffectedPlane`. All eight belong
to the two classes the second PR covers; gating one now would assert a correspondence nothing
returns, which is the reason #996 gave for gating only four.

The gate was proved to catch drift in the three new enums, not only assumed to:

| Injection into `GDTRead.swift` | `--verify` |
|---|---|
| rename `DimensionModifier.anyCrossSection` to `crossSection` | exit 1, `MEMBER DimensionModifier: position 19` |
| delete `AngularQualifier.equal` | exit 1, `COUNT AngularQualifier: OCCT declares 4, Swift has 3` |
| change `DimensionQualifier.avg` to raw value 4 | exit 1, `MEMBER DimensionQualifier: position 3` |
| restored | exit 0 |

## Injection matrix for the new tests

Seven new tests in `GDTDimensionAccessorTests`. Each guard was removed in turn against the real
bridge, the suite re-run, and the file restored, per `okf/policies/prove-the-test-fails.md`.

| Injection into `OCCTBridge_Document.mm` | Tests that fail |
|---|---|
| A qualifier not read (reports 0 always) | qualifier round-trip; angular qualifier; two dimensions |
| B angular qualifier not read (reports 0 always) | angular qualifier |
| C decimal-place presence test replaced by an unconditional `true` | decimal places; two dimensions |
| D modifier count reported as 0 | modifier sequence; two dimensions |
| E modifier index ignored (every position reads the first) | modifier sequence |
| F `IsDimensionalLocation` answers false for every type | type classifiers |
| G create path hands OCCT a non-neutral qualifier | qualifier round-trip; two dimensions |
| H negative decimal-place counts no longer refused | out-of-range indices |
| I out-of-range dimension index no longer refused by the modifier setter | out-of-range indices |

Nothing came back green, so no case is decorative. Two remarks on what the rows do and do not
isolate:

- **A and B are distinguished by strict superset, not by disjointness.** Both fail the angular
  qualifier test, because that test asserts the pair together on purpose, so that neither accessor
  can be reading the other. A additionally fails two more; B fails only the one. That is what
  separates "the dimension qualifier is broken" from "the angular one is".
- **H and I produce the same row and neither isolates the other.** They are two guards inside one
  test. The test covers both, the matrix cannot tell them apart, and splitting the test to make it
  do so would buy a distinction nobody reads.

## Verification

- `swift build` clean, no new warnings.
- Full `swift test`: **5,714 tests in 1,478 suites, 0 failures** (branch baseline 5,707 in 1,477;
  the difference is this PR's own suite).
- `swift test --filter GDTDimensionAccessorTests`: 7/7. `--filter GDTUnifiedReadTests` (#996's
  suite, whose struct this widens): 10/10.
- All seven gates, both censuses and the merge-history audit: clean.
- Every `--self-test`: `check-bridge-index` 18/18, `check-null-handle-guards` 24/24,
  `check-docs-defaults` 13, `check-docs-existence` 27, `check-borrowed-handles` 15/15,
  `derive-bridge-header-split` 8/8, `derive-gdt-enums` 12/12, `census-unmeasured-values` 36/36,
  `census-doc-occt-attribution` 20, `check-changelog-transcription` 23/23,
  `check-style-manifest` 6/6, `comment-ratio-check` 5/5.
- `derive-gdt-enums.py --verify` 0 problems, `--reverify-headers` matches the pinned headers.
- `count-operations.py` 4,333 -> 4,339, README and API_REFERENCE rewritten by `--fix`.
- `check-style-manifest.py --base origin/refactor/385-pass4a` clean. None of the four touched
  source files was on either manifest, and each was confirmed clean with `swift-format lint
  --strict` / `clang-format --dry-run --Werror` before and after, rather than taken on trust.
  `swiftlint --strict`: 0 violations in 224 files.

**One deliberate style exception, stated rather than hidden.** `docs/reference/Annotation.md`
carries 96 pre-existing em-dashes, and `writing-style.md` asks for them to be cleared from a file
being edited. They are left, and every line this PR adds to that file uses a colon instead. The
reason is measured, not convenience: `check-docs-existence.py` parses `' — '` as the heading
qualifier separator (`Scripts/check-docs-existence.py:498`), the convention runs across 40+
`docs/reference/*.md` pages, and changing it in one file is a gate-input change that wants its own
chore PR.

## Deferred to the second PR

`GeomToleranceObject`'s 20 unexposed accessors, `DatumObject`'s 20, their eight enums, and #1021's
`XCAFDoc_DimTolTool` coverage row. The repro directory's README says so, so the boundary is not
implicit.

### GD&T: a dimension's qualifiers, modifiers and decimal places are now readable (#1004)

`Document.Dimension` gains `qualifier`, `angularQualifier`, `decimalPlaces` and `modifiers`, and
`Document.DimensionType` gains `isDimensionalLocation` and `isDimensionalSize`. Each read accessor
ships with the mutator that authors it: `setDimensionQualifier(at:_:)`,
`setDimensionAngularQualifier(at:_:)`, `setDimensionDecimalPlaces(at:left:right:)` and
`setDimensionModifiers(at:_:)`.

Three new enums, `Document.DimensionQualifier`, `Document.AngularQualifier` and
`Document.DimensionModifier`, are transcribed from the pinned kernel's own headers and join
`Scripts/derive-gdt-enums.py`'s gate, which now covers 7 enums and 129 members.

Absence is representable everywhere it needed to be. `decimalPlaces` is `nil` rather than a
fabricated `(0, 0)`, gated on the same `theL > 0 || theR > 0` condition OCCT itself stores the pair
under. The two qualifiers are not optional, because OCCT's own enums carry a `_None` member and
wrapping that in an `Optional` would give two spellings of one state.

`OCCTDocumentCreateDimension` now assigns the qualifier, angular qualifier, decimal places and
tolerance class explicitly. `XCAFDimTolObjects_DimensionObject`'s constructor leaves all four
unassigned and `XCAFDoc_Dimension::SetObject` reads them unconditionally, so the create path was
handing OCCT storage nobody wrote. Measured against the pinned kernel it reads back as zero, so no
behaviour changes today.

Twenty-six accessors are deliberately not wrapped, each with its reason in
`docs/occtswift-wrapping-gaps.md`. `GetSemanticName` is the one worth knowing: it reads the GD&T
table label's own `TDataStd_Name`, which OCCT initialises to `"DGT:Dimension"`, so an unnamed
dimension reports that marker string and a name cannot be cleared once set.

## SemVer impact

MINOR. Every change is additive. `Document.Dimension` gains four stored properties, but its
memberwise initializer is internal (no `public init`), so no consumer can be constructing one and
none will fail to compile. `Document.DimensionType` gains two computed properties, three new nested
enums appear on `Document`, and four new methods appear on `Document`. Nothing is removed or
renamed, and no existing signature changes.


gsdali:
Follow-up #1028 is stacked on this branch and carries the tolerance and datum half plus #1021's `XCAFDoc_DimTolTool` verdict. Merge this one first; GitHub retargets #1028 to `refactor/385-pass4a` automatically. Per `okf/feedback-stacked-pr-merging`, do **not** pass `--delete-branch` when merging this one while #1028 is still open.

gsdali:
Correction to the note above: #1028 has been **retargeted to `refactor/385-pass4a`**, so it is no longer based on this branch and the `--delete-branch` caution does not apply. Merge this one normally.

The reason for the retarget is worth recording. `ci.yml`, `code-style.yml` and `code-structure.yml` all trigger only on a base of `main` or `refactor/**`, so a PR based on `feat/385-gdt-accessors` dispatched **no GitHub Actions at all**, only the third-party review bot. Retargeting alone did not fix it either: a base change fires `pull_request: edited`, which is not in the default trigger types, so #1028 had to be closed and reopened to get a `reopened` event. Worth knowing before anyone else stacks a PR in this repo.

The cost is that #1028 shows both commits until this one merges. Its body says which commit to review.

Issue #1030, opened 2026-08-20
Guard the shared datum lookup against #1022's uncatchable crash until a rebuilt kernel ships


gsdali:
`Scripts/patches/0029-*` fixes [#1022](https://github.com/SecondMouseAU/OCCTSwift/issues/1022) in the
kernel, but it is in no built kernel: `Package.swift` pins the v3.0.0 asset, which carries
`0010`-`0012` and `0014`-`0027`. Until a rebuilt asset ships, an uncatchable SIGSEGV is live for
every consumer, and this issue is what tracks the decision to guard bridge-side or not.

## What is exposed

`occtDocumentDatumObjectAt` (`OCCTBridge_Document.mm`, hoisted by #1025) calls
`XCAFDoc_Datum::GetObject()`, and seven bridge functions route through it:

| Bridge function | Kind |
|---|---|
| `OCCTDocumentGetDatumInfo` | read (`Document.datums`, `Document.datum(at:)`) |
| `OCCTDocumentGetDatumModifier` | read |
| `OCCTDocumentSetDatumPosition` | write |
| `OCCTDocumentSetDatumModifiers` | write |
| `OCCTDocumentSetDatumModifierWithValue` | write |
| `OCCTDocumentSetDatumTargetPlacement` | write |
| `OCCTDocumentSetDatumTarget` | write |

The helper calls `GetObject()` before any caller can look at what it returned, so a write that never
touches the point still takes the crash. The trigger is a datum carrying a point without an
annotation plane.

## Whether that shape is reachable

Measured in `Scripts/repro/1022-datum-point-from-plane-array/`, not inferred:

- **A STEP import alone cannot produce it.** `STEPCAFControl_Reader` sets a datum plane and never a
  datum point; its only `SetPoint` calls are on an `XCAFDimTolObjects_DimensionObject`.
- **An OCAF load can.** The reproducer's section D saves a point-only datum to BinXCAF, reloads it
  into a fresh `TDocStd_Application`, and crashes on the reload.
  `XCAFDimTolObjects_DatumObject::HasPlane()`/`HasPoint()` are independent flags and
  `XCAFDoc_Datum::SetObject` writes the two label blocks independently, so this is what the public
  OCCT API produces, not a hand-forged label tree.
- **Nothing in this repo authors it today.** `OCCTDocumentCreateDatum` sets a name and nothing else.
  That is a property of the current write surface, and it is the only reason the suite is green.

## The guard, and what it costs

[...]

Issue #999, opened 2026-08-20
Thirteen bridge functions ignore a declared parameter, several promising a capability the implementation lacks


gsdali:
Thirteen bridge functions declare a named parameter and never read it. All eleven checked are
reachable from public Swift API, so in each case a caller passes a value that cannot affect the
answer. Several are worse than inert: the parameter names a capability the implementation does not
have, or the function does not call the OCCT API it is named for.

Found by `Scripts/repro/385-coverage-sample/detect-dead-parameters.py`, committed with the
measurement it produced.

## This is a recurrent failure, not a new one

Two closed issues are this exact class, each found by hand, one per pass:

- **#431** (Pass 1a): "FillingParameters.maxDegree/maxSegments don't control what their names say."
- **#837** (Pass 2a): "Shape.fixed(fixShell:fixFace:fixWire:) parameters are dead code, silently ignored."

Same two parameter names as one of the sites below, in a different entry point. Nobody built a
detector; a mechanical one finds thirteen more in ten minutes. That is the argument for gating this
rather than continuing to find them one per pass.

### A. The parameter has no OCCT counterpart at all

| site | dead | measured |
|---|---|---|
| `OCCTBridge_Geom2d.mm:8928` `OCCTCurve2DToBSpline` | `tolerance` | `Geom2dConvert::CurveToBSplineCurve(C, Convert_ParameterisationType)` takes **no tolerance**. The conversion is exact. Swift: `Curve2D.toBSpline(tolerance: Double = 1e-6)`. Loosening the tolerance cannot simplify the result, which is what a caller would reasonably expect. |
| `OCCTBridge_Healing.mm:3966` `OCCTWireCheckOuterBound` | `prec` | `ShapeAnalysis_Wire::CheckOuterBound(APIMake: bool = true)` takes a **bool**, not a precision. And the bridge never calls it: it runs a `TopExp_Explorer` for any wire. Swift: `SAWireAnalysis.checkOuterBound(face:precision:)`, sitting among siblings that all take a real `precision` and do call `ShapeAnalysis_Wire`. |

### B. The counterpart exists and is hardcoded past

| site | dead | measured |
|---|---|---|
[...]

PR #1032, opened 2026-08-20
fix(#1018, #1022): carry kernel patches 0028 and 0029 for GeomPlate_BuildPlateSurface and XCAFDoc_Datum

### LINKED ISSUE #1018: GeomPlate_BuildPlateSurface's G0/G1/G2Error return uninitialised memory for a point-only plate
`GeomPlate_BuildPlateSurface::G0Error()`, `G1Error()` and `G2Error()` return uninitialised memory
after a `Perform()` that had only point constraints. This is an upstream defect in the pinned 8.0.1
kernel, and it is the reason `Surface.plateErrors` is deleted by
[#1015](https://github.com/SecondMouseAU/OCCTSwift/pull/1015) rather than repaired.

## Mechanism

Read from `Libraries/occt-src/.../GeomPlate/GeomPlate_BuildPlateSurface.cxx`:

- `myG0Error` / `myG1Error` / `myG2Error` are plain `double` members with **no in-class
  initialiser** (`GeomPlate_BuildPlateSurface.hxx:261-263`), and none of the three constructors
  assigns them.
- `VerifSurface()` is the **only** writer (`:2597`, `myG0Error = 0, myG1Error = 0, myG2Error = 0;`
  followed by the max loop at `:2625-2635`).
- `Perform()` calls `VerifSurface()` only on the branch that has curve constraints (`:684`). The
  point-only branch (`:699-726`) ends with `VerifPoints(di, an, cu)`, which computes its deviations
  into three **locals** and discards them.

So a plate built from point constraints alone leaves all three members untouched, and the three
public getters (`:1237-1255`, each a bare `return myXError;`) hand back whatever was in that memory.
The `G0Error(Index)` overloads do not help: they go through `EcartContraintesMil(Index, ...)`, which
is about `myLinCont`, the curve constraints.

## Measured

Directly, in C++ against the pinned kernel, on a 25-point wavy grid:

```
builder(3, 10, 5, Tol2d=1e-5, Tol3d=1e-1)  G0=3.03135716117e-314 G1=6.17902279823e-314 ...
```

and the same inputs on the next run of the same binary give `3.046e-314`, then `3.016e-314`. Not
reproducible run to run, which is what distinguishes uninitialised memory from a bad computation.

Through the real bridge, before its deletion, three consecutive `Surface.plateErrors` calls in one
process on one fixture:

```
run 0: (g0Error: 1.9355577472e-313,       g1Error: 5e-324,   g2Error: 4.217484478e-314)
run 1: (g0Error: -3.105038551588571e+231, g1Error: 1.6e-322, g2Error: 3.16e-322)
run 2: (g0Error: 3.818182542e-314,        g1Error: -nan,     g2Error: 3.16e-322)
```

Full transcripts and the probe: `Scripts/repro/999-nlplate-plate-errors/`.

## Blast radius in this repo

`OCCTGeomPlateErrors` was the only reader of these three getters, and it is deleted. The other five
`GeomPlate_BuildPlateSurface` sites (`OCCTShapePlatePoints`, `OCCTShapePlateCurves`,
`OCCTShapePlatePointsAdvanced`, `OCCTShapePlateMixed`, `OCCTSurfacePlateThrough`,
`OCCTGeomPlateSurface`) never call them, so nothing is currently exposed. This is filed so the next
person to reach for `G0Error()` finds the measurement rather than the number.

## What the fix is, and what it is not

Initialising the three members in the constructors is a one-line upstream hygiene fix and removes
the undefined behaviour. It does **not** make the API correct: it would make a point-only plate
report `0`, which says "no error" about something that was never measured, and that is the failure
class `Scripts/census-unmeasured-values.py` exists for (#726). A point-only plate would need
`VerifPoints`' three values kept rather than discarded, which is a larger change to a public class's
semantics.

Suggested sequence: carry the initialisation as a patch and file it upstream as the memory-safety
fix it is, then decide separately whether OCCTSwift wants a `plateErrors(curves:)` entry point, which
would report real numbers because `VerifSurface` runs on that branch.

Found while working #999 (Pass 4a, #385).

### LINKED ISSUE #1022: XCAFDoc_Datum::GetObject reads the datum point's X from the annotation plane, and SIGSEGVs when there is no plane

gsdali:
## What & why

Two upstream OCCT defects, each a one-file kernel fix, each carried as a patch and filed upstream. They land in one PR because they share `Package.swift`'s manifest comment and the in-repo lists of what is carried, and splitting those edits across two PRs would misattribute them and conflict on every shared file.

**#1018, patch `0028`.** `GeomPlate_BuildPlateSurface::G0Error()`, `G1Error()` and `G2Error()` return `myG0Error`/`myG1Error`/`myG2Error`, which have no in-class initialiser and no constructor that writes them. `VerifSurface()` is their only writer and `Perform()` reaches it only on the branch with at least one curve constraint; the point-only branch measures the same three deviations in `VerifPoints()` and throws them away, so a point-only plate reads uninitialised memory through all three accessors.

Proved by placement-new'ing the class over a `0x5A`-filled buffer: the pattern (`1.78388675173e+127` as a double) survives the constructor **and** a point-only `Perform()` in all three members, which is only possible if nothing wrote them.

The patch has four parts: in-class initialisers, `Perform()` clearing the three alongside the `myGeomPlateSurface.Nullify()` it already does on entry, the point-only branch keeping what `VerifPoints()` measured (`1.74210553841e-12` on a 25-point grid, not `0`), and `VerifPoints()` accumulating a maximum rather than overwriting, since the accessors document a maximum. The curve branch is deliberately untouched: `BRepFill_Filling::Build()` consumes `G0Error()` for a vertex tolerance and an approximation threshold, so folding point deviations into a mixed build would move geometry. The control confirms it does not: `2.08788369733e-05` curve-only and `0.00104645741524` curve-plus-points, identical before and after.

**#1022, patch `0029`.** `XCAFDoc_Datum::GetObject`'s `ChildLab_Pnt` block builds the point from `aLoc`, the annotation plane's array, instead of `aPnt`. `XCAFDoc_GeomTolerance::GetObject`'s otherwise identical block reads `aPnt`, so it is a one-character divergence. Two faces: a datum written with plane location `(6,6,6)` and point `(7,7,7)` reads back `(6,7,7)`, and a datum with a point and no plane dereferences a null handle, an uncatchable SIGSEGV. Each gets its own test, because a guard that stopped only the crash would leave the wrong value in place.

Reachability was measured rather than inherited from the report. A STEP import alone cannot produce the crashing shape (`STEPCAFControl_Reader` sets a datum plane and never a datum point). An OCAF load can, and the reproducer's section D proves it by saving the point-only datum to BinXCAF, reloading into a fresh `TDocStd_Application`, and crashing on the reload. Seven bridge functions route through `occtDocumentDatumObjectAt` since #1004 hoisted it, five of them write paths, and the helper calls `GetObject()` before any caller can look at the result, so the exposure is not read-side only.

Closes #1018
Closes #1022

## Prove the test fails

| Suite | Unpatched | Patched |
|---|---|---|
| `GeomPlate_BuildPlateSurface_Test.cxx`, unfiltered | 6 passed / 4 failed | 10 passed |
| `XCAFDoc_GDT_Test.cxx` | 9 passed / 1 failed, plus the crash case exiting 139 alone (gtest cannot catch a SIGSEGV, so it cannot share a process) | 11 passed, one unfiltered run |

Both transcripts are committed as `gtest.txt` in the two repro directories, and both harnesses compile the class out of `Libraries/occt-src`, applying the carried patch to a scratch copy for the AFTER lane, so neither depends on which branch the upstream checkout is sitting on. Each checks the GTest source by name for its new cases first, so a checkout on the wrong branch fails loudly instead of reporting a pristine result under an "AFTER" banner.

`0028`'s `Perform()`-entry reset was isolated on its own rather than trusted to the all-or-nothing run, since a four-part patch's all-or-nothing result cannot say which part a case is testing: with the rest applied and **only** that line removed, exactly one case fails, and it fails reporting `1.7421055384149805e-12`, the first build's own deviation, which is the stale-value mechanism rather than a coincidence.

Repo-side: full `swift test` green, 5,748 tests in 1,485 suites; all eight static gates and their `--self-test`s green; `swift build` clean.

## Upstream

- [OCCT#1481](https://github.com/Open-Cascade-SAS/OCCT/pull/1481) for `0028`
- [OCCT#1483](https://github.com/Open-Cascade-SAS/OCCT/pull/1483) for `0029`

Both went straight to a PR with no companion issue, per `okf/policies/upstream-occt-style.md`. Neither file has drifted: `GeomPlate_BuildPlateSurface.cxx`/`.hxx` and `XCAFDoc_Datum.cxx` are byte-identical between the pinned `V8_0_1` tree and upstream `master` at `7d2efad9c`, checked rather than assumed.

## Neither patch is in the pinned kernel

`ls Scripts/patches/*.patch | wc -l` now answers **19** against `Package.swift`'s list of **17**. That is CLAUDE.md's #585 shape and it is deliberate, so both `Package.swift` and `CLAUDE.md` enumerate the two and say how they differ rather than leaving the divergence to be rediscovered.

**One narrowing, measured on this PR rather than assumed.** The writeups originally said "exercised by no CI job at all", copying the general rule. That is true of `ci.yml`'s `build-and-test`, which resolves the pinned asset, but `kernel-integration.yml` triggers on `Scripts/patches/**`, so **this** PR does build `V8_0_1` plus all nineteen from source and run the full suite against that binary. What it proves is that both patches apply, compile and regress nothing. It cannot prove either fix works, because neither has a Swift-reachable assertion, and it does not run on any later PR that leaves `Scripts/patches/` alone. `Package.swift`, `CLAUDE.md` and both patch entries now say that rather than the blanket claim. None of it makes "read `kernel-integration.yml` instead of `ci.yml`" good advice; #585 is what discredited that.

The two patches still differ in what a rebuilt asset would buy:

- `0028` gains no Swift-side coverage from one either. `OCCTGeomPlateErrors`, the only bridge reader of the three accessors, was deleted by #999 (PR #1015), and `BRepFill_Filling`'s own forwarding of them is unreachable on the affected branch. The upstream GTests are its only coverage anywhere.
- `0029` is the opposite: an uncatchable SIGSEGV that nothing protects a consumer from until a rebuilt asset ships.

## The bridge guard for #1022 is filed, not shipped

**#1030.** A bridge-side guard is possible, and the reason it is not here is an ownership boundary rather than a judgement that it is wrong: `OCCTBridge_Document.mm` was held by a concurrent agent in this lane, so a guard written alongside `0029` would have raced their edits in the one file the guard belongs in.

#1030 carries the measurement in full: the guard must precede `GetObject()`, `XCAFDoc_Datum` exposes no `HasPlane()`, and `ChildLab_PlaneLoc`/`ChildLab_Pnt` are values of a file-local anonymous enum (tags `14` and `17` at this pin), so any guard hard-codes two private tags upstream can renumber with no compile-time signal here. It also records why `XCAFDoc_Datum::GetName()` is not a way around it.

### `GeomPlate_BuildPlateSurface` and `XCAFDoc_Datum` kernel patches carried (#1018, #1022)

Two upstream OCCT defects are now carried as `Scripts/patches/0028` and `0029`, and filed upstream as [OCCT#1481](https://github.com/Open-Cascade-SAS/OCCT/pull/1481) and [OCCT#1483](https://github.com/Open-Cascade-SAS/OCCT/pull/1483). Neither is in the pinned kernel asset, so `Scripts/patches/` holds nineteen against the asset's seventeen until a rebuild; `Package.swift` and `CLAUDE.md` name both and say what each costs.

`0028` (#1018): `GeomPlate_BuildPlateSurface::G0Error()`/`G1Error()`/`G2Error()` return uninitialised members after a `Perform()` whose constraints were all point constraints, because `VerifSurface()` is their only writer and the point-only branch measures the same deviations and discards them. The patch initialises the three, clears them when `Perform()` starts, keeps what the point-only branch measured, and makes `VerifPoints()` accumulate a maximum rather than overwrite, matching what the accessors document. This is the defect that made [#1015](https://github.com/SecondMouseAU/OCCTSwift/pull/1015) delete `Surface.plateErrors` rather than repair it; nothing in the Swift API reads those accessors today.

`0029` (#1022): `XCAFDoc_Datum::GetObject` builds the datum point's X from the annotation plane's array instead of the point's own, a wrong answer when a datum has both and an uncatchable SIGSEGV when it has a point and no plane. Seven bridge functions reach it, five of them write paths, for any OCAF document whose datum carries a point without an annotation plane. A STEP import cannot produce that shape; an OCAF load can. A bridge-side guard, needed until a rebuilt kernel ships, is tracked as [#1030](https://github.com/SecondMouseAU/OCCTSwift/issues/1030).

`docs/API_REFERENCE.md` drops `Surface.plateErrors` from the `Surfaces` row, deleted by #999 (PR #1015) and missed there because the gate does not read bare table text.

## SemVer impact

NONE. No public Swift API changes. Both carried patches are kernel-side and neither is in the pinned asset, so no consumer behaviour moves until a kernel rebuild is pinned. `docs/API_REFERENCE.md`'s corrected `Surfaces` row documents a deletion that already shipped in #1015.

## Checklist

- [x] New or changed behavior is covered by a unit test in the same PR. The tests are upstream GTests, since both fixes are kernel-side and neither has a Swift-reachable path this repo can assert against: `0028`'s only bridge reader was deleted by #999, and `0029`'s trigger cannot be authored by this package's write surface. Both GTest suites are run both ways by committed harnesses.
- [x] Every new test was run once with its subject broken, and the failure is reported above. `0028`'s entry reset was additionally isolated on its own.

## Notes for the reviewer

**Pre-PR review findings carried forward, with what was measured.**

- **`std::max` without `<algorithm>` in `0028`.** Checked twice rather than fixed: `GeomPlate_BuildPlateSurface.cxx` already uses `std::max`/`std::min` at four pre-existing sites (`:330`, `:331`, `:2419`, `:2541`), so the patch introduces no new dependency and adding the include would be noise in an upstream diff.
- **`0028` reports `0` for the three accessors after an early `Perform()` return.** `0` is not a measurement, which is the `census-unmeasured-values.py` shape this same patch's part 3 exists to avoid. It is defensible here for a reason that was checked: on every path that leaves the three at `0`, `Surface()` is null too, because `Perform()` nullifies it on the same entry and every early return precedes the `myGeomPlateSurface = new GeomPlate_Surface(...)` assignment. A `UserBreak` on a later loop iteration leaves the previous iteration's `VerifSurface()` values rather than the reset's. So "no result" stays observable, and the alternative (a sentinel) would change the accessors' return contract for every existing caller. Written up in the `0028` entry.
- **Four in-repo copies of the carried-patch list had drifted**, two of them badly (`docs/occt-upgrades.md` seven patches stale, `okf/references/carried-occt-patches.md` six). Both are brought current, and the okf table now enumerates all **five** statements of the set and what each one answers, including `docs/CHANGELOG.md`'s released-version header, which describes the asset that shipped with v3.0.0 and is correct as written even though the tree has moved past it. That header is deliberately not touched.
- **`docs/API_REFERENCE.md`'s category-row sum** now derives 3,324 against the 3,325 stated at `:28`. That prose figure predates this PR and belongs with whoever owns the row totals; `count-operations.py` prints the sum but never compares it, so the gate is green either way. Flagging rather than editing, because the surrounding text is another change's.

**Why one PR and not two.** Both patches edit `Package.swift`, `CLAUDE.md`, `Scripts/patches/README.md`, `docs/occt-upgrades.md` and `okf/references/carried-occt-patches.md`, all of which state the carried set as a whole. Two PRs would conflict on all five and would each state a count the other invalidates.


gsdali:
**Kernel-integration result, now measured rather than predicted.**

The PR body says `kernel-integration.yml` builds `V8_0_1` plus all nineteen carried patches from source and runs the suite against that binary. It has now done so, on a confirmed cache miss:

| Job | Result |
|---|---|
| `build patched OCCT.xcframework (macOS)` | success |
| `swift test against patched kernel (macOS)` | success, `Test run with 5748 tests in 1485 suites passed after 275.370 seconds` |

5,748 / 1,485 is byte-for-byte the same count as the local run against the **unpatched** pinned kernel, so `0028` and `0029` apply, compile, and regress nothing.

What this does not do is unchanged from the body: neither fix has a Swift-reachable assertion, so a green suite here cannot confirm either one works. `0028`'s only bridge reader was deleted by #999 (PR #1015), and `0029`'s trigger cannot be authored by this package's write surface. The upstream GTests remain the only coverage of the fixes themselves, and this job does not run on any later PR that leaves `Scripts/patches/` alone.

All seven checks are green: `gate-scripts` (required), `code-style`, `code-structure`, `swift build (iOS Simulator, smoke)`, `swift build + test (macOS)`, Kilo Code Review (no issues), and the kernel-integration pair above.

PR #1039, opened 2026-08-20
fix(#1036, #1037, #1038): the perspective eye anchor, five unchecked enum casts, and a setter that persisted nothing

### LINKED ISSUE #1036: .perspective(focus:) anchors the eye at the world origin, and three regimes past it return a silently wrong drawing
`.perspective(focus:)` anchors the eye at the world origin, and three regimes past it return a silently wrong drawing.

Found by the Pass 4a review panel against merged #999 work. Before #999 no caller could reach `.perspective` at all, since the projection type was discarded, so this is new surface.

## Where

- `Sources/OCCTSwift/Drawing.swift`, `ProjectionType.perspective(focus:)`
- `docs/reference/Drawing.md`, the `perspective(focus:)` section
- `Sources/OCCTBridge/src/OCCTBridge_Modeling.mm`, `OCCTDrawingCreate`

## The anchor the docs do not name

`OCCTDrawingCreate` builds `gp_Ax2 projAxis(gp_Pnt(0, 0, 0), viewDir)` unconditionally, and
`HLRAlgo_Projector::Project` divides by `R = 1 - Z/focus` with `Z` the point's coordinate in that
frame. So the eye sits at `focus * direction` measured from the **world origin**, the picture plane
passes through the origin, and there is no parameter that moves either.

The Swift doc says the eye is `focus` units "back along the view direction" and that "a projection
nearer than the shape's own extent along the view direction produces an empty drawing". The
threshold is not the shape's own extent, it is the shape's coordinates relative to the origin, and
what it produces is not an empty drawing.

## The three wrong-answer regimes

Measured against the pinned kernel with `Scripts/repro/1030-perspective-eye-anchor/probe.mm`, on a
10-unit cube spanning x `[20, 30]` viewed down +Z, so the sign of the projected x exposes mirroring
that a bounding-box width cannot:

| shape z | focus | projected x | verdict |
|---|---|---|---|
| `[0, 10]` | 50 | `[20.000, 37.500]` | correct |
| `[1000, 1010]` | 50 | `[-1.579, -1.042]` | **mirrored through the origin, 19x under-scaled, non-nil** |
| `[0, 10]` | 5.5 | `[-24.444, 20.000]` | **eye plane cuts the shape, one half mirrored against the other** |
| `[0, 10]` | 10 | `[-1.126e17, 20.000]` | **eye exactly on a face, R is 0, coordinates of order 1e17** |
| `[1000, 1010]` | 2000 | `[40.000, 60.606]` | correct |

All three bad rows return a real `Drawing` with a plausible edge count. Only the sign separates the
mirrored answer from the correct one, so nothing at the call site reads as failure.

Trigger:

```swift
Drawing.project(
    Shape.box(origin: SIMD3(-5, -5, 1000), width: 10, height: 10, depth: 10)!,
    direction: SIMD3(0, 0, 1),
    type: .perspective(focus: 50))
```

returns a non-nil `Drawing` that is mirrored and 19x under-scaled, while satisfying the documented
rule.

The bridge already rejects a non-positive focus, and its comment says why: a negative focus "returns
a real but differently-scaled projection, neither reads as failure at the call site". The identical
hazard on the far side of the eye is neither guarded nor documented.

## Second defect, same guard

The guard comment at `OCCTBridge_Modeling.mm` cites "focus 0/1e-12/5/15 all return an empty
VCompound" as its justification, but 5 and 15 are positive and both are accepted by the test beside
it. The guard is right (it also catches NaN); the evidence line is not. What focus 5 and 15 actually
produce on that fixture is the straddling case in row 3 above.

## Suggested resolution

Reject the `R <= 0` regime outright and rewrite both docs to name the real anchor. Rejecting matches
the existing `focus > 0` guard's own stated rationale, and points at or beyond a pinhole are not
projectable onto a picture plane in front of it under any reading. Anchoring the eye somewhere the
caller controls is a separable API question: the measurements show the foreshortening ratio is
already translation invariant (1.25 for the same cube at z 0 and at z 1000 with the eye 40 units
away in both cases), so only the uniform scale follows the shape's world position, and no
perspective information is lost by the origin anchor.

### LINKED ISSUE #1037: Five GD&T setters cast a raw int32_t into an OCCT enum with no range check
Five GD&T setters cast a raw `int32_t` into an OCCT enum with no range check, while every neighbour added in the same commits validates

Found by the Pass 4a review panel against merged #1004 work.

## Where

All in `Sources/OCCTBridge/src/OCCTBridge_Document.mm`:

| function | cast |
|---|---|
| `OCCTDocumentSetDimensionClassOfTolerance` | `(XCAFDimTolObjects_DimensionFormVariance)formVariance` |
| `OCCTDocumentSetDimensionClassOfTolerance` | `(XCAFDimTolObjects_DimensionGrade)grade` |
| `OCCTDocumentSetDimensionModifiers` | `(XCAFDimTolObjects_DimensionModif)modifiers[i]` |
| `OCCTDocumentSetGeomToleranceModifiers` | `(XCAFDimTolObjects_GeomToleranceModif)modifiers[i]` |
| `OCCTDocumentSetDatumModifiers` | `(XCAFDimTolObjects_DatumSingleModif)modifiers[i]` |

## What happens

`OCCTDocumentSetDimensionModifiers(doc, 0, (int32_t[]){9999}, 1)` writes 9999 into the document and
reads it back. `DimensionModifier(rawValue:)` then fails on the way out, so the Swift reader reports
a sequence that has silently lost its element rather than a call that was refused.

`SetClassOfTolerance` is worse than storage: an out-of-range `formVariance` also makes
`IsDimWithClassOfTolerance()` true, because that predicate is a bare `!= _None` test. So the reader
reports `hasClassOfTolerance == true` with `classOfTolerance == nil`, which is a dimension carrying a
tolerance class nobody can name.

Enum ranges at the pinned kernel: `DimensionModif` 0 to 23 (`_Between`), `DimensionFormVariance` 0
to 28 (`_ZC`), `DimensionGrade` 0 to 19 (`_IT18`), `GeomToleranceModif` 0 to 16 (`_All_Over`),
`DatumSingleModif` 0 to 21 (`_Translation`).

## Why this is an inconsistency, not a missing convention

Every neighbouring setter added in the same commits already validates, with the identical
pre-`try` shape:

- `OCCTDocumentSetDimensionQualifier`
- `OCCTDocumentSetDimensionAngularQualifier`
- `OCCTDocumentSetGeomToleranceTypeOfValue`
- `OCCTDocumentSetGeomToleranceMaterialRequirement`
- `OCCTDocumentSetDatumModifierWithValue`
- `OCCTDocumentSetDatumTarget`

and `occtMapSubShapes` in `Sources/OCCTBridge/include/OCCTBridge_Internal.h` states the convention
explicitly. So this is five sites inside one commit family that did not follow the rule the rest of
the family follows.

## Note for whoever fixes it

Not reachable from Swift. `DimensionModifier`, `FormVariance`, `Grade`, `GeomToleranceModifier` and
`DatumModifier` are complete typed Swift enums, so no Swift caller can express an out-of-range
value. The defect lives at the C boundary and has to be tested there, through `import OCCTBridge`.

For the array setters, validate the whole array before storing any of it, so a rejected call leaves
the previous sequence intact rather than half-written.

### LINKED ISSUE #1038: OCCTDocumentSetDatumTargetPlacement returns true for a call that persists nothing
`OCCTDocumentSetDatumTargetPlacement` returns true for a call that persists nothing

Found by the Pass 4a review panel against merged #1004 work.

## Where

- `Sources/OCCTBridge/src/OCCTBridge_Document.mm`, `OCCTDocumentSetDatumTargetPlacement`
- `Sources/OCCTBridge/include/OCCTBridge_Document.h`, its declaration
- `Sources/OCCTSwift/GDTWrite.swift`, `setDatumTargetPlacement(at:location:normal:reference:length:width:)`
- `docs/reference/Annotation.md`, the same entry

## The precondition nothing records

`XCAFDoc_Datum::SetObject` nests the whole axis/length/width/number block inside
`if (theObject->IsDatumTarget())`, and inside that takes an `Area` branch that stores the target's
own shape via `TNaming_Builder` and writes no placement at all. The bridge sets the axis, length and
width on the in-memory object and then calls `SetObject`, which discards every one of them in two
cases:

1. the datum has had no `setDatumTarget(at:type:number:)` applied, so `IsDatumTarget()` is false
2. the datum's target type is `.area`

Both return `true` today. In case 1 the caller is told the placement was stored and then reads back
`datum(at:)?.target == nil`; in case 2 it reads back a target whose `length` and `width` are both
`nil`.

Neither the bridge header nor the Swift doc records the precondition. The Swift doc's own example
happens to call `setDatumTarget` first, so the ordering is load-bearing but undocumented.

## Not the third precondition it looks like

`SetObject`'s non-`Area` branch is additionally gated on `theObject->HasDatumTargetParams()`, which
looks like a third way to fail. It is not reachable from this bridge function:
`SetDatumTargetAxis`, `SetDatumTargetLength` and `SetDatumTargetWidth` each set `myIsValidDT = true`
as a side effect, so the bridge's own three calls satisfy it before `SetObject` runs. That side
effect is the reason the three are one bridge call in the first place.

## Suggested resolution

Return `false` for both cases rather than reporting success for a call that persisted nothing, and
record the precondition in the bridge header, the Swift doc comment and `docs/reference/Annotation.md`.
The object fetched by `occtDocumentDatumObjectAt` is `outAttr->GetObject()`, so it already reflects
the stored state and the precondition is checkable in the bridge without any new API.

gsdali:
## What & why

Three findings from the Pass 4a review panel, all against code #999 and #1004 landed, none of which
had an issue before this branch. They are one PR because they are one review's output on one branch,
not because they share a mechanism.

**#1036, the perspective eye anchor.** `OCCTDrawingCreate` builds `gp_Ax2 projAxis(gp_Pnt(0, 0, 0),
viewDir)` unconditionally, so the eye sits at `focus * direction` measured from the world origin and
`HLRAlgo_Projector::Project` divides by `R = 1 - Z/focus` with `Z` in that same frame. A shape at or
beyond the eye has `R <= 0` and came back mirrored through the origin rather than refused. The docs
named the wrong anchor (the shape) and the wrong threshold (the shape's own extent), and said the
failure mode was an empty drawing, which it is not.

**#1037, five unchecked enum casts.** Five GD&T setters turned a caller-supplied `int32_t` straight
into an OCCT enum, so a value naming no enumerator was stored and read back verbatim. Every
neighbouring setter added in the same commits already validated.

**#1038, a setter that persisted nothing.** `OCCTDocumentSetDatumTargetPlacement` returned `true`
for a call `XCAFDoc_Datum::SetObject` discards outright.

Closes #1036
Closes #1037
Closes #1038

### The decision on #1036's regime, and why

The issue offered three routes: reject the regime, anchor the eye where the caller controls, or
document it precisely. This PR does the **first and the third, and deliberately not the second.**

Measured first, with `Scripts/repro/1036-perspective-eye-anchor/probe.mm`, on a 10-unit cube
spanning x `[20, 30]` viewed down +Z. An x-offset box is the fixture rather than a centred one
because a mirrored projection of a centred box has the same bounding-box width as a correct one, and
only the sign separates them:

| shape z | focus | projected x | verdict |
|---|---|---|---|
| `[0, 10]` | 50 | `[20.000, 37.500]` | correct |
| `[1000, 1010]` | 50 | `[-1.579, -1.042]` | mirrored through the origin, 19x under-scaled |
| `[0, 10]` | 5.5 | `[-24.444, 20.000]` | eye plane cuts the shape, one half mirrored against the other |
| `[0, 10]` | 10 | `[-1.126e17, 20.000]` | eye exactly on a face, `R` is 0 |
| `[0, 10]` | 10.0001 | `[20.000, 3000030.000]` | correct, and extreme |
| `[1000, 1010]` | 2000 | `[40.000, 60.606]` | correct |

Two of those rows are regimes the review had not found. The `focus 1005 -> NULL` row in the review's
own table turned out to be a coincidence of that fixture rather than a refusal: the same straddling
geometry on an offset box returns a real, garbled drawing.

**Rejecting is right because `R <= 0` is not a differently-scaled answer, it is not an answer.**
Points at or beyond a pinhole cannot be projected onto a picture plane in front of it; every
renderer clips them. The measurements show what comes back instead is mirrored coordinates or
coordinates of order 1e17, both reported as success with a plausible edge count. That is the same
hazard the existing `focus > 0` guard already rejects, in the same function, for the reason its own
comment gives, so leaving the far side unguarded was an inconsistency rather than a policy.

**Anchoring the eye is not needed for correctness, which is why it is not here.** The obvious worry
is that an origin-anchored eye distorts a shape modelled far from the origin. Measured, it does not:
the same cube at z `[0, 10]` with focus 50 and at z `[1000, 1010]` with focus 1050, both with the
eye 40 units from the near face, have an identical foreshortening ratio of 1.25 and differ only by a
uniform factor of 21 in scale. No perspective information is lost or bent by the anchor; a uniform
scale is what a drawing's scale factor is for, and translating the shape to the origin recovers it
exactly. Adding an eye-position parameter would be new public API that fixes no wrong answer, so it
is a separable enhancement rather than part of this correctness fix. `docs/reference/Drawing.md` now
states the anchor, both consequences and the workaround, so the choice is discoverable rather than
implicit.

**The guard bounds the shape by its axis-aligned bounding box**, which is an upper bound on its reach
along the view direction. It errs toward refusing a marginal call, never toward admitting a wrong
one, and that is documented. Measured at the transition on a box with its top face at z = 10: focus
9.9999 and 10 are refused (both garbled), 10.0001 and up are accepted (all correct). The boundary
lands exactly where the arithmetic says it should.

### #1036's second defect, the guard's own evidence line

The guard comment cited "focus 0/1e-12/5/15 all return an empty VCompound" as justification for a
test that only rejects `focus <= 0`. 5 and 15 are positive and both pass that test, so they were
never evidence for it.

**The measurements themselves were right, and my first attempt at fixing this was wrong.** I
rewrote the line to say focus 0 and 1e-12 return a *mirrored* projection, then measured that claim
on the comment's own fixture before pushing it, and it is false: on a 100x50x30 centred box all four
of 0, 1e-12, 5 and 15 do return an empty `VCompound`, exactly as written. What was wrong was the
inference, not the data. That box spans z `[-15, 15]`, so focus 5 and 15 put the eye inside it or on
its face, which means those two rows were measuring the straddling case rather than the
non-positive-focus case. The comment now says so, and the list is trimmed to the two values the test
actually rejects. `Scripts/repro/1036-perspective-eye-anchor/guard_comment_probe.mm` is the
re-measurement, kept so the next reader does not have to take either version on trust.

This is also why the review's own `focus 1005 -> NULL` row was a coincidence of its fixture: whether
a straddling eye plane returns an empty compound or a real, garbled one depends on the shape, and
both happen. The offset-box fixture in the table above returns the garbled one.

### `.perspective(focus:)` refuses a shape at or beyond the eye, and the docs name the real anchor (#1036)

`Drawing.project(_:direction:type:)` used to return a real, mirrored `Drawing` for a shape sitting
past the perspective eye point, and the documentation described an anchor and a threshold that were
both wrong.

`OCCTDrawingCreate` builds its projection frame as `gp_Ax2(gp_Pnt(0, 0, 0), viewDir)`
unconditionally, so the eye is at `focus * direction` measured from the **world origin**, the picture
plane passes through the origin, and `HLRAlgo_Projector::Project` divides by `R = 1 - Z/focus` with
`Z` in that frame. Any point with `Z >= focus` therefore has `R <= 0` and was drawn mirrored through
the origin instead of not at all. Measured on a 10-unit cube spanning x `[20, 30]`, z `[1000, 1010]`
viewed down +Z: focus 50 returned seven edges spanning x `[-1.579, -1.042]` where the correct answer
is `[40, 60.606]`. Two further regimes behaved the same way: an eye plane cutting the shape returned
one half mirrored against the other (x `[-24.444, 20]`), and an eye landing exactly on a face
returned coordinates of order 1e17. All three reported success with a plausible edge count.

`.perspective` now returns `nil` whenever the shape reaches the eye plane, matching the refusal the
non-positive-focus guard beside it already made for the same stated reason. Correct regimes are
unchanged, including the extreme ones: a focal distance just clear of the shape still projects, and
so does a shape far behind the picture plane.

```swift
let far = Shape.box(origin: SIMD3(-5, -5, 1000), width: 10, height: 10, depth: 10)!

// Was a mirrored, 19x under-scaled Drawing. Now nil.
Drawing.project(far, direction: SIMD3(0, 0, 1), type: .perspective(focus: 50))

// The eye has to clear the shape's reach from the ORIGIN, not from the shape.
Drawing.project(far, direction: SIMD3(0, 0, 1), type: .perspective(focus: 2000))
```

`Drawing.ProjectionType.perspective(focus:)` and `docs/reference/Drawing.md` now state where the eye
actually sits, that the drawing's scale follows the shape's world position while its foreshortening
ratio does not, and that translating the shape to the origin is how to make the scale independent of
where the part was modelled.

### Five GD&T setters reject values that name no enumerator (#1037)

`OCCTDocumentSetDimensionClassOfTolerance` (two casts),
`OCCTDocumentSetDimensionModifiers`, `OCCTDocumentSetGeomToleranceModifiers` and
`OCCTDocumentSetDatumModifiers` cast a caller-supplied `int32_t` straight into an OCCT GD&T enum, so
`OCCTDocumentSetDimensionModifiers(doc, 0, (int32_t[]){9999}, 1)` wrote 9999 into the document and
read it back. An out-of-range `formVariance` was worse than storage: it also made
`IsDimWithClassOfTolerance()` true, since that predicate is a bare `!= _None` test, so the reader
reported `hasClassOfTolerance == true` alongside `classOfTolerance == nil`.

All five now range-check, matching the six neighbouring setters added in the same commits that
already did. For the three array setters the whole array is validated before any of it is stored, so
a rejected call leaves the previous sequence intact rather than half-written. Not reachable from
Swift, whose GD&T enums are complete and typed; this closes the C boundary.

### `setDatumTargetPlacement` refuses a call that would persist nothing (#1038)

`XCAFDoc_Datum::SetObject` nests its whole axis/length/width/number block inside
`if (theObject->IsDatumTarget())`, and inside that takes an `Area` branch that stores the target's
own shape rather than a placement. So calling
`setDatumTargetPlacement(at:location:normal:reference:length:width:)` on a datum that has had no
`setDatumTarget(at:type:number:)` applied, or whose target type is `.area`, set every value on the
in-memory object and then discarded all of them, while returning `true`.

Both cases now return `false`. The precondition is recorded on the bridge declaration, the Swift doc
comment and `docs/reference/Annotation.md`, and the Swift doc's example now shows the required
ordering explicitly.

```swift
doc.setDatumTarget(at: idx, type: .rectangle, number: 1)   // required first
doc.setDatumTargetPlacement(
    at: idx, location: SIMD3(1, 2, 3), normal: SIMD3(0, 0, 1), reference: SIMD3(1, 0, 0),
    length: 30, width: 18)
```

## SemVer impact

MINOR for #1036, PATCH for #1037 and #1038.

#1036 narrows what `Drawing.project(_:direction:type:.perspective(focus:))` accepts: a call whose
shape sits at or beyond the eye now returns `nil` where it previously returned a `Drawing`. No
consumer can have been relying on the old result deliberately, since it was a mirrored projection
reported as a success, but it is a behaviour change a consumer can observe rather than a pure fix,
so it is called MINOR rather than PATCH. Migration: pass a `focus` larger than the shape's reach
along `direction` measured from the world origin, or translate the shape toward the origin first.
Nothing about the accepted regime moved, and no signature changed.

#1037 and #1038 are PATCH: both only turn a previously-successful call that stored nothing (or
stored nonsense) into a `false` return, and neither is reachable through a correct call. #1037 is
not reachable from Swift at all, since the Swift enums are complete and typed.

## Checklist

      verification).
- [x] Every new test was run once with its subject broken, and the failure is reported below.

### The removal matrix

Four injections, each stating which mechanism it isolates. The two #1036 rows fail **disjoint** sets,
which is what distinguishes isolation from coincidence: neither row's failures overlap the other's.

| row | injection | isolates | tests failed | tests passed |
|---|---|---|---|---|
| A | the `#1036` reach guard deleted from `OCCTDrawingCreate` | the eye-plane refusal | 3/6: beyond-the-eye, cutting-plane, exactly-on-a-face | 3/6: anchor, behind-picture-plane, extreme-focus |
| B | `projAxis` anchored on the shape's bbox centre instead of `gp_Pnt(0,0,0)`, guard left in place | the world-origin anchor, which row A does not touch | 3/6: anchor, behind-picture-plane, extreme-focus | 3/6: beyond-the-eye, cutting-plane, exactly-on-a-face |
| C | all five `#1037` range checks reverted to bare casts | the enum range rejection | 4/5, one per setter | 1/5: the `count 0` clearing control |
| D | the `#1038` precondition deleted | the datum-target precondition | 3/4 | 1/4: the rectangle-target control |

Row B is the row that matters most, because the anchor test asserts on geometry that the unfixed
code already produced, so row A alone would leave it unproven. Under row B all three of its numeric
assertions move: half-width 6.25 becomes 6.944, 131.25 becomes 5.024, and the ratio 21 becomes 0.90.
It is not vacuous.

Rows C and D each keep a passing control (the `count 0` clear, and the correctly-ordered rectangle
target), so neither guard passes by refusing everything.

### Test counts, both measured

| | tests | suites |
|---|---|---|
| `origin/refactor/385-pass4a` at `f0857b7f` (baseline) | 5750 | 1485 |
| this branch | 5765 | 1488 |

Both green, both run locally from source with `OCCTSWIFT_LOCAL=1` and the prebuilt-bridge
environment variables cleared, so the bridge `.mm` edits are actually compiled rather than resolved
from the published `OCCTBridge.xcframework`. The baseline was run in its own worktree at the base
commit rather than derived by subtraction. The delta is exactly the 15 new tests in 3 new suites.

### Two of my own arithmetic errors, caught by the tests

Worth recording because both produced a plausible wrong number from a real measurement, which is the
shape `okf/policies/measure-dont-assume.md` warns about:

- The behind-the-picture-plane test asserted `min == 20/21`, taking `R` from the near face. The two
  faces divide by different `R` and the extremes come from opposite ones, so the correct value is
  `20/21.2`. The test failed by 0.009 and the code was right.
- The class-of-tolerance test asserted grade ordinal 7 reads back as `.it7`. `XCAFDimTolObjects_DimensionGrade`
  opens with `IT01` and `IT0`, so ordinal 7 is `IT6`. The reader was right.

Both are now stated in the test comments with the reasoning, rather than being silently corrected.

### What is deliberately not here

An eye-position parameter for `.perspective`. The measurement in "What & why" is the argument: the
origin anchor costs a uniform scale factor and nothing else, and adding public API that fixes no
wrong answer belongs in its own issue with its own case, not smuggled into a correctness fix.

`GetStatus()`-style partial reporting for #1038. The bridge returns `false`, which is the only
signal any caller in this tree reads.

### Probes

`Scripts/repro/1036-perspective-eye-anchor/` holds three ground-truth probes, a README and a
transcript for each, matching the convention in the neighbouring repro directories:

- `probe.mm` reproduces the review's table and the two regimes it missed, and takes the
  translation-invariance measurement that argued against a caller-controlled eye position.
- `guard_probe.mm` measures where the bounding-box guard's boundary lands, and checks a shape behind
  the picture plane, a rotated box and a sphere.
- `guard_comment_probe.mm` re-measures the four values the pre-existing guard comment cited, which is
  what caught my own wrong correction to it.

All three replicate the bridge's own projector construction rather than calling the bridge, so they
measure OCCT rather than the wrapper, and all three compile per CLAUDE.md's ground-truth recipe.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,12 +60,12 @@ Scripts/tsan-stress.sh all           # ThreadSanitizer gate: REQUIRED for concur
 
 ### Static Gate Scripts
 
-Seven gates, two censuses and one merge-history audit, all pure Python over the repo's own text. No
+Eight gates, two censuses and one merge-history audit, all pure Python over the repo's own text. No
 OCCT, no build, no network, ~3s for the lot. **CI runs every gate, plus every `--self-test` including the census's, in `ci.yml`'s
 `gate-scripts` job** — a separate
 `ubuntu-latest` job, not a step inside the macOS build, so it reports in under a minute and keeps
 its own status check when `build-and-test` is red for an unrelated reason. Each exits 1 on a
-defect and 0 when clean; `check-bridge-index`, `check-null-handle-guards` and
+defect and 0 when clean; `check-bridge-index`, `check-null-handle-guards`, `derive-gdt-enums` and
 `derive-bridge-header-split` exit **2** if run from anywhere but the repo root (#625).
 
 **`gate-scripts` is a required status check on `main`** (#649), via a repository ruleset (id
@@ -117,6 +117,7 @@ python3 Scripts/check-docs-defaults.py           # every default docs/reference/
 python3 Scripts/check-docs-existence.py          # every symbol docs/ documents as current still exists in Sources (#802)
 python3 Scripts/check-borrowed-handles.py        # no struct/enum stores an OCCT*Ref it has no deinit to release (#965)
 python3 Scripts/derive-bridge-header-split.py --verify  # every declaration sits in the header its .mm owns (#673)
+python3 Scripts/derive-gdt-enums.py --verify      # the GD&T enums still match the pinned XCAFDimTolObjects headers (#996)
 python3 Scripts/count-operations.py              # README + API_REFERENCE totals match the derived count
 python3 Scripts/census-unmeasured-values.py      # CENSUS, not a gate: values returned as measurements that were never computed (#726)
 python3 Scripts/census-doc-occt-attribution.py   # CENSUS, not a gate: docs attributing a method to an OCCT class its bridge fn never reaches (#928)
@@ -141,7 +142,7 @@ hand-adjudicated sample** (`Scripts/repro/928-over-coverage-detector/`, re-scora
 `score_sample.py`). That rate is why it reports rather than gates; promoting it is a separate
 decision on a better number, and it comes with a rename to `check-` per the convention above.
 
-Six of the seven gates take `--self-test`, as do both censuses and the merge-history audit, each running a fixture battery proving the *detector* catches each
+Seven of the eight gates take `--self-test`, as do both censuses and the merge-history audit, each running a fixture battery proving the *detector* catches each
 failure mode. Run it whenever you change one of these scripts — three gate scripts on this branch
 were confidently wrong (#618, #624/#630, #626), and a detector that reports "all clear" because it
 is blind looks exactly like one reporting "all clear" because the tree is clean.
@@ -149,7 +150,7 @@ is blind looks exactly like one reporting "all clear" because the tree is clean.
 running the report, so writing `count-operations.py --self-test` to match its siblings fails loudly
 instead of passing forever.
 
-**Optional pre-commit hook** (`Scripts/git-hooks/pre-commit`) runs sixteen of CI's seventeen
+**Optional pre-commit hook** (`Scripts/git-hooks/pre-commit`) runs eighteen of CI's nineteen
 invocations, flag for flag. The one it omits is `check-changelog-transcription.py`'s real run, which
 audits the branch's merge history and so answers a question about the branch rather than about the
 commit you are making; its `--self-test` does run. That is the only deliberate divergence between
--- a/Scripts/repro/1030-perspective-eye-anchor/guard_probe.mm
+++ b/Scripts/repro/1030-perspective-eye-anchor/guard_probe.mm
@@ -1,0 +1,155 @@
+// Second probe for OCCTSwift #1030: what a bounding-box guard would decide, and how much slack
+// Bnd_Box's own gap adds. The guard under test rejects a perspective projection whose shape reaches
+// the eye plane, i.e. max over the shape of (P . viewDir) >= focus.
+
+#include <BRepBndLib.hxx>
+#include <BRepBuilderAPI_Transform.hxx>
+#include <BRepPrimAPI_MakeBox.hxx>
+#include <BRepPrimAPI_MakeSphere.hxx>
+#include <Bnd_Box.hxx>
+#include <HLRAlgo_Projector.hxx>
+#include <HLRBRep_Algo.hxx>
+#include <HLRBRep_HLRToShape.hxx>
+#include <TopExp.hxx>
+#include <TopExp_Explorer.hxx>
+#include <TopTools_IndexedMapOfShape.hxx>
+#include <TopoDS.hxx>
+#include <TopoDS_Shape.hxx>
+#include <TopoDS_Vertex.hxx>
+#include <gp_Ax2.hxx>
+#include <gp_Dir.hxx>
+#include <gp_Pnt.hxx>
+#include <gp_Trsf.hxx>
+
+#include <BRep_Tool.hxx>
+#include <cstdio>
+
+// The support function of the axis-aligned bounding box in direction d: the largest value of
+// (P . d) over the box, which bounds the same quantity over the shape inside it.
+static double bboxSupport(const TopoDS_Shape& shape, const gp_Dir& d, bool& ok, double& gap)
+{
+  Bnd_Box bb;
+  BRepBndLib::Add(shape, bb);
+  if (bb.IsVoid())
+  {
+    ok = false;
+    return 0.0;
+  }
+  ok = true;
+  gap = bb.GetGap();
+  double xmin, ymin, zmin, xmax, ymax, zmax;
+  bb.Get(xmin, ymin, zmin, xmax, ymax, zmax);
+  return (d.X() > 0 ? xmax * d.X() : xmin * d.X()) + (d.Y() > 0 ? ymax * d.Y() : ymin * d.Y())
+         + (d.Z() > 0 ? zmax * d.Z() : zmin * d.Z());
+}
+
+// The true maximum of (P . d) over the shape's vertices, for comparison on polyhedra.
+static double vertexMax(const TopoDS_Shape& shape, const gp_Dir& d)
+{
+  TopTools_IndexedMapOfShape verts;
+  TopExp::MapShapes(shape, TopAbs_VERTEX, verts);
+  double best = -1e300;
+  for (int i = 1; i <= verts.Extent(); i++)
+  {
+    gp_Pnt p = BRep_Tool::Pnt(TopoDS::Vertex(verts(i)));
+    double v = p.X() * d.X() + p.Y() * d.Y() + p.Z() * d.Z();
+    if (v > best)
+      best = v;
+  }
+  return best;
+}
+
+static void row(const char* label, const TopoDS_Shape& shape, const gp_Dir& d, double focus)
+{
+  bool   ok  = false;
+  double gap = 0.0;
+  double sup = bboxSupport(shape, d, ok, gap);
+  double vm  = vertexMax(shape, d);
+
+  double xmin = 0, xmax = 0;
+  int    nEdges = 0;
+  bool   built  = false;
+  try
+  {
+    gp_Ax2               projAxis(gp_Pnt(0, 0, 0), d);
+    HLRAlgo_Projector    projector(projAxis, focus);
+    Handle(HLRBRep_Algo) hlrAlgo = new HLRBRep_Algo();
+    hlrAlgo->Add(shape);
+    hlrAlgo->Projector(projector);
+    hlrAlgo->Update();
+    hlrAlgo->Hide();
+    HLRBRep_HLRToShape shapes(hlrAlgo);
+    TopoDS_Shape       visible = shapes.VCompound();
+    if (!visible.IsNull())
+    {
+      for (TopExp_Explorer e(visible, TopAbs_EDGE); e.More(); e.Next())
+        nEdges++;
+      if (nEdges > 0)
+      {
+        Bnd_Box bb;
+        BRepBndLib::Add(visible, bb);
+        double a, b, c, e2, f2, g2;
+        bb.Get(a, b, c, e2, f2, g2);
+        xmin  = a;
+        xmax  = e2;
+        built = true;
+      }
+    }
+  }
+  catch (...)
+  {
+  }
+
+  printf("%-34s focus=%-9g support=%11.5f vertexMax=%11.5f gap=%.2e guard=%s  built=%s",
+         label,
+         focus,
+         sup,
+         vm,
+         gap,
+         sup >= focus ? "REJECT" : "accept",
+         built ? "yes" : "no ");
+  if (built)
+    printf(" x=[%12.4f,%12.4f]", xmin, xmax);
+  printf("\n");
+}
+
+int main()
+{
+  gp_Dir dz(0, 0, 1);
+
+  printf("== a shape far BEHIND the picture plane (negative Z) is a correct, shrunken answer ==\n");
+  {
+    TopoDS_Shape s = BRepPrimAPI_MakeBox(gp_Pnt(20, -5, -1010), gp_Pnt(30, 5, -1000)).Shape();
+    row("z=[-1010,-1000] offset", s, dz, 50);
+  }
+
+  printf("\n== the boundary: box top at z=10, focus straddling it ==\n");
+  {
+    TopoDS_Shape s = BRepPrimAPI_MakeBox(gp_Pnt(20, -5, 0), gp_Pnt(30, 5, 10)).Shape();
+    row("z=[0,10] offset", s, dz, 9.9999);
+    row("z=[0,10] offset", s, dz, 10.0);
+    row("z=[0,10] offset", s, dz, 10.0001);
+    row("z=[0,10] offset", s, dz, 10.001);
+    row("z=[0,10] offset", s, dz, 10.01);
+    row("z=[0,10] offset", s, dz, 11.0);
+  }
+
+  printf("\n== conservatism: a rotated box, where the AABB overshoots the true extent ==\n");
+  {
+    gp_Trsf t;
+    t.SetRotation(gp_Ax1(gp_Pnt(0, 0, 0), gp_Dir(1, 0, 0)), 0.7853981633974483);
+    TopoDS_Shape s0 = BRepPrimAPI_MakeBox(gp_Pnt(20, -5, 0), gp_Pnt(30, 5, 10)).Shape();
+    TopoDS_Shape s  = BRepBuilderAPI_Transform(s0, t).Shape();
+    row("rotated 45deg about X", s, dz, 12.0);
+    row("rotated 45deg about X", s, dz, 20.0);
+  }
+
+  printf("\n== a curved shape, where vertices under-report and the bbox does not ==\n");
+  {
+    TopoDS_Shape s = BRepPrimAPI_MakeSphere(gp_Pnt(0, 0, 0), 10).Shape();
+    row("sphere r=10 at origin", s, dz, 9.5);
+    row("sphere r=10 at origin", s, dz, 10.5);
+    row("sphere r=10 at origin", s, dz, 30.0);
+  }
+  return 0;
+}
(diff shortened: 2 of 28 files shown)


gsdali:
Note on the red `Kilo Code Review` check, so nobody reads it as a review finding.

It is infrastructural, not a verdict on the diff:

```
Kilo Code Review failed
Review failed: Repository request was rate limited: termination nonzero exit, exit code 128
```

The bot completed twice earlier on this PR and posted `Status: No Issues Found | Recommendation: Merge`
both times, over the same eleven files. The only changes since its last completed run are the
`Scripts/repro/1036-perspective-eye-anchor/` README and transcripts, and two prose corrections in
`docs/reference/Drawing.md` (a sentence that had lost its verb, and a distance stated as 50 units
where the measurement is 40). No source file changed after that run.

Re-running the check is enough to clear it if a maintainer wants it green before merging.

Everything else is green on the final commit, including `gate-scripts`, which is the required one.

Issue #1187, opened 2026-08-27
Cosmetic thread minor-diameter formula duplicated between side-view and end-view annotation builders


gsdali:
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/DrawingThreadAnnotation.swift:42` — `DrawingAnnotation.cosmeticThreadSideView(axisStart:axisEnd:majorDiameter:pitch:callout:)`, local `let minorDiameter = max(majorDiameter - 1.0825 * pitch, majorDiameter * 0.8)`
- `Sources/OCCTSwift/DrawingThreadAnnotation.swift:87` — `DrawingAnnotation.cosmeticThreadEndView(centre:majorDiameter:pitch:)`, local `let minorDiameter = max(majorDiameter - 1.0825 * pitch, majorDiameter * 0.8)`

Both lines are byte-for-byte identical, including the comment above line 42 ("Minor diameter per ISO 68 (metric V thread): D_minor = D - 1.0825·P") which is not repeated at line 87, so the end-view copy carries no explanation of where its own 0.8-floor/1.0825 constants come from — a reader at line 87 has no way to know the two are meant to stay in lockstep without cross-referencing line 42 by hand.

## Divergence
[...]

Issue #1176, opened 2026-08-27
GDTSymbol's doc comment claims a round-trip with Document.GeomToleranceType that cannot exist and does not exist


gsdali:
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/DrawingSymbols.swift:111-117` — `GDTSymbol`, a `String`-backed enum (15 cases: `straightness, flatness, circularity, cylindricity, profileOfLine, profileOfSurface, perpendicularity, parallelism, angularity, position, concentricity, symmetry, coaxiality, circularRunout, totalRunout`), whose doc comment at line 111 reads: "Matches `Document.GeomToleranceType` raw values for easy round-trip from XDE into drawings."
- `Sources/OCCTSwift/GDTRead.swift:96-113` — `Document.GeomToleranceType`, an `Int32`-backed enum (16 cases: adds `none = 0`, and spells the circularity case `circularityOrRoundness = 3` rather than `circularity`), read from XCAF/STEP GD&T data via `OCCTDocumentGetGeomToleranceInfo`'s `info.type`.

## Divergence
Two verified, independent breaks in the doc comment's claim, not just a maintenance-cost note:

1. **The raw values cannot match at all, by construction.** `GDTSymbol: String` (line 112) has raw values like `"straightness"`/`"circularity"`; `GeomToleranceType: Int32` (line 96) has raw values `0`-`15`. A `String` raw value and an `Int32` raw value are different types and can never be equal or interchangeable — "matches ... raw values" is not approximately true, it is false on its face for anyone who reads both declarations.
2. **The case names have already drifted.** `GeomToleranceType` spells the case `circularityOrRoundness` (line 100, matching OCCT's own `XCAFDimTolObjects_GeomToleranceType` naming); `GDTSymbol` spells the corresponding case `circularity` (line 113). A caller trying to bridge the two by case name (e.g. `GDTSymbol(rawValue: someType.name)` or a switch mapping one to the other) breaks on this one case. `GeomToleranceType` also carries a `none = 0` case that `GDTSymbol` has no counterpart for at all.
[...]

Issue #1174, opened 2026-08-27
HLR bridge: `deflection` param doc omits the same "(must be > 0)" precondition that `focus`'s doc states, despite identical enforcement


gsdali:
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTBridge/include/OCCTBridge_HLR.h:33` — `@param focus` doc on `OCCTDrawingCreate`: `Focus distance for perspective projection (must be > 0)`
- `Sources/OCCTBridge/include/OCCTBridge_HLR.h:45` — `@param deflection` doc on `OCCTDrawingCreatePoly`: `Deflection for triangulation` (no precondition stated). Note: the original finding cited line 44 for this; line 44 is actually the `@param dirX, dirY, dirZ` line immediately above it — the `deflection` doc itself is line 45. Confirmed by direct read of the file; the divergence claim itself holds regardless of the off-by-one.
- `Sources/OCCTBridge/src/OCCTBridge_HLR.mm:65-66` — enforcement site for `focus`, inside `OCCTDrawingCreate`: `if (projectionType == OCCTProjectionPerspective && !(focus > 0)) return nullptr;`
- `Sources/OCCTBridge/src/OCCTBridge_HLR.mm:201-202` — enforcement site for `deflection`, inside `OCCTDrawingCreatePoly`: `if (deflection <= 0) return nullptr;`
- `Sources/OCCTSwift/Drawing.swift:362-387` — `Drawing.project(_:direction:type:)`, the Swift wrapper for `OCCTDrawingCreate`; its doc explicitly states "or nil if ... the perspective focal distance is not positive" (line 361), i.e. the precondition is documented a second time, correctly, at the Swift layer.
- `Sources/OCCTSwift/Drawing.swift:432-447` — `Drawing.projectFast(_:direction:deflection:)`, the Swift wrapper for `OCCTDrawingCreatePoly`; its doc says only "or nil if projection fails" (line 431), with no mention that a non-positive `deflection` is what triggers that `nil`. The same asymmetry that exists in the C header is reproduced at the Swift layer.

## Divergence
[...]

Issue #1171, opened 2026-08-27
`Shape.compound(from:)` in Section2D.swift silently fuses instead of grouping, shadowing `Shape.compound(_:)`


gsdali:
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/Section2D.swift:112` — `internal static func compound(from shapes: [Shape]) -> Shape?` (private-to-module helper, only call site is `Sources/OCCTSwift/Section2D.swift:77` inside `section2D(planeOrigin:planeNormal:planeU:deflection:)`)
- `Sources/OCCTSwift/Shape.swift:920` — `public static func compound(_ shapes: [Shape]) -> Shape?` (the public, documented grouping API)

Both are `static func … compound…(…) -> Shape?` on `Shape`, differing only by argument label (`from:` vs. none), so `Shape.compound(from: xs)` and `Shape.compound(xs)` read as two spellings of the same operation to anyone skimming call sites or autocomplete.

## Divergence
The two are not just differently named, they perform **opposite topological operations**:

- `Shape.compound(_:)` (Shape.swift:920) is documented "no boolean, just grouping" and calls the bridge function `OCCTShapeCreateCompound`, which builds a real `TopoDS_Compound` via `BRep_Builder::MakeCompound` + `builder.Add(...)` for every input shape — a pure container, no geometry is fused, cut, or otherwise altered, and overlapping/duplicate shapes stay exactly as given (this is what lets `Issue1026NullShapeTypeGuardTests`/`Issue979SubShapeIndexIdentityTests`/etc. reason about sub-shape counts and identity across the inputs unchanged).
[...]

PR #1194, opened 2026-08-27
fix(drawing): Sheet.inset A4 margins per ISO 5457 (#1168)

### LINKED ISSUE #1168: Sheet.inset returns A0-A3 margins for A4, contradicting its own ISO 5457 doc comment
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/DrawingSheet.swift:142-143` — `Sheet.inset`, `case .a0, .a1, .a2, .a3: return (left: 20, right: 10, top: 10, bottom: 10)`
- `Sources/OCCTSwift/DrawingSheet.swift:144-145` — `Sheet.inset`, `case .a4: return (left: 20, right: 10, top: 10, bottom: 10)` (verbatim copy of the arm above; should be `(7, 7, 7, 10)` per the doc comment two lines above at 138-139)

Both arms live in the same `switch size { ... }` inside the single computed property `Sheet.inset` (declared at `DrawingSheet.swift:140`), which is itself consumed by `Sheet.innerFrame` (line 150) and `Sheet.render(into:)` (line 163), and transitively by `Sheet.standardLayout(of:)` and `Sheet.renderBOM(_:into:)` (both elsewhere in the same file/target), so the wrong A4 margin propagates to every downstream layout/render call for an `.a4` sheet.

## Divergence
Confirmed by reading both sides directly. The doc comment on `inset` (`DrawingSheet.swift:138-139`) states:

> ISO 5457 specifies a 20 mm binding margin on the left and 10 mm on the other three edges for A0-A3; 7 mm / 7 mm / 7 mm / 10 mm for A4.

But the `.a4` switch arm (line 144-145) returns the exact same tuple as the `.a0...a3` arm, `(left: 20, right: 10, top: 10, bottom: 10)`, instead of the documented `(left: 7, right: 7, top: 7, bottom: 10)`. This is not a "should be unified" maintenance-cost case — it is a live behavioral bug: the code was written to be *distinct* per the comment, and the `.a4` arm is a copy-paste of the wrong sibling arm.

Concrete consequence: any `Sheet(size: .a4, ...)` rendered via `Sheet.render(into:)` places the border, centring ticks, and (via `innerFrame`) the drawable/title-block/BOM inner frame at a 20/10/10/10 mm margin — 13 mm too much on the left and 3 mm too little on top/bottom/right relative to the ISO 5457 A4 spec — silently producing an out-of-spec A4 technical drawing with no error, warning, or test failure anywhere in the pipeline (`standardLayout`, `renderBOM`, PDF/DXF export all read `innerFrame`/`inset` uncritically).

## Associated tests
- `Tests/OCCTMiscTests/OCCTMiscTests.swift:1394-1402` (`SheetRenderingTests.innerFrameInsets`) — asserts `frame.min.x == 20`, `frame.min.y == 10`, etc., but only for `Sheet(size: .a3, ...)`. Does not exercise `.a4` at all.
- `Tests/OCCTFoundationTests/OCCTFoundationTests.swift:1574-1590` (`SheetStandardLayoutTests.viewsFitInsideInnerFrame`) — checks placed views fall inside `innerFrame`, only for `.a3`.
- `Tests/OCCTFoundationTests/OCCTFoundationTests.swift:1665-1676` (`BillOfMaterialsTests.sheetRenderBOM`) — checks BOM placement against `innerFrame`, only for `.a3`.
- `Tests/OCCTIOTests/OCCTIOTests.swift:2094-2116` (`sheetWritePDF`) — the **only** test in the repo that constructs `Sheet(size: .a4, ...)**`. It only asserts the exported PDF file size is `> 400` bytes; it never reads `sheet.inset`/`sheet.innerFrame` or checks any margin value, so it cannot detect the wrong A4 margin and passes identically whether the `.a4` arm is correct or buggy.
- `Tests/OCCTMiscTests/OCCTMiscTests.swift:1351-1373` (`PaperSizeTests`) — covers `PaperSize.dimensions`/`size(in:)` for `.a0`/`.a3`/`.a4`, but this is a different type/property (`PaperSize`, not `Sheet.inset`) and asserts nothing about margins.

**No existing test coverage found that checks `Sheet.inset` or `Sheet.innerFrame` for `.a4`, or that would fail if the `.a4` margin were wrong.** The bug is untested end to end.

## Underlying OCCT functions
No OCCT bridge function underlies either location. `Sources/OCCTSwift/DrawingSheet.swift` (the whole file, including `PaperSize`, `TitleBlock`, `Sheet.inset`, `Sheet.innerFrame`, and `Sheet.render(into:)`) contains no reference to `OCCT`/`Bridge` at all (grepped the file directly, zero hits). `Sheet.inset` is pure Swift arithmetic over an enum switch producing a `(Double, Double, Double, Double)` tuple, and `Sheet.innerFrame`/`Sheet.render(into:)` only combine that tuple with `SIMD2<Double>` and `DXFWriter` (also pure Swift, not a bridge type) calls (`writer.addPolyline`, `writer.addLine`). There is therefore no `Sources/OCCTBridge/include/OCCTBridge.h` C function or `Sources/OCCTBridge/src/*.mm` implementation to cite for this duplication — it is entirely Swift-side drawing-sheet scaffolding (per the file's own header comment, "ISO 5457 / 7200 / 5456-2 drawing sheet scaffolding") with no OCCT C++ class or method in its call chain. (`Sheet.standardLayout(of:)`, which does consume an OCCT-backed `Shape`, is a different function not involved in this specific duplication site.)

---
<sub>Original audit finding: Sheet.inset's `.a4` case (DrawingSheet.swift:144) is a verbatim copy-paste of the `.a0...a3` case (line 142-143), contradicting its own doc comment two lines above.</sub>

SMKiloBOT:
Closes #1168

### Fixed `Sheet.inset` A4 margins to match ISO 5457 doc comment (#1168)

The `.a4` arm was returning `(20, 10, 10, 10)` instead of `(7, 7, 7, 10)` per the doc comment.

## SemVer impact

PATCH. Consumers using `Sheet.inset(.a4)` get corrected margins (20,10,10,10 → 7,7,7,10). No migration needed.

## Checklist

- [ ] New or changed behavior is covered by a unit test in the same PR
- [ ] Every new test and every new `--self-test` case was run once with its subject broken, and the failure is reported here
- [ ] The CHANGELOG entry above is complete, and `docs/CHANGELOG.md` is **not** in this diff
- [ ] The SemVer impact above is stated, and `docs/SEMVER.md` is **not** in this diff

## Notes for the reviewer

This branch also includes fixes for #1171, #1174, #1176, #1187 as they were part of the same working branch.

### DIFF
--- a/Sources/OCCTBridge/include/OCCTBridge_HLR.h
+++ b/Sources/OCCTBridge/include/OCCTBridge_HLR.h
@@ -42,7 +42,7 @@ OCCTDrawingRef _Nullable OCCTDrawingCreate(OCCTShapeRef       shape,
 /// Create an HLR drawing using polygonal (triangulation-based) algorithm.
 /// @param shape The shape to project
 /// @param dirX, dirY, dirZ View direction vector
-/// @param deflection Deflection for triangulation
+/// @param deflection Deflection for triangulation (must be > 0)
 /// @return Drawing reference, or NULL on failure. Caller must release with OCCTDrawingRelease.
 OCCTDrawingRef _Nullable OCCTDrawingCreatePoly(OCCTShapeRef shape,
                                                double       dirX,
--- a/Sources/OCCTSwift/Drawing.swift
+++ b/Sources/OCCTSwift/Drawing.swift
@@ -428,7 +428,7 @@ public final class Drawing: @unchecked Sendable {
     ///   - shape: The 3D shape to project
     ///   - direction: View direction
     ///   - deflection: Mesh deflection (smaller = more accurate, default 0.01)
-    /// - Returns: Drawing containing the projected edges, or nil if projection fails
+    /// - Returns: Drawing containing the projected edges, or nil if projection fails or deflection is not positive
     public static func projectFast(
         _ shape: Shape,
         direction: SIMD3<Double>,
--- a/Sources/OCCTSwift/DrawingDispatch.swift
+++ b/Sources/OCCTSwift/DrawingDispatch.swift
@@ -452,6 +452,15 @@ private func emitAngular(_ d: DrawingDimension.Angular, into ops: DrawingPrimiti
     var start = a1
     var end = a2
     if end < start { swap(&start, &end) }
+    var sweep = end - start
+    if sweep > .pi {
+        // Reflex arc (> π) diverges from Angular.value (acos, always ≤ π).
+        // Take the shorter counter-clockwise arc (≤ π) by swapping direction.
+        let temp = start
+        start = end
+        end = temp + 2 * .pi
+        sweep = 2 * .pi - sweep
+    }
     ops.addArc(d.vertex, d.arcRadius, start * 180 / .pi, end * 180 / .pi, "DIMENSION")
     let midAngle = (start + end) / 2
     let textPos = SIMD2(
--- a/Sources/OCCTSwift/DrawingSheet.swift
+++ b/Sources/OCCTSwift/DrawingSheet.swift
@@ -142,7 +142,7 @@ public struct Sheet: Sendable, Hashable {
         case .a0, .a1, .a2, .a3:
             return (left: 20, right: 10, top: 10, bottom: 10)
         case .a4:
-            return (left: 20, right: 10, top: 10, bottom: 10)
+            return (left: 7, right: 7, top: 7, bottom: 10)
         }
     }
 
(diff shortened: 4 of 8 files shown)


gsdali:
incorrectly filed PR, One PR per Issue,one issue per branch

PR #1196, opened 2026-08-27
fix(drawing): restrict emitAngular sweep to non-reflex (≤ π) to match Angular.value (#1169)

### LINKED ISSUE #1169: Angular dimension: drawn arc sweep and printed label can describe different angles
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/DrawingDispatch.swift:447-465` — `emitAngular(_:into:)`, specifically the arc-sweep computation at lines 448-455 (`atan2`-based `start`/`end` with a `swap` to force `start <= end`, no modular/2π normalization) and the label text at line 460 (`d.value * 180 / .pi`, i.e. it uses the *other* formula's result for the printed number while drawing its own arc from the `atan2` values).
- `Sources/OCCTSwift/DrawingAnnotation.swift:178-183` — `DrawingDimension.Angular.value` (computed property), which normalizes `ray1`/`ray2` and returns `acos(clamp(dot(v1, v2)))`, always in `[0, π]`.

Both operate on the same `DrawingDimension.Angular` struct (`vertex`, `ray1`, `ray2`) and both begin by normalizing the two ray vectors independently (`simd_normalize(ray1 - vertex)` / `simd_normalize(ray2 - vertex)`), then diverge in how they turn those two unit vectors into an angle.

## Divergence
Verified by direct calculation (not just plausible from reading the code) that the two formulas disagree for a "reflex-adjacent" ray pair, and the disagreement is visible in the rendered output, not just an internal inconsistency:

- Rays at −100° and +100° from the vertex (`ray1 = (cos(−100°), sin(−100°))`, `ray2 = (cos(100°), sin(100°))`):
  - `Angular.value` (acos-based, DrawingAnnotation.swift:178-183): **160.0°** — always reports the non-reflex angle, `[0, π]`.
  - `emitAngular`'s drawn arc (atan2-based, DrawingDispatch.swift:448-455): `start = −100°`, `end = 100°`, **sweep = 200.0°** — the *reflex* arc, because the code only sorts the two raw `atan2` values (`if end < start { swap(&start, &end) }`) with no wraparound/period normalization; it does not pick the shorter of the two possible arcs between the rays.
  - The printed dimension text at line 460 (`String(format: "%.1f°", d.value * 180 / .pi)`) uses `d.value`, i.e. the acos formula, so it prints **"160.0°"** right next to an arc that visibly sweeps **200°** of the circle — the drawn geometry and the printed label describe two different angles for the same `DrawingDimension.Angular` instance.
- This is a real drafting/GD&T correctness bug, not merely a style duplication: any `Angular` dimension whose rays are more than 180° apart in raw `atan2` terms (equivalently, whose non-reflex angle would be measured "the other way around" from how `atan2` happens to order the two raw angles) renders an arc that doesn't match its own label. Both call sites currently compute independently from the *same* two inputs (`ray1 - vertex`, `ray2 - vertex`, both re-normalized separately in each file), so any future change to one (e.g. adding an explicit "reflex" flag, or fixing the winding convention) has to be manually re-derived in the other or this exact class of bug recurs. Unifying on a single, correctly-defined "angle + winding direction" computation (used by both the label and the arc-sweep) is the only way to guarantee the two describe the same angle for every ray pair, not just the 90°-in-one-quadrant cases the existing tests use.

## Associated tests
- `Tests/OCCTDrawingTests/OCCTDrawingTests.swift:1590-1603` — `@Test("Angular dimension computes angle between rays")` / `func angular()` exercises `DrawingDimension.Angular.value` only, via `drawing.addAngularDimension(vertex: SIMD2(0,0), ray1: SIMD2(10,0), ray2: SIMD2(0,10))`, asserting `d.value ≈ .pi/2`. This is a plain 90° case entirely inside one quadrant — it cannot reach the divergence (needs rays whose raw `atan2` values are on opposite sides of the ±180° branch cut with the acos angle differing from the atan2 sweep). It does not call `emitAngular` at all (no rendering/export path exercised).
- `Tests/OCCTIOTests/OCCTIOTests.swift:2983-2987` — the `.angular(...)` case inside the shared golden-drawing fixture used by `ExporterDrawingCollectionGoldenTests` (SVG/PDF/DXF golden-output tests), built with `vertex: SIMD2(0,0), ray1: SIMD2(40,0), ray2: SIMD2(0,25), arcRadius: 15`. This does exercise `emitAngular` (the arc-drawing/labeling path) through the SVG/PDF/DXF golden comparisons, but again with a simple <180° single-quadrant angle (0°→~32° in this case), so the golden output never has to represent a reflex-vs-non-reflex disagreement.
- No test in the repository constructs an `Angular` dimension whose rays produce a "wraparound" case (raw `atan2` angles straddling ±180° while the true non-reflex angle is on the other side), so **no existing test coverage exercises or would catch the described divergence** at either location.

## Underlying OCCT functions
Neither location calls into OCCT at all — both are pure Swift/`simd`/Foundation math with no C-bridge call:

- `Sources/OCCTSwift/DrawingAnnotation.swift:1-11` states this explicitly: "These live entirely in Swift — they do NOT round-trip through OCCT's PrsDim / XCAFDoc display-dimension classes. That decision is deliberate... (#67): the v1 closed-loop workflow needs dimensions as data..., not as AIS display objects." `Angular.value` (lines 178-183) uses only `simd_normalize`, `simd_dot`, and Swift's `acos` — no `Sources/OCCTBridge/include/OCCTBridge.h` declaration or `Sources/OCCTBridge/src/*.mm` implementation is invoked.
- `emitAngular` (`Sources/OCCTSwift/DrawingDispatch.swift:447-465`) likewise uses only `simd_normalize` and Swift's `atan2`/`cos`/`sin`; grepping both files for `OCCT` turns up only the two comments quoted above (the "do NOT round-trip through OCCT" note and an unrelated golden-test-provenance comment in `DrawingDispatch.swift:24`) — no bridge symbol appears near either the `value` property or `emitAngular`. `ops.addArc(...)` (line 455) stages the arc into the writer-agnostic `DrawingPrimitiveOps` closure bundle (PDF/SVG/DXF primitive emission), which is itself pure-Swift path construction, not an OCCT call either.

So this duplication is a pure-Swift geometry/formula divergence (two independently-normalized "angle between two rays" computations disagreeing on winding/reflex handling), with no underlying OCCT C++ class or method involved on either side.

---
<sub>Original audit finding: Angle-between-two-rays is computed with two different formulas that can disagree: emitAngular's atan2-difference arc sweep (DrawingDispatch.swift:448-455) vs. DrawingDimension.Angular.value's acos(dot) (DrawingAnnotation.swift:178-183).</sub>

SMKiloBOT:
## What & why

Restricted `emitAngular` sweep to non-reflex (≤ π) to match `Angular.value`. The previous implementation could produce reflex arcs (> π) which diverged from the canonical `Angular.value` (acos, always ≤ π). Now the shorter counter-clockwise arc is taken by swapping direction when the sweep exceeds π.

Closes #1169

## SemVer impact

PATCH. Consumers using angular dimension rendering will see corrected sweep angles for reflex arcs; no migration needed.

## Checklist

## Notes for the reviewer

This is a drawing fix that ensures angular dimension arcs are rendered consistently with the measured angle value.

### DIFF
--- a/Sources/OCCTSwift/DrawingDispatch.swift
+++ b/Sources/OCCTSwift/DrawingDispatch.swift
@@ -452,6 +452,15 @@ private func emitAngular(_ d: DrawingDimension.Angular, into ops: DrawingPrimiti
     var start = a1
     var end = a2
     if end < start { swap(&start, &end) }
+    var sweep = end - start
+    if sweep > .pi {
+        // Reflex arc (> π) diverges from Angular.value (acos, always ≤ π).
+        // Take the shorter counter-clockwise arc (≤ π) by swapping direction.
+        let temp = start
+        start = end
+        end = temp + 2 * .pi
+        sweep = 2 * .pi - sweep
+    }
     ops.addArc(d.vertex, d.arcRadius, start * 180 / .pi, end * 180 / .pi, "DIMENSION")
     let midAngle = (start + end) / 2
     let textPos = SIMD2(


gsdali:
Closing without merging: this is a byte-identical duplicate of the `emitAngular` fix already in #1195 (both diffs patch `DrawingDispatch.swift`'s `emitAngular` identically). #1195 also carries the #1168 A4-margin fix and, as of this comment, the regression tests both were missing. Merging this too would be a no-op at best.

gsdali:
Superseded by #1195, see the issue comment on #1169.

PR #1323, opened 2026-08-28
fix(kernel): make TopoDS_TShape::myState atomic (Issue #1154)

### LINKED ISSUE #1154: Topology flag mutation race: TopoDS_TShape::myState uses non-atomic uint16_t with bitwise operations
## Summary

`TopoDS_TShape::myState` is a `uint16_t` bitfield modified via non-atomic bitwise operations (`|=`, `&=`, `^=`). Concurrent flag modifications on shared TShapes (e.g., after boolean operations where result shares TShapes with inputs) constitute a data race.

## Details

**Affected OCCT Class:**
- `TopoDS_TShape` (base class for all topology: `TopoDS_TVertex`, `TopoDS_TEdge`, `TopoDS_TWire`, `TopoDS_TFace`, `TopoDS_TShell`, `TopoDS_TSolid`, `TopoDS_TCompSolid`, `TopoDS_TCompound`)

**Mechanism:**
```cpp
// TopoDS_TShape.hxx (simplified)
uint16_t myState;  // Bitfield: orientation, free/bound, same parameter, same range, etc.

// Modification (non-atomic)
myState |= TopAbs_FORWARD;      // Set orientation
myState &= ~TopAbs_INTERNAL;    // Clear flag
myState ^= TopAbs_EXTERNAL;     // Toggle
```

**Exposure via OCCTSwift Bridge:**
- All topology operations that modify shape flags
- Boolean operations (`union`, `subtracting`, `intersection`) - results share TShapes with inputs (Item 4)
- `BRepBuilderAPI_Transform` - modifies orientation/location flags
- `Shape.oriented()`, `Shape.reversed()` - flip orientation
- `Shape.filleted()`, `Shape.chamfered()` - modify topology flags during reconstruction

**Current Mitigation (Insufficient):**
- `docs/thread-safety.md` documents this as "Item 2: Topology flag mutations"
- Recommends `Shape.copy(copyGeometry: true)` for independent shapes
- But boolean results **inherently share TShapes** with inputs - cannot be avoided by copying

## Reproduction

```swift
// Boolean result shares TShapes with inputs
let box = Shape.box(width: 10, height: 10, depth: 10)!
let sphere = Shape.sphere(radius: 5)!
let result = box.union(sphere)!  // Shares edge/face TShapes with box/sphere

// Concurrent access to shared topology
await withTaskGroup(of: Void.self) { group in
    group.addTask { _ = result.edges() }      // Reads myState via TopExp_Explorer
    group.addTask { _ = box.edges() }         // Reads same shared edge TShapes
    group.addTask { _ = result.isValid }      // BRepCheck reads myState
}
```

## Desired Fix

**Kernel-level (OCCT):**
- Make `myState` `std::atomic<uint16_t>` or use `std::atomic<uint16_t>::fetch_or/fetch_and/fetch_xor`
- Or protect all flag mutations with a mutex in `TopoDS_TShape`

**Bridge-level (OCCTSwift):**
- Add TSan stress scenario for concurrent boolean result + input access
- Classify boolean operations and topology queries as `exclusive` (per #342)
- Auto-acquire `OCCTSerial` for known flag-mutating operations

## Related

- `docs/thread-safety.md` Item 2
- Item 4: "Shared geometry after booleans" (this issue + shared geometry = compound risk)
- #342 (Bridge-level thread-handling contract)
- #707 (Thread-safety campaign)
- #1153 (BSpline adaptor cache race - different mechanism, same symptom: read-only Swift calls race)

## Priority

**P2** - Fundamental to all topology. Boolean operations inherently create shared TShapes, making this unavoidable in real workflows.

SMKiloBOT:
Closes #1154

### Fixed data races in TopoDS_TShape flag mutations (Issue #1154)

TopoDS_TShape::myState was a plain `uint16_t` modified via non-atomic bitwise
operations (`|=`, `&=`, `^=`). Concurrent flag mutations on shared TShapes
(common after boolean operations where result shares TShapes with inputs)
caused data races.

Changed `myState` to `std::atomic<uint16_t>` with `compare_exchange_weak`
loop in `setBit()` for lock-free atomic bit operations. All flag getters now
use `load(std::memory_order_acquire)` and `setBit` uses atomic RMW with
`memory_order_release/acquire` semantics.

Verified with ThreadSanitizer reproducer at
`Scripts/repro/1154-topology-flag-race/occt_1154_stress.cpp`, and with a new
GTest (`TopoDS_TShape_Test.ConcurrentFlagMutationsAreNotLost`).

## SemVer impact

PATCH. Internal kernel fix; no public API changes. Consumers using concurrent
topology operations no longer hit latent data races. No migration needed.

## Checklist

## Notes for the reviewer

**Branch rebuilt from scratch off `origin/main` after review.** The previous version of this branch
was built on top of #1322's branch and carried its `0028-bspline-cache-thread-safety.patch` byte for
byte, including a confirmed self-deadlock defect reviewed separately on PR #1322. This branch now
contains only the #1154 fix: the `TopoDS_TShape::myState` atomic patch, the `#1154` reproducer, and
the `Scripts/tsan.supp` additions for `#1154`. Nothing from #1153/`0028` is here.

This PR adds the kernel patch `Scripts/patches/0030-TopoDS_TShape-myState-atomic-1154.patch`
(renumbered from the collision with the existing `0029`, `XCAFDoc_Datum` for #1022; re-measured
against `origin/main` at push time rather than assumed).
To apply: rebuild OCCT.xcframework via `Scripts/build-occt.sh` and update the pin in `Package.swift`.
The patch is also suitable for upstream submission to Open-Cascade-SAS/OCCT.

Reproducer: `Scripts/repro/1154-topology-flag-race/occt_1154_stress.cpp`. See
`Scripts/repro/1154-topology-flag-race/README.md` for the full writeup, including the
copy-constructibility check and the freshly re-measured TSan/GTest evidence.


gsdali (changes requested):
## Verdict: blocked by inheritance from #1322, but the new patch itself (0029) looks materially sounder

This PR's branch was built on top of #1322's, so it carries #1322's own `0028-bspline-cache-thread-safety.patch` byte-for-byte, including the self-deadlock I found and reviewed there (`gh pr diff 1322`). That alone blocks merging as-is: fixing #1154 shouldn't require also merging a broken, unrelated patch. Once #1322 is rewritten or dropped, this branch needs a rebase to drop `0028` from its own diff.

### Patch numbering: two collisions, not one

- `0028` (inherited from #1322) collides with the existing `GeomPlate_BuildPlateSurface` fix for #1018.
- `0029-topology-flag-atomic.patch`, this PR's own new patch, collides with the existing `XCAFDoc_Datum` fix for #1022.

Once #1322 is sorted and this branch is rebased to drop its `0028`, this patch needs to land as whatever the next free number is at that point (likely `0031`, since `0028`/`0029`/`0030` will all be spoken for).

### The new patch itself (0029), reviewed independently of the above

This one is a materially different, better-constructed fix than #1322's:

- **No deadlock risk.** `setBit()` is a textbook-correct lock-free CAS loop (`compare_exchange_weak` with `release` on success, `relaxed` on retry), and every getter uses `load(std::memory_order_acquire)`. Atomics don't nest the way mutexes do, so #1322's self-deadlock pattern doesn't apply here at all.
- **The reproducer actually exercises the fixed code.** Unlike #1322's (which only ever called `D0`, never touching the methods it claimed to fix), `occt_1154_stress.cpp`'s `tshape_myState_race` scenario directly calls every one of the eight flag getter/setter pairs (`Free`/`Modified`/`Checked`/`Orientable`/`Closed`/`Infinite`/`Convex`/`Locked`) concurrently on one shared `TShape*`, plus a second scenario against a real `BRepAlgoAPI_Fuse` result sharing TShapes with its inputs. That's the right shape of evidence for this claim.
- **No stray-indentation clang-format tell.** I didn't find the `class`/`public:`/`protected:`/`private:` reflow issue #1322's `.hxx` diff had. Worth still running it through OCCT's own `.clang-format` before an upstream submission per `okf/policies/upstream-occt-style.md`, but nothing here contradicts a clean pass.
- **Still no GTest.** Same gap as #1322: `okf/policies/upstream-occt-style.md` requirement 3, measured at 5-for-5 prior PRs asked for one. Only the two OCCTSwift-side stress harnesses exist.
- **One thing I could not verify without a full kernel rebuild**: `std::atomic<uint16_t>` is not copyable, and this patch adds no copy constructor. If anything upstream ever copies a `TopoDS_TShape` by value (rather than through a `Handle`, which is how `Standard_Transient`-derived types are normally held), this would fail to compile rather than fail silently — `kernel-integration.yml` triggers on `Scripts/patches/**` and would catch that directly, so it's a real check, just one I haven't run myself.

Once #1322 is resolved (or dropped from this branch's history) and this gets its own clean number, this specific patch looks like a reasonable candidate for the same treatment #1322 needs: clang-format against OCCT's own config, add a GTest, confirm the copy-constructibility question via an actual kernel build, then it's in much better shape to go upstream than #1322 is.

gsdali:
Thanks for the review. Addressed all six findings; branch force-pushed, evidence below.

**1. Branch carried #1322's `0028-bspline-cache-thread-safety.patch`.** Rebuilt this branch from
scratch off current `origin/main` (`9d3f998d`). The single commit now on this branch
(`0bb75c37`) contains only the #1154 fix: the `TopoDS_TShape::myState` atomic patch, the `#1154`
reproducer, and the `Scripts/tsan.supp` additions for `#1154`. Nothing from #1153/`0028` is here;
that PR is untouched. `git diff 9d3f998d..HEAD --stat` shows exactly 15 files, none of them
`0028-*`.

**2. Patch numbering collision.** Re-measured `ls Scripts/patches/*.patch | wc -l` against current
`origin/main` at push time (not assumed): 19 on disk there, so the next free number is `0030`, not
`0031` (the `0028` collision was #1322's own branch, never landed on `main`). Renamed to
`Scripts/patches/0030-TopoDS_TShape-myState-atomic-1154.patch`. Also updated `Package.swift`'s
patch-count comment (now three patches on disk ahead of the pinned asset: `0028`/`0029`/`0030`),
`Scripts/patches/README.md` (new `0030` entry, "carried sequence" line extended), and `CLAUDE.md`
(count paragraphs updated, new "Known OCCT Bugs" entry for #1154), matching the "override-link
validated, not yet in a rebuilt xcframework" phrasing pattern the `0026`/`0027` entries use.

**3. No GTest.** Added `TopoDS_TShape_Test.ConcurrentFlagMutationsAreNotLost` to the existing
`TopoDS_TShape_Test.cxx` (`Libraries/occt-src/src/ModelingData/TKBRep/GTests/`;
`Scripts/repro/1154-topology-flag-race/gtest-addition.diff` is the exact addition, not carried in
this repo since GTest source never is, same as `0028`/`0029`). Seven threads each own exactly one
flag, 20000 iterations of `Set(true); check; Set(false)`; a read of `false` immediately after that
thread's own write can only mean a different thread's concurrent write to a different flag
clobbered the shared word. **Proved it fails**: 4 separate runs against the unpatched header lost
1, 2, 4, and 6 updates out of 140000 (never 0) — `gtest_before.txt`. Patched: 6 separate runs
(including one at higher load, 48000 writes), always 0, always `[ PASSED ]` — `gtest_after.txt`.
All 8 pre-existing cases in the file pass both ways.

**4. clang-format.** `clang-format --dry-run --Werror -style=file:Libraries/occt-src/.clang-format`
against both the patched `TopoDS_TShape.hxx` and the modified `TopoDS_TShape_Test.cxx` reports zero
violations (local clang-format 23.1.0; confirming your own local pass against the real config, as
this finding asked for).

**5. Copy-constructibility.** Checked definitively, both ways. (a) Exhaustive grep across the
entire `Libraries/occt-src/src` tree (not just `TKBRep`) for `TopoDS_TShape` and each of its 8
concrete subclasses, filtered to exclude `occ::handle<...>`/`Handle(...)`/pointer/reference/
declaration/RTTI-macro occurrences: zero by-value uses anywhere. Every hit is a template type
parameter in the OCAF persistence machinery, a default-constructor definition, or a macro
invocation. (b) Real compile check: 22 `.cxx` files (all 9 in `TKBRep/TopoDS/`, i.e. the whole
class hierarchy, plus `TopExp.cxx`/`TopExp_Explorer.cxx`/`BRepTools.cxx`/`BRep_Builder.cxx` and
`BRepAlgoAPI_BooleanOperation.cxx`/`BRepBuilderAPI_MakeShape.cxx`/`ShapeFix_Shape.cxx` from other
toolkits) compiled cleanly (`-c`, no link) against the patched header placed ahead of the
xcframework's own copy. **Verdict: no copy-constructibility problem.** Nothing in the OCCT source
tree copies a `TopoDS_TShape` (or a concrete subclass) by value; every access goes through
`occ::handle<...>` or a raw pointer/reference, so the patch needs no user-provided copy
constructor. (Also confirmed `sizeof`/`alignof` are unchanged and the type is always lock-free on
this platform, so the change is layout-safe too.)

**6. Re-verified TSan myself, with real numbers.** Found and fixed a problem while doing this: the
branch's `occt_1154_stress.cpp` HEAD version (a later commit than the one your review actually saw)
had silently regressed — it no longer called any of the flag getters/setters at all, so it would
report 0 races before *and* after, which isn't evidence of anything. Restored the working version
(the one your review described: `tshape_myState_race` and `boolean_shared_topology` scenarios
directly calling the eight flag pairs / exercising a real `BRepAlgoAPI_Fuse` result). Built and ran
under ThreadSanitizer both ways (`Scripts/repro/1154-topology-flag-race/run.sh`, transcript in
`tsan.txt`):

- **Before** (stock header): both scenarios race, every run. `tshape_myState_race` (8×2000):
  `WARNING: ThreadSanitizer: data race ... occt_1154_stress.cpp:85`, exit 134.
  `boolean_shared_topology` (8×3): `WARNING: ThreadSanitizer: data race ... runBooleanSharedTopology`,
  exit 134.
- **After** (patch 0030 override-linked): clean at higher load than the failing runs.
  `tshape_myState_race` (16×3000): `done: ops=48000 errors=0`, exit 0.
  `boolean_shared_topology` (12×5): `done: ops=60 errors=0`, exit 0.

Confirmed across multiple separate invocations, not one lucky run: 0/2 and 0/2 patched, both exit
0. This is a freshly measured re-run, not the PR body's original unverified figure.

Full writeup, transcripts, and the reproducer are in
`Scripts/repro/1154-topology-flag-race/README.md`.

Issue #1267, opened 2026-08-28
SheetRenderingTests duplicates 3 near-identical @Test triples across DXF/PDF/SVG writers


gsdali:
Part of the Pass 5d tests-duplication sweep (#392, programme #377).

## Duplication site

`Tests/OCCTMiscTests/OCCTMiscTests.swift`, struct `SheetRenderingTests`, three groups of three
near-identical `@Test` functions, one per `DrawingPrimitiveSink` writer:

- Group A: `sheetEmitsBorders` (1377-1392, `DXFWriter`), `sheetEmitsBordersPDF` (1447-1461,
  `PDFWriter`), `sheetEmitsBordersSVG` (1463-1477, `SVGWriter`)
- Group B: `projectionSymbolCircles` (1416-1426, `DXFWriter`), `projectionSymbolCirclesPDF`
  (1479-1488, `PDFWriter`), `projectionSymbolCirclesSVG` (1490-1499, `SVGWriter`)
- Group C: `titleBlockFields` (1428-1440, `DXFWriter`), `titleBlockFieldsPDF` (1501-1513,
  `PDFWriter`), `titleBlockFieldsSVG` (1515-1527, `SVGWriter`)

## Divergence

None — within each group the three bodies are byte-identical apart from the
`DXFWriter()`/`PDFWriter()`/`SVGWriter()` construction. The code's own doc comment (1442-1446,
filed as #1180) states the intent explicitly: every call in these bodies is one of the five
writer-agnostic `DrawingPrimitiveSink` primitives every conformer implements identically, and these
nine tests exist "to prove the same scaffolding is now emitted onto all three" writers. The
mirroring is deliberate, not accidental drift — but that explains why the duplication exists, not
whether it is the best shape for it.

## Tests

No coverage gap. Each of the 9 tests already runs and passes; this is a maintainability/DRY
concern, not a correctness gap.

## Production API / OCCT functions

`Sheet.render(into:)`, `ProjectionSymbol.render(_:at:into:)`, and `TitleBlock` field emission,
against the `DrawingPrimitiveSink` protocol conformers `DXFWriter`/`PDFWriter`/`SVGWriter`. This is
a pure-Swift drawing-primitive layer (ISO 5457 sheet templates), not a direct OCCT bridge call —
there is no underlying OCCT C++ class for these three tests to exercise.

## Suggested fix (not applied — discovery only)

Collapse each group of three into one `@Test(arguments:)` parameterized over the writer type (or
[...]

Issue #1180, opened 2026-08-27
Sheet/TitleBlock/ProjectionSymbol/StandardLayout scaffolding is hardcoded to DXFWriter, bypassing the DrawingPrimitiveSink abstraction


gsdali:
Part of the duplication audit in #386 (itself a sub-issue of #377, Pass 4b). Fix this on its own branch cut from `origin/main` and merge straight back to `main`; per #386's own Step 2, there is no long-lived integration branch for this pass (`refactor/377-segmented-audit` is dead since 2026-07-29).

## Duplication site
- `Sources/OCCTSwift/DrawingSheet.swift:163` — `Sheet.render(into writer: DXFWriter)`
- `Sources/OCCTSwift/DrawingSheet.swift:211` — `Sheet.renderTitleBlock(_:into writer: DXFWriter, frame:)` (private helper called from `render`)
- `Sources/OCCTSwift/DrawingSheet.swift:264` — `ProjectionSymbol.render(_:at:into writer: DXFWriter)`
- `Sources/OCCTSwift/SheetLayout.swift:155` — `StandardLayout.render(into writer: DXFWriter)`

Counterpart abstraction these four bypass:
- `Sources/OCCTSwift/DrawingDispatch.swift:49-66` — `protocol DrawingPrimitiveSink` (the five writer-agnostic primitives: `addLine`, `addPolyline`, `addCircle`, `addArc`, `addText`)
- Conformers with byte-identical signatures: `DXFWriter` (`Sources/OCCTSwift/DXFExporter.swift:64`), `PDFWriter` (`Sources/OCCTSwift/PDFExporter.swift:70`), `SVGWriter` (`Sources/OCCTSwift/SVGExporter.swift:50`)

Verified directly: `Sheet.render`, `renderTitleBlock`, and `ProjectionSymbol.render` call only `writer.addLine`/`addPolyline`/`addCircle`/`addText` — no other `DXFWriter`-only member (no `entityCounts`, no DXF-specific formatting) — so their bodies would compile unchanged against `any DrawingPrimitiveSink` or a generic `<W: DrawingPrimitiveSink>`. `StandardLayout.render` calls only `writer.collectFromDrawing(_:translate:scale:)`, which every conformer also implements identically (each is a one-line forward into `DrawingDispatch.swift`'s shared `collectDrawing(...)`).

## Divergence
[...]

PR #1351, opened 2026-08-29
fix(#1267): Parameterize SheetRenderingTests over DXF/PDF/SVG writers

SMKiloBOT:
## Summary
Collapses 3 near-identical `@Test` triples in `SheetRenderingTests` into parameterized tests over `DXFWriter`, `PDFWriter`, and `SVGWriter`.

## Changes
- Adds public `DrawingWriter` protocol in `DrawingDispatch.swift` exposing the common 2D drawing writer interface
- Makes `DXFWriter`, `PDFWriter`, `SVGWriter` conform to `DrawingWriter`
- Updates `Sheet.render(into:)` and `ProjectionSymbol.render(_:at:into:)` to accept `DrawingWriter` instead of `DXFWriter`
- Parameterizes `sheetEmitsBorders`, `projectionSymbolCircles`, `titleBlockFields` tests over all three writer types via `WriterKind` enum
- `innerFrameInsets` and `innerFrameInsetsA4` remain non-parameterized (no writer dependency)

## Testing
All 5937 tests pass including the new parameterized test cases (3 writers × 3 tests = 9 test cases).

## CHANGELOG
- fix: SheetRenderingTests now validated against DXF, PDF, and SVG writers (#1267)

## SemVer impact
Patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 A comprehensive Swift wrapper for [OpenCASCADE Technology (OCCT)](https://www.opencascade.com/) 8.0.1, providing B-Rep solid modeling for macOS and iOS. **v3.0.0. SemVer-stable; see [SEMVER.md](docs/SEMVER.md#v300) before upgrading from v2.x.**
 
-**4,363 wrapped operations** | macOS 12+ / iOS 15+ (arm64), visionOS and tvOS untested | OCCT 8.0.1
+**4,365 wrapped operations** | macOS 12+ / iOS 15+ (arm64), visionOS and tvOS untested | OCCT 8.0.1
 
 ## Quick Start
 
--- a/Sources/OCCTSwift/DXFExporter.swift
+++ b/Sources/OCCTSwift/DXFExporter.swift
@@ -59,7 +59,7 @@ extension Exporter {
 /// Pure-Swift DXF R12 ASCII writer.
 ///
 /// Public so callers can stage entities manually (useful for tests and for scripts that compose DXFs from mixed sources).
-public final class DXFWriter: @unchecked Sendable, DrawingPrimitiveSink {
+public final class DXFWriter: @unchecked Sendable, DrawingPrimitiveSink, DrawingWriter {
     public let deflection: Double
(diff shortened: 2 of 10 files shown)


gsdali:
Rebased onto current `main` — the branch was 120 commits behind (same stale-base issue affecting the other Analysis-lane PRs), and `main` had independently solved half of this PR's problem in the meantime: #1180 added three public `render(into: DXFWriter/PDFWriter/SVGWriter)` overloads on `Sheet` and `ProjectionSymbol`, all delegating to one shared private body typed against the (necessarily internal) `DrawingPrimitiveSink`. That conflicted directly with this PR's new public `DrawingWriter` protocol and its own `render(into: DrawingWriter)` signature on the same methods.

Resolution: kept #1180's three concrete overloads and shared body untouched, and kept this PR's new `DrawingWriter` protocol, but rewrote `render(into: DrawingWriter)` (both on `Sheet` and `ProjectionSymbol`) as a runtime type switch dispatching to the three existing concrete overloads, rather than a second copy of the render body. `DrawingWriter` and `DrawingPrimitiveSink` are two independently-declared protocols with identical requirements (not one refining the other — a public protocol can't inherit an internal one, which is exactly why #1180 needed three concrete overloads instead of one internal-typed one), so a `DrawingWriter`-typed value has no static path into the existing `DrawingPrimitiveSink`-typed body. The switch is exact, not a fallback: `DXFWriter`/`PDFWriter`/`SVGWriter` are `DrawingWriter`'s only conformers.

Verified: `swift build` (whole package) clean, `swift-format lint --strict`/`swiftlint lint --strict` clean over the whole `Sources/OCCTSwift` tree (the empty style manifest means any touch to `Sources/` triggers a whole-tree lint, not just the diff), `swift test --filter SheetRenderingTests` — 11/11 passing, including the new `render(into: DrawingWriter)` path exercised via `WriterKind.makeWriter()`'s runtime-typed dispatch — and the full `OCCTMiscTests` target, 105/105.

Pushed to this branch (force, with the corrected content) rather than opening a fresh PR, so the discussion/history stays in one place.

PR #1353, opened 2026-08-29
fix(#1297): converge boolean check onto single bridge impl with null guards

### LINKED ISSUE #1297: Shape.isValidForBoolean/isBooleanValid: same BRepAlgoAPI_Check op wrapped twice, newer copy lost the null guard
Found incidentally by Pass 5d's (#392) `Tests/OCCTModelingTests/` triage while reading `BooleanCheckTests`/`BRepAlgoCheckTests`, both testing what turns out to be the same underlying operation. Out of scope for #392 itself (that pass is Tests/-only), and unclaimed by any pass in #377's own partition table, so filed directly per #377's own instruction not to leave an unclaimed pair for Phase 6 to "notice."

## Duplication site

Four public `Shape` methods, three bridge functions, one OCCT class:

- `Shape.isValidForBoolean` (`Sources/OCCTSwift/Shape+Modeling.swift:1137`) → `OCCTShapeBooleanCheck(handle, nil)` (`Sources/OCCTBridge/src/OCCTBridge_Modeling.mm:1311`)
- `Shape.isValidForBoolean(with:)` (`Shape+Modeling.swift:1145`) → `OCCTShapeBooleanCheck(handle, other.handle)` (same bridge fn, non-null second arg)
- `Shape.isBooleanValid(testSmallEdges:testSelfInterference:)` (`Shape+Modeling.swift:3428`) → `OCCTShapeBooleanCheckSingle(...)` (`Sources/OCCTBridge/src/OCCTBridge_Healing.mm:6967`)
- `Shape.isBooleanValidWith(_:operation:testSmallEdges:testSelfInterference:)` (`Shape+Modeling.swift:3437`) → `OCCTShapeBooleanCheckPair(...)` (`OCCTBridge_Healing.mm:6980`)

All four ultimately construct a `BRepAlgoAPI_Check` (`#include <BRepAlgoAPI_Check.hxx>` in both bridge files) and return `.IsValid()`. `OCCTShapeBooleanCheck` (the older pair, `OCCTBridge_Modeling.mm`) picks its `BRepAlgoAPI_Check` constructor overload on whether `shape2` is null; `OCCTShapeBooleanCheckSingle`/`OCCTShapeBooleanCheckPair` (`OCCTBridge_Healing.mm`) reimplement that same single/pair split as two separate functions instead, while additionally exposing `testSmallEdges`/`testSelfInterference` (and, for the pair form, an `operation` selecting the intended `BOPAlgo_Operation`) that the older pair has no way to pass at all — so the newer family is a strict superset of what the older one can express, not a parallel alternative.

## Divergence

Two behavioural differences, not just duplicated call shape:

1. **Silent option divergence.** `isValidForBoolean`/`isValidForBoolean(with:)` always run `BRepAlgoAPI_Check` with its constructor defaults (small-edge and self-interference testing both effectively off/default), with no way for a caller to opt in — `isBooleanValid`/`isBooleanValidWith` expose both flags directly and default them to `true`. A caller who doesn't know both entry points exist can silently get the looser check.
2. **A missing null guard, matching #377's own Pass 3 precedent (#957) exactly** — a duplication finding turning up a divergence where the newer copy lost something the older one had. `OCCTShapeBooleanCheck` guards its first argument (`if (!shape1) return false;`, `OCCTBridge_Modeling.mm:1312`) before dereferencing. `OCCTShapeBooleanCheckSingle` and `OCCTShapeBooleanCheckPair` (`OCCTBridge_Healing.mm:6967-6999`) have **no such guard on any of their `OCCTShapeRef` parameters** — each goes straight from `static_cast<OCCTShape*>(shape)` to `s->shape` with nothing between. This is exactly the shape CLAUDE.md's "Adding a New Wrapped Operation" section documents as required (guard the handle, not just hope `catch (...)` saves you — a null-pointer dereference is an OS signal the bridge's `catch (...)` cannot catch). `python3 Scripts/check-null-handle-guards.py` currently reports 0 violations and says nothing about either function — a real blind spot in the gate, not evidence the site is safe; worth a look when this is fixed. **Swift-side reachability of an actually-null `OCCTShapeRef` (as opposed to a valid handle wrapping a null `TopoDS_Shape`) has not been confirmed** — `Shape.handle` is set once at init and not otherwise nilled, so this may or may not be reachable through the public Swift API today. Measure before assuming either way.

## Tests

`BooleanCheckTests` (`Tests/OCCTModelingTests/OCCTModelingTests.swift:1640`) covers `isValidForBoolean`/`isValidForBoolean(with:)`. `BRepAlgoCheckTests` (`OCCTModelingTests.swift:4245`) covers `isBooleanValid`/`isBooleanValidWith`. Both suites exist independently with no cross-reference; neither exercises a null/invalid-handle input.

## OCCT functions

`BRepAlgoAPI_Check` (both single-shape and two-shape-plus-`BOPAlgo_Operation` constructors), via `Sources/OCCTBridge/src/OCCTBridge_Modeling.mm` and `Sources/OCCTBridge/src/OCCTBridge_Healing.mm`.

## Suggested fix

Converge onto one bridge implementation exposing the fuller `testSmallEdges`/`testSelfInterference`/`operation` surface (the `isBooleanValid*` family), guard all `OCCTShapeRef` parameters the way `OCCTShapeBooleanCheck` already does, have `isValidForBoolean`/`isValidForBoolean(with:)` forward onto it with today's implicit defaults preserved (so this is non-breaking), and merge the two test suites the way `Shape.defeaturedWithFullHistory`'s doc comment (`Shape+Modeling.swift:3450`) already models for a prior instance of exactly this "N spellings converge on one" pattern (#536).

SMKiloBOT:
## What & why

The same `BRepAlgoAPI_Check` operation was wrapped twice:
- `OCCTShapeBooleanCheck` in `OCCTBridge_Modeling.mm` (older, limited API, no `testSmallEdges`/`testSelfInterference` params)
- `OCCTShapeBooleanCheckSingle` and `OCCTShapeBooleanCheckPair` in `OCCTBridge_Healing.mm` (newer, full API, but missing null guards)

The newer implementations in `OCCTBridge_Healing.mm` had no null guards on the `OCCTShapeRef` parameters, which would crash on null input. This fix:
1. Adds null guards to both `OCCTShapeBooleanCheckSingle` and `OCCTShapeBooleanCheckPair`
2. Updates the Swift `isValidForBoolean` and `isValidForBoolean(with:)` to forward to the fuller implementations with implicit defaults
3. Removes the duplicate `OCCTShapeBooleanCheck` from `OCCTBridge_Modeling.mm` and its header
4. Fixes the stale index entry in `OCCTBridge.h`
5. Merges the test suites (`BooleanCheckTests` → `BRepAlgoCheckTests`)

Closes #1297

### Converge `BRepAlgoAPI_Check` bridge onto single implementation with null guards (#1297)

- `OCCTShapeBooleanCheckSingle` / `OCCTShapeBooleanCheckPair` in `OCCTBridge_Healing.mm` now guard null handles
- `Shape.isValidForBoolean` / `isValidForBoolean(with:)` now call the fuller implementations (with `testSmallEdges=true`, `testSelfInterference=true` defaults)
- Duplicate `OCCTShapeBooleanCheck` in `OCCTBridge_Modeling.mm` removed
- Test suites merged: `BooleanCheckTests` deleted, cases folded into `BRepAlgoCheckTests`

## SemVer impact

PATCH. Consumers using `isValidForBoolean` / `isValidForBoolean(with:)` now get the same deeper checks (`testSmallEdges`, `testSelfInterference`) that `isBooleanValid()` already exposed. No API surface changed; the behavior is strictly more correct (null-safe). No migration needed.

## Checklist

## Notes for the reviewer

- All 7 gate scripts pass (`check-null-handle-guards`, `derive-bridge-header-split --verify`, `check-bridge-index`, `check-docs-existence`, `check-docs-defaults`, `count-operations`, `check-borrowed-handles`)
- All 8 tests in the merged `BRepAlgoCheckTests` suite pass
- Related stress tests (`StressExhaustiveAPITests.isBooleanValid`) also pass


gsdali:
Superseded by #1370 (merged as 559ab3bc), which fixed the same issue (#1297): null guards added to `OCCTShapeBooleanCheckSingle`/`OCCTShapeBooleanCheckPair`, the older `OCCTShapeBooleanCheck` removed, `isValidForBoolean`/`isValidForBoolean(with:)` forwarded onto the fuller implementation, and the test suites merged. Same shape of fix, already landed. Closing without merging.

